PR #17, opened 2012-09-17
Fix creation of example.bai filename from example.bam

peterjc:
Old code would edit the first 'bam' in the filename,
thus example_bam_file.bam --> example_bai_file.bam,
and so not find the index file example_bam_file.bai
(if the prefered example_bam_file.bam.bai were missing).

Bug reported on the samtools-help mailing list:
https://sourceforge.net/mailarchive/message.php?msg_id=29820839

This commit also adds a stderr message listing the two
BAI filenames tried if the index is not found.


peterjc:
Can we close this PR?

PR #25, opened 2013-02-03
please replace thread option -@ with -p

domibel:
The unusual option -@ can cause unexpected problems with languages like perl.


domibel:
Are you still considering adding the -p flag (additional to -@) to samtools ? Should I prepare a new pull request?

PR #26, opened 2013-02-05
Added block size parameter for reduced I/O operations

specht:
## Summary

When running samtools on a cluster, the number of I/O syscalls becomes and important factor. An optional, larger buffer greatly improves overall speed because the number of filesystem calls can be reduced by a large amount when a slightly larger I/O buffer is used on each node.

For this purpose, I introduced a new option -z which can be used to specify the buffer size in kB for each input and output file. This currently only works with samtools view, sort and index. All remaining operations are not affected.

I measured the number of read/write syscalls using strace and a 126 MB SAM file, without and with a 4 MB buffer:

    $ strace -c ./samtools view [-z 4096] -Sbt hg19.fa A.sam > A-unsorted.bam 
    
            | read  write
    --------+------------
    default | 8040  3274
    -z 4096 |   25    11
    
    $ strace -c ../samtools sort [-z 4096] A-unsorted.bam A-sorted
    
            | read  write
    --------+------------
    default | 3285  3274
    -z 4096 |   17     7
    
    $ strace -c ../samtools index [-z 4096] A-sorted.bam 
    
            | read  write
    --------+------------
    default | 3290   231
    -z 4096 |   16     1

I also verified that the resulting checksums are the same with and without the -z option, as expected.

## Implementation details

In order to make the buffer size varibale (g_block_size) globally available, I added globals.c and global.h which currently serve the sole purpose of hosting the variable and making it accessible. I am aware that this is probably not the best solution, but considering the fact that the variable is required in many places throughout the code and this is C, I hope the solution is ok.

Whenever a gzip file is opened via gzopen or gzdopen, gzbuffer is used to set the buffer size. Whenever a file is opened via fopen, a buffer is allocated via malloc, assigned to the FILE* via setvbuf, and later freed again.

Finally, I had to change the underlying implementation in knetfile.c from using file descriptors to using FILE* because as far as I know, buffering is not possible using file descriptors. For this purpose, I added FILE *fp alongside the int fd to be used when a local file is being read.

Please don't hestitate to contact me with questions or comments at micha.specht@gmail.com.

Cheers,
Micha.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,13 +3,15 @@ CFLAGS=		-g -Wall -O2
 #LDFLAGS=		-Wl,-rpath,\$$ORIGIN/../lib
 DFLAGS=		-D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE -D_USE_KNETFILE -D_CURSES_LIB=1
 KNETFILE_O=	knetfile.o
-LOBJS=		globals.o bgzf.o kstring.o bam_aux.o bam.o bam_import.o sam.o bam_index.o	\
+LOBJS=		bgzf.o kstring.o bam_aux.o bam.o bam_import.o sam.o bam_index.o	\
 			bam_pileup.o bam_lpileup.o bam_md.o razf.o faidx.o bedidx.o \
-			$(KNETFILE_O) bam_sort.o sam_header.o bam_reheader.o kprobaln.o bam_cat.o
-AOBJS=		globals.o bam_tview.o bam_plcmd.o sam_view.o \
+			$(KNETFILE_O) bam_sort.o sam_header.o bam_reheader.o kprobaln.o \
+			bam_cat.o globals.o 
+AOBJS=		bam_tview.o bam_plcmd.o sam_view.o \
 			bam_rmdup.o bam_rmdupse.o bam_mate.o bam_stat.o bam_color.o \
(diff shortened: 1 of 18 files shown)


wookietreiber:
@lh3 @specht concerning compatibility with older zlib:

`zlib.h` includes `#define` directives about its own version:

```
$ grep 'define ZLIB_VER' /usr/include/zlib.h 
#define ZLIB_VERSION "1.2.3"
#define ZLIB_VERNUM 0x1230
```

These version directives could be checked to use `gzbuffer` only for `ZLIB_VERNUM >= 0x1235` (1.2.3.5) since this was the version that introduced `gzbuffer`.

`ZLIB_VERNUM` has been introduced in 1.2.0.2 (2003 AD) so checking this is safe.

Thus, we could use:

```c
  gzFile f = gzopen("foo.gz", "r");
#if defined(ZLIB_VERNUM) && ZLIB_VERNUM >= 0x1235
  gzbuffer(f, 1048576);
#endif
  gzclose(f);
```

lh3:
I believe the performance gain is primarily due to the change of `int fd` in knetfile.c to `FILE *fp`. Using `gzbuffer()` is largely irrelevant here. Samtools also has its own buffering mechanism. The more recent version uses hfile for local files.

Given that this PR is outdated. I will close it.

PR #63, opened 2013-07-27
Improve the performance of mpileup when -B is given.

mkb2000:
By default, mpileup reads the whole reference chromosome regardless whether option -B is given. This means every mpileup command would read about 200MB of ref for a run. However, when option -B is given to mpielup, which indicates no need to realignment, it is not necessary to read the whole ref chr. My modification makes mpileup only read the ref bases of given region from ref file. And this results in about 100-times faster for a single process working on one region when option -B is given. Feel free to contact me: Kangbo Mo kmo@student.unimelb.edu.au, the University of Melbourne, Australia

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -322,7 +322,16 @@ static int mpileup(mplp_conf_t *conf, int n, char **fn)
         bca->per_sample_flt = conf->flag & MPLP_PER_SAMPLE;
 	}
 	if (tid0 >= 0 && conf->fai) { // region is set
-		ref = faidx_fetch_seq(conf->fai, h->target_name[tid0], 0, 0x7fffffff, &ref_len);
+		if(conf->flag&MPLP_REALN){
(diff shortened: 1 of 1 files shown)


daviesrob:
I have a rebased version of this, but having made and tested it I'm not sure it's a good idea.  It shaves a second or two off a short mpileup on human chr. 1, but it only works if the -B option is used (disables BAQ realignment) and the -C option (mapping quality recalculation) is not used.  This is because both BAQ realignment and mapping quality recalculation may cause reference look-ups beyond the range given in the -r option, and it's not easy to work out how many extra bases need to be added (depends on the read lengths).  As -B is not default when running mpileup, it's possible that most people won't see any improvement.

Another problem is that while I think this now always gives the same result as before in all cases, it's difficult to be sure due to the complexity of mpileup.  It also leaves a potential for future bugs if anyone writes a function like bam_prob_realn_core that expects to see more of the reference and doesn't realise that it may not be present under some conditions.

daviesrob:
Closing as risks may outweigh benefits.

PR #66, opened 2013-08-29
Fix for unreferenced linker lib -ltinfo

abrenner:
This fixes the linker for NCURSES:
`/data/apps/binutils/2.23.2/bin/ld: bam_tview_curses.o: undefined reference to symbol 'keypad'
/data/apps/binutils/2.23.2/bin/ld: note: 'keypad' is defined in DSO /lib64/libtinfo.so.5 so try adding it to the linker command line
/lib64/libtinfo.so.5: could not read symbols: Invalid operation`

Signed-off-by: Adam Brenner aebrenne@uci.edu

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ PROG=		samtools
 INCLUDES=	-I.
 SUBDIRS=	. bcftools misc
 LIBPATH=
-LIBCURSES=	-lcurses # -lXCurses
+LIBCURSES=	-lcurses -ltinfo # -lXCurses
 
 
 .SUFFIXES:.c .o


yesimon:
Sorry to bring up an old issue but I think this is still a problem for CentOS and RHEL users. We know that curses was split into `libcurses` and `libtinfo`. In Debian and friends, the content of file

```
$ cat /usr/lib/x86_64-linux-gnu/libcurses.so
INPUT(libncurses.so.5 -ltinfo)
```

However, in CentOS, the library is such:
```
$ cat /usr/lib64/libcurses.so
INPUT(-lncurses)
```

Therefore Debian automatically picks up `-ltinfo` while CentOS does not, therefore resulting in the above problem. This is even true for CentOS 7 http://rpmfind.net//linux/RPM/centos/7.1.1503/x86_64/Packages/ncurses-devel-5.9-13.20130511.el7.x86_64.html.

The ways to solve this is adding `-ltinfo` to the Makefile, changing `-lcurses` to `-lncurses` or compiling without curses. I don't know what's the best way to address this, but this is something that should just work.

jmarshall:
This seems like a bug in CentOS surely (cf. [6668](http://bugs.centos.org/view.php?id=6668)).  Anyway, this will get addressed by a soon-to-be-committed configure script, which will detect appropriate curses libraries.

YiChingTang:
Hi, 
I changed -lcurses to -lncurses, but still get the error /lib64/libtinfo.so.5: could not read symbols: Invalid operation. 
any suggestions?

PR #87, opened 2013-11-10
make mplp_func and its requisite functions part of the public API

dzerbino:
The mplp_func is extremely useful to create your own iterators outside of the Samtools library, allowing it to be shared makes for easier development. Is it fine to open it up?

Thanks!

Daniel


daviesrob:
Closing as this would need much more work.

Please raise an issue if you would still like to do something like this.

PR #193, opened 2014-04-07
Bugfix of mpileup for outputing base positions

shengqh:
Fix the bug that the number of base positions is not equal to the number of based filtered in mpileup result.


DaGaMs:
I just added another pull request for the same bug to the current develop branch that passes all tests. I think your code will add unnecessary commas.

jmarshall:
Fixed by dffc67f33e6534acb96c6fea08675d5b7ce9b0f9 (PR #466).

PR #206, opened 2014-04-22
Convert rmdup to htslib

mp15:
Convert rmdup code to use htslib calls and constructs.


jmarshall:
Eventually superseded by fff21fe564679684b18db17ce90c0ed96469a9ae.

mp15:
Agreed though I would like to see the bam_library code unit test ported over.

PR #208, opened 2014-04-25
Check that both key and value exist in sam_header2key_val

ernfrid:
If either the key or the value don't exist then move on. The previous behavior resulted in segmentation faults when either of the two didn't exist in the header.


valeriuo:
Merged.

PR #253, opened 2014-07-21
Samtools sort multi-threaded does not error properly when disk writes fail.

godotgildor:
We found that when a disk was full, samtools sort using multi-cores
would not error out as expected, but instead continue, leading to
temp files of size 0.  This commit checks return code of writes
in this portion of the code and will error if writes are not
successful.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -637,6 +637,7 @@ int bam_merge_core2(int by_qname, const char *out, const char *mode, const char 
 		free(name);
 		if (tid < 0) {
 			fprintf(stderr, "[%s] Malformated region string or undefined reference name\n", __func__);
+			free(translation_tbl); free(fp); free(heap); free(iter);
 			return -1;
 		}
 		for (i = 0; i < n; ++i) {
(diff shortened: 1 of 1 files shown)


daviesrob:
The ideas here were eventually included via https://github.com/samtools/samtools/commit/0e260d917477da4c2323530ac973fd0245a2eded.  Since 1.3.1 samtools notices that the write failed when the partition fills up.

```
# Single thread:
~/git/samtools$ ./samtools sort -m 4M -T /tmp/small/test -o /tmp/srt.bam /tmp/16521_1#41.bam
[E::bgzf_flush] hwrite error (wrong size)
[E::bgzf_close] file write error
samtools sort: failed to create temporary file "/tmp/small/test.0048.bam": No space left on device

# Multi-thread
~/git/samtools$ ./samtools sort -@ 2 -m 4M -T /tmp/small/test -o /tmp/srt.bam /tmp/16521_1#41.bam
[E::bgzf_flush] hwrite error (wrong size)
samtools sort: failed to create temporary file "/tmp/small/test.0046.bam": No space left on device

```

PR #307, opened 2014-10-01
Some 32-bit fixes.

jkbonfield:
Fixed a couple compiler warnings with %ld vs size_t in printf.

Fixed some overly large seeds to srand48.

Changed some of the test SAM files to no longer use values greater
than 2^31 (the limit required by the SAM specification).


jkbonfield:
It appears that most of these changes made their way in via commit a25229a939999a3c065a823a20824cb0940e61d5.

The missing ones are the size of the integer fields in the test data, which are still 32-bit  unsigned instead of 31-bit max.  However the SAM spec has since been updated to allow -2^31 to +2^32 so the max values now are valid (and the code has evidently been fixed to no longer produce failures).

Therefore I'm closing this PR as already fixed (in 2014 no less!).

Issue #313, opened 2014-11-04
samtools bam2fq produces FASTA / FASTQ mixture


peterjc:
The name suggests the ``samtools bam2fq`` command will output FASTQ files, however it appears to produce FASTA records for reads with just ``*`` as their ``QUAL``:

```
$ ./samtools 2>&1 | grep -i Version:
Version: 1.1-22-gc61d8d1 (using htslib 1.1-74-g845c515)
$ ./samtools bam2fq examples/toy.sam 
>r001/2
TTAGATAAAGAGGATACTG
>r002
AAAAGATAAGGGATAAA
>r003
AGCTAA
>r004
ATAGCTCTCAGC
>r003
GCCTA
>r001/1
ATGGCGCTG
@x1
AGGTTTTATAAAACAAATAA
+
????????????????????
@x2
GGTTTTATAAAACAAATAATT
+
?????????????????????
@x3
TTATAAAACAAATAATTAAGTCTACA
+
??????????????????????????
@x4
CAAATAATTAAGTCTACAGAGCAAC
+
?????????????????????????
@x5
AATAATTAAGTCTACAGAGCAACT
+
????????????????????????
@x6
TAATTAAGTCTACAGAGCAACTA
+
???????????????????????
```

Input data:
```
$ more examples/toy.sam 
@SQ     SN:ref  LN:45
@SQ     SN:ref2 LN:40
r001    163     ref     7       30      8M4I4M1D3M      =       37      39      TTAGATAAAGAGGATACTG     *       XX:B:S,12561,2,20,112
r002    0       ref     9       30      1S2I6M1P1I1P1I4M2I      *       0       0       AAAAGATAAGGGATAAA       *
r003    0       ref     9       30      5H6M    *       0       0       AGCTAA  *
r004    0       ref     16      30      6M14N1I5M       *       0       0       ATAGCTCTCAGC    *
r003    16      ref     29      30      6H5M    *       0       0       TAGGC   *
r001    83      ref     37      30      9M      =       7       -39     CAGCGCCAT       *
x1      0       ref2    1       30      20M     *       0       0       aggttttataaaacaaataa    ????????????????????
x2      0       ref2    2       30      21M     *       0       0       ggttttataaaacaaataatt   ?????????????????????
x3      0       ref2    6       30      9M4I13M *       0       0       ttataaaacAAATaattaagtctaca      ??????????????????????????
[...]

PR #315, opened 2014-11-05
FASTQ output from samtools bam2fq even if no qualities.

peterjc:
Alternative fix for bug #313 where samtools bam2fq would produce FASTA records for reads without any quality scores.

Unlike my suggestion on PR #314, as per suggestion on #313 discussion, this instead this always gives FASTQ. When the quality is missing, it uses dummy quality strings of PHRED score 1 (this value can be changed via a new command line argument).


mp15:
#314 was merged

PR #322, opened 2014-11-21
Add warning about samtools depth's approx 8000 coverage limit

peterjc:
It might be sensible to add the same sort of warning to ``samtools mpileup`` as well?

Furthermore, I would prefer to see the magic 8000 limit defined as a ``#define`` constant in ``bam.h`` - would a pull request doing that be welcomed? Any preference on the macro name?

### DIFF
--- a/bam2depth.c
+++ b/bam2depth.c
@@ -70,6 +70,7 @@ int read_file_list(const char *file_list,int *n,char **argv[]);
 int main_depth(int argc, char *argv[])
 {
     int i, n, tid, beg, end, pos, *n_plp, baseQ = 0, mapQ = 0, min_len = 0, status = EXIT_SUCCESS, nfiles;
+    int max_depth = 8000;
     const bam_pileup1_t **plp;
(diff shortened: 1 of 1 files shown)


peterjc:
@jmarshall Does this PR need any further work from me?

jkbonfield:
This commit id 15f26e82 added depth -m parameter to do this.  I don't know why the -d PR wasn't used instead (it's more sensible IMO to keep the same arguments between pileup and depth) and the usage is a little different as -m is honoured unlike a hardcoded minimum of 8000 across all samples combined in mpileup, however it gets the job done.

Sorry it didn't get merged promptly Peter as it'd have saved work for all concerned.  I'll add a -d synonym to the -m for consistencies sake between the two commands and close this PR.

peterjc:
@jkbonfield Thanks James. Your proposal sounds very sensible.

See also #426 which pulls out my addition to the usage help text from this old commit as a new pull request.

PR #325, opened 2014-11-28
copy RG, BC and QT tags to the FASTQ header

mcshane:
bam2fq: pull in option from @lh3 htsbox repo to copy RG, BC and QT tags to the FASTQ header


jmarshall:
Apologies for the delay… rebased to rather more recently and merged.  Thanks.

PR #332, opened 2014-12-15
Mark dup alternative ii - FOR REVIEW ONLY

mp15:
This is the latest iteration of my work on mark duplicates and adds two new mark duplicates algorithms to samtools.  The first algorithm is a name sorted mark duplicates algorithm and the second a coordinate sorted algorithm which uses a novel data structure to reduce memory usage compared to the previous iteration.  I haven't fully tested this code yet, so I'm mainly after comments on whether how I've structured it is acceptable for a pull.


mp15:
Meh I don't have time to finish this

PR #336, opened 2015-01-09
[BUG FIX] remove cap of 8000 samtools depth

klugem:
Path from Tadigotla, Vasisht to fix cap of 8000 for coverage when running samtools depth

http://sourceforge.net/p/samtools/mailman/message/28166483/


domibel:
This pull request is not necessary anymore.
With https://github.com/samtools/samtools/commit/15f26e821f73ee02f58c4a496a998fb86ade3f54 it is now possible to change the default depth.
e.g. samtools depth -m 1000000

jmarshall:
As noted by @domibel, this has now been addressed.

Issue #260, opened 2014-08-04
Sort with thousands of contigs


tk2:
As discussed with Martin P. just now, it appears that the sort command on v0.2.0-rc11 fails to complete after many hours for a BAM file that contains reads mapped to thousands of contigs/scaffolds (e.g. when mapping reads to a genome assembly).

Issue #203, opened 2014-04-16
samtools merge very slow with many files


jrandall:
samtools merge appears to work fine for a small number of files (up to 50 or so it goes fairly fast).  However, when merging larger numbers of files, the operations to examine and build the header (before the actual data merge even begins) starts to dominate the time. 

Each additional input file that is opened is processed more and more slowly. I think this is because the regexes are searching through all of the data that has been loaded previously (this appears to especially be a problem for the PG ID searches). 

By the 100th file, it takes around a second to process each additional file. By the 200th, each file open takes a few seconds. By the 500th file, it takes around 50s each. By the 1200th file, it is around 200s. We had to stop testing beyond that, but expect that opening the 5000th file would take nearly an hour to process, and the whole process of getting all files opened and headers processed would take several months to complete. 

I have refactored some of the code in bam_sort.c to assist with profiling (07e333effffe6ce72e2d9fad124a52debc0e2906), and I believe the issue comes down to a few of the regexes dealing with PG lines. 

If we had a header parser, we could replace the regex implementation with one that built sensible data structures to keep track of the needed merge information.

SamStudio8:
See #481

daviesrob:
Closing as this was made more efficient a while back.

PR #337, opened 2015-01-09
Vastly improve performance when many header lines

nathanweeks:
We observed very slow execution (i.e., running for days) of the merge phase of samtools sort while running Trinity, which uses bowtie to map reads to Inchworm contigs, and sorts the resulting BAM. Profiling the execution for a few hours with Oracle Solaris Studio Performance Analyzer revealed that the vast majority of the execution time was spent executing spent executing regexec() in pretty_header().

After running some microbenchmarks (see below), I discovered that at least the glibc regexec() implementation on Ubuntu 12.04 LTS and RHEL 6.5, as well as the BSD implementation on FreeBSD 9.4, performs very poorly when the search string is very large (as an aside, OS X does not suffer this issue; the OS X regexec man page states "The regex implementation in Mac OS X 10.8 and later is based on a heavily modified subset of TRE (http://laurikari.net/tre/).") . The the calls to regexec() in pretty_header() use the entire set of header lines as the search string, which in the case of our Trinity run contains over 6.3 million lines (~146MB) due to the large number of Inchworm-generated contigs.

To address this issue, I replaced the use of regular expressions in pretty_header() with a few string.h functions.  systems with affected regular expression implementations, it can now complete in under a second.

In addition to running the regression tests, I performance-tested the updates using a "samtools merge" on the following two bam files (warning: ~190 MB each):

https://iastate.box.com/shared/static/reeory8xntayv15ke2ei.bam
https://iastate.box.com/shared/static/0nwiupf3xa43c5gz4y0r.bam

The updated samtools merge completed in about 1 minute 50 seconds on our Ubuntu 12.04 LTS server with AMD Opteron 6272 processors. Profiling indicated that the majority of the runtime was spent performing compression.

###Microbenchmarks

The following microbenchmark using regular expressions processed ~1100 per minute of the 6,358,111 header lines in our bam file 0000.bam (```samtools view -h 0000.bam | awk '/^[^@]/{exit}{print}' > header.txt```), so processing the entire file should take well over 3.5 days (for comparison, my OS X 10.9.5 workstation processes the entire 6,358,111 header lines in ~13 seconds):

```
// USAGE: ./a.out header.txt
#include <fcntl.h>
#include <regex.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/stat.h>

int main(int argc, char *argv[]) {
    const char record_type[] = "@SQ";
    struct stat buf;
    int count = 0;
    char *text_in_out; 

    FILE * header = fopen(argv[1], "r");
    fstat(fileno(header), &buf);
    text_in_out = calloc(buf.st_size, 1);
    fread(text_in_out, 1, buf.st_size, header);
    
    regex_t sq_regex;
    regmatch_t matches[1];
    if (regcomp( &sq_regex, "^@SQ.*$", REG_EXTENDED|REG_NEWLINE )) abort();
    char* sq_pointer = text_in_out;
    while (text_in_out+buf.st_size > sq_pointer && regexec( &sq_regex, sq_pointer, 1, &matches[0], 0) == 0) {
        count++;
        sq_pointer += matches[0].rm_eo + 1;
    }

    printf("count: %d\n", count);
}
```

The following version, which uses string.h functions, processes the same input file in under a half second:

```
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/stat.h>

int main(int argc, char *argv[]) {
    const char record_type[] = "@SQ";
    struct stat buf;
    int count = 0;
    char *text_in_out; 

    FILE * header = fopen(argv[1], "r");
    fstat(fileno(header), &buf);
    text_in_out = calloc(buf.st_size, 1);
    fread(text_in_out, 1, buf.st_size, header);

    char* record_begin_pointer, *record_end_pointer = text_in_out;
    size_t record_length, record_type_length = strlen(record_type);
    while ((record_begin_pointer = strstr(record_end_pointer, record_type))) {
        record_end_pointer = strchr(record_begin_pointer, '\n');
        if (record_begin_pointer == text_in_out || *(record_begin_pointer - 1) == '\n')
            count++;
    }

    printf("count: %d\n", count);
}
```

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -139,6 +139,24 @@ static void trans_tbl_destroy(trans_tbl_t *tbl) {
     kh_destroy(c2c,tbl->pg_trans);
 }
 
+// scan text_in, copying all lines beginning with the string record_type to
+// the buffer output_pointer. Return a pointer to the element in
+// output_pointer after the last copied newline.
+static char* copy_records(char* output_pointer, const char* record_type, const char* text_in) {
+    const char* record_begin_pointer;
+    const char* record_end_pointer = text_in;
+    size_t record_length;
+    while ((record_begin_pointer = strstr(record_end_pointer, record_type))) {
+        record_end_pointer = strchr(record_begin_pointer, '\n');
+        if (record_begin_pointer == text_in || *(record_begin_pointer - 1) == '\n') {
+            record_length = record_end_pointer-record_begin_pointer+1;
+            memcpy(output_pointer, record_begin_pointer, record_length);
+            output_pointer += record_length;
+        }
+    }
+    return output_pointer;
+}
+
 // Takes in existing header and rewrites it in the usual order HD, SQ, RG, PG CO, other
 static void pretty_header(char** text_in_out, int32_t text_len)
 {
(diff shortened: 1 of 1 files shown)


pdblood:
I confirmed that this patch solves similar issues I was encountering with Trinity when it calls 'samtools sort'.  I hope this can be pulled into the next release of samtools, as it is likely affecting a large number of Trinity users.

jmarshall:
Thanks for analysing this and contributing, and apologies for the delay in merging.  It was a bit untried to merge immediately before making the last release, but we should have got back to it sooner.

SamStudio8:
See also #203 from April 2014 where performance issues with large merges had already been identified previously.

jmarshall:
@SamStudio8:  There have been a number of _separate_ performance problems with this code, notably #203 (too many pretty_header calls), #260 (O(N<sup>2</sup>) algorithm for `@SQ` headers), this one (fails prey to tragic regexp implementation on some platforms).

PR #342, opened 2015-01-30
Added 'insert size' to the head of the IS section

nathanhaigh:
The output of the `samtools stats` command contains a header with a description of how to extract the following data section for further analysis. It also states what the columns will correspond to. In the header of the IS section, the insert size is not mentioned. Using the supplied command it will be the first column of the extracted data.

This diff adda this additonal info to the header and modifies the test files as well.


pd3:
Thank you, merged.

PR #357, opened 2015-03-04
Fix bugs in handling supplementary and secondary reads in samtools stats

mp15:
Fix several double counting bugs in handling supplementary and secondary reads in samtools stats.


mp15:
This has now been restructured with feedback from @pd3 and rebased against current develop.

PR #358, opened 2015-03-06
wgsim: Clarify that -S0 seeds with current time

sjackman:
I expected `-S0` to use a consistent seed for the random number generator, since the default value is `-1`, which uses the current time. Both `-S-1` and `-S0` use the current time. The value must be positive to set the seed. See https://github.com/samtools/samtools/blob/develop/misc/wgsim.c#L414
```c
srand48(seed > 0? seed : time(0));
```


jmarshall:
Thanks Shaun; merged as d4f5b07e91b56e909c34d2e8b4bee8327101aa26.

PR #361, opened 2015-03-09
Add sam_index_load() and samfetch() to legacy API

jmarshall:
Add legacy API sam.h functions sam_index_load() and samfetch() providing bam_fetch()-style iteration over either BAM or CRAM files.  (In general we recommend recoding against the htslib API directly, but this addition may help existing libbam-using programs to be CRAM-enabled easily.)

### DIFF
--- a/bam.h
+++ b/bam.h
@@ -460,7 +460,7 @@ extern "C" {
 
     /*!
       @abstract Retrieve the alignments that are overlapped with the
-      specified region.
+      specified region.  (For BAM files only; see also samfetch() in sam.h.)
 
       @discussion A user defined function will be called for each
       retrieved alignment ordered by its start position.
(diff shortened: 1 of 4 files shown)


jkbonfield:
I have no problems with the additional functions and the sam_fetch vs samfetch seems to match expectations (ie sam_open vs samopen et.al), but I wonder if we have anyone coding against the now defunct ones?  Is it best to leave them in?  On one hand it is robust to do so, but on the other hand it's the old deprecated API so we have no obligation to keep them around forever and persuading people to change is a good thing.

jmarshall:
The now-defunct `sam_fetch` was local to bam_tview.c and introduced in b3376c4dcd4e76286aaea361337856f95a95066b, i.e., on the way to 1.0, for the purposes of tview.  It wasn't actually `static` but it's used once in bam_tview.c and not declared in any headers.  So it's probably not too unfair to say anyone who used it is on their own...

jkbonfield:
Agreed.

Issue #362, opened 2015-03-11
`samtools index` silently fails to detect corrupted BAM files and `samtools view` reports them as truncated rather than corrupted


jrandall:
We have for a long time been teaching users that running `samtools index` on a bam file is good way to verify that a BAM file is not truncated (in addition to usefully generating the index). 

I have recently discovered a truncated BAM file which `samtools index` on 0.1.19 catches as possibly truncated, but for which samtools versions 1.0, 1.1, and 1.2 give no warnings (although `samtools view` does give a warning: `[main_samview] truncated file.` as does `samtools flagstat`: `[bam_flagstat_core] Truncated file? Continue anyway.`). 

I am working on preparing a minimal example that displays this behaviour, but so far I haven't figured out in which way this file is truncated such that it isn't caught by `samtools index` (as it does catch most truncated files). 

One hint may be the error this particular truncated file gives from Picard/GATK (from htsjdk): 
```
htsjdk.samtools.SAMFormatException: Did not inflate expected amount
        at htsjdk.samtools.util.BlockGunzipper.unzipBlock(BlockGunzipper.java:98)
```

jrandall:
Ok, after some digging I've managed to contrive a test bam that exhibits this behaviour. 

In fact, the file that causes this issue is technically not *truncated* (despite what `samtools view` assertively says) but *corrupted*. See below for a use case (based on data in the mpileup test dir: https://github.com/samtools/samtools/blob/develop/test/mpileup/ce%23large_seq.bam) that demonstrates this issue. 

I think there are two separate issues here:
- the misreporting of the nature of the problem when processing a corrupted file by `main_samview` - stating that the file is truncated when in fact the EOF block is intact and the actual problem has to do with a failure to successfully decompress an internal BGZF block. 
- the failure for `samtools index` to report any issues when a BGZF decompression error causes it to fail to completely process the BAM file.

```bash
$ cp ce#large_seq.bam test_corrupted.bam
$ md5sum test_corrupted.bam
bc6b0096f76e42675bec983c0c7d48ca  test_corrupted.bam
$ echo "corruption" | dd of=test_corrupted.bam bs=1 count=10 seek=150000 conv=notrunc
10+0 records in
10+0 records out
10 bytes (10 B) copied, 0.000480286 s, 20.8 kB/s
$ md5sum test_corrupted.bam
c185d46b4682750307ef26e28b5a5ad3  test_corrupted.bam
$ samtools --version
samtools 1.2
[...]

PR #363, opened 2015-03-15
Adds tests for and fixes handling of corrupted and truncated BAMs

jrandall:
Adds tests of `samtools view` and `samtools index` behaviour when working with corrupted and/or truncated BAMs (these use the `bats` framework from https://github.com/sstephenson/bats).

Modifies `bam_index.c` so that it actually returns the error status from `bam_index_build`.

Changes the error `main_samview` reports on an error from sam_read1 to say "corrupted" rather than "truncated" ("truncated" should be reserved for a lack of EOF block only whereas "corrupted" could mean any kind of corruption - in this case a failure to decompress a BGZF block).

Along with htslib PR samtools/htslib#189, this should fix issue #362

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,7 @@ check test: samtools $(BGZIP) $(BUILT_TEST_PROGRAMS)
 	test/split/test_expand_format_string
 	test/split/test_filter_header_rg
 	test/split/test_parse_args
-
+	test/bats/bin/bats test/bats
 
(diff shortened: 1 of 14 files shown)


jrandall:
N.B. tests should be expected to fail unless built against samtools/htslib#189

jmarshall:
The equivalent of the bam_index.c change has been made in 9bbc79ec8d793705007e2141c8f9483b39a4b397; the sam_view.c message change probably needs applying.  This PR remains open to represent the test cases it adds.

Issue #138, opened 2014-02-11
Samtools view -l no longer works


daviesrob:
The feature to filter reads by library is currently missing.

In samtools-0.1.19:
```
samtools view -l 3815246 test/dat/mpileup.1.bam | head -n 1
ERR156632.12704932	163	17	1	29	36S64M	=	195	293	TGGAGAAGGGGACAAGAGGTCCCCAACTTCTTTGCAAAGCTTCTCACCCTGTTCCTGCATAGATAATTGCATGACAATTGCCTTGTCCCTGCTGAATGTG	BFAFGFEIGFEFHHEIDKJGHHHJIIE=@KKGGKJGIBLLMFKMDIIHJKKHFELLLKFIHMHIHHIHLKJFCHFJIJAID=JHKFGHJIHKKCH:@HD?	MD:Z:64	RG:Z:ERR156632	AM:i:29	NM:i:0	SM:i:29	MQ:i:29	XT:A:M	BQ:Z:@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@hakklejlchhgijjgedkkkjehglghgghgkjiebgeihi`hc\igjefgihgjjbgY_gc^
```

In commit 66797d8:
```
samtools view -l 3815246 test/dat/mpileup.1.bam
Samtools-htslib-API: bam_get_library() not yet implemented
Aborted (core dumped)
```

jkbonfield:
Prerequisite: header API.

jmarshall:
PR #461 (in particular 13de48f8dd6bcd71277128143fadc9ab3aa50576) implemented `bam_get_library()`, thus providing the bare minimum of a header API to get `view -l` working again.

PR #368, opened 2015-03-23
Implement bam_get_library() for issue #138 ( Samtools view -l no longer works )

kirkmcclure:


jmarshall:
Thank you for the patch.  Unfortunately the approach doesn't really work: it calls `sam_header_parse2()` every time, thus reparsing the headers for every read, and leaks the memory that is allocated.

There is a proper SAM header API in the works that will fix `bam_get_library()` at a stroke, but unfortunately it keeps getting pushed down our priority list…

PR #369, opened 2015-03-25
add option -t for turning off translation; add option -p and -i for samtools merge

ionzhangj8:
1. Added option -t to turn on/off translation in samtools merge. By default translation is on. If multiple input bam fils contain same RG name or PG name, turning on translation will append some string to RG name or PG name. As a result, if we run the same merge multiple times we'll get different size of output bam files. By turning off translation (using -t0 at command line) will give us identical output bam files, which will behave as samtools-0.1.19.
2. Added option -p for multithreading bam loading in samtools merge. By default there is no extra thread for bam loading. -pN will create N extra threads for bam loading and speed up the process of samtools merge.
3. Add option -i for creating indexing file in samtools merge. When we run samtools merge, if we use option -i [output], the indexing file [output.bai] will be created at the end of merge. Because the data is already in the memory, creating indexing in merge will save us time on running samtools index. Mainly we can save the time of loading the merged bam file for indexing.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -32,11 +32,14 @@ DEALINGS IN THE SOFTWARE.  */
 #include <regex.h>
 #include <time.h>
 #include <unistd.h>
+#include <pthread.h>
 #include "htslib/ksort.h"
 #include "htslib/khash.h"
 #include "htslib/klist.h"
 #include "htslib/kstring.h"
 #include "htslib/sam.h"
+#include "htslib/bgzf.h"
+#include "htslib/hfile.h"
 
 #if !defined(__DARWIN_C_LEVEL) || __DARWIN_C_LEVEL < 900000L
 #define NEED_MEMSET_PATTERN4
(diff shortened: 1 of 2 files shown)


mp15:
Hi,
Got a few questions and comments:
1. What does -t do that -cp doesn't?
2. Could you fix up the indentation? It will make reviewing the patch easier.
3. It seems there is htslib changes that this depends upon? Could you list the pull request for these here please?

domibel:
The required htslib change is here:
https://github.com/samtools/htslib/pull/193

ionzhangj8:
1. option -cp generates same merged bam file for multiple runs. But it does not generate the identical merged bam file as samtools-0.1.19. With option -t0 the merged bam is identical to samtools-0.1.19 result.
2. I'll fix up the indentation.
3. Thanks to domibel, the dependency of htslib is samtools/htslib#193.

mp15:
The reason we don't generate an identical BAM file to samtools 0.1.19 is that this version of samtools can generate an invalid BAM file. Specifically because it ignores the headers it can generate from valid input files a BAM file where records without an @RG for the ID they refers to exist (likewise to @PG). Could you provide a use case where such would be useful to help me understand?

ionzhangj8:
We're using samtools 0.1.19. After we switch to samtools 1.2 some of our validation may fail because of using simple diff between old and new merged bam files. I may take option -t out and just use it for our internal purpose.
Thanks.

ionzhangj8:
I checked in my fix for indentation in both htslib and samtools. I also removed my option -t from samtools merge.
Build of my samtools depends on my checked in htslib.

ionzhangj8:
A change was made to https://github.com/samtools/htslib/pull/193, file htslib/hts.c for handling the case "samtools merge -i /dev/stdout ...".

PR #381, opened 2015-04-27
Adds missing sections to samtools man page

### LINKED ISSUE #372: man page text for samtools flagstat
Missing man page text for `samtools flagstat`

jrandall:
Adds (I think) all missing commands to samtools man page. 

Fixes #372

### DIFF
--- a/samtools.1
+++ b/samtools.1
@@ -1,11 +1,12 @@
-.TH samtools 1 "2 February 2015" "samtools-1.2" "Bioinformatics tools"
+.TH samtools 1 "20 April 2015" "samtools-1.3" "Bioinformatics tools"
 .SH NAME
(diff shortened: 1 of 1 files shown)


samtools.1
```diff
@@ -1,11 +1,12 @@
-.TH samtools 1 "2 February 2015" "samtools-1.2" "Bioinformatics tools"
+.TH samtools 1 "20 April 2015" "samtools-1.3" "Bioinformatics tools"
```
jmarshall: I bump these when I make the release… so will probably temporarily undo this line when this is merged.

jmarshall:
Um… this would be rather more reviewable if the insertions and the rearrangements were in separate commits.

jkbonfield:
Maybe we can just do man -l samtools.1 output as that's what ultimately matters.

PR #394, opened 2015-05-21
Filename to RG ID header fix when merging, fix tab for space in merge tests

mp15:
Fixes the fact that the -r option for samtools merge did not add entries to
the header.
Fixes the merge tests where a space accidentally crept in place of a tab in
test_input_1_b.


jmarshall:
Merged, though I dropped the unrelated

    -s VALUE override random seed (defaults to number of seconds since the Epoch)

bit.  Feel free to do that again, but for goodness sake just write something like _by default, different for each samtools invocation_.

PR #395, opened 2015-05-26
Output fmt option

jkbonfield:
Still to do, but maybe an issue for another pull request:
- Sanitisation of help formatting. The indentation differs for each and every sub-command.
- Sanitisation of help long/short options.  Even the already existant long opts differ between some sub-commands.
- Use of hts_opt_add and cast of htsFormat->specific.  Should this be migrated entirely into htslib and hide hts_opt structure internal to it?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,17 +198,17 @@ check test: samtools $(BGZIP) $(BUILT_TEST_PROGRAMS)
 	test/split/test_parse_args
 
 
-test/merge/test_bam_translate: test/merge/test_bam_translate.o test/test.o $(HTSLIB)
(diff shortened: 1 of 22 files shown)


jmarshall:
As `SAM_OPT_GLOBAL_OPTIONS()` takes one (single-char) argument for each global option rather than a C string containing all of them as before, adding a new global option now means that _every subcommand_ needs to be updated accordingly.  This can construed as a feature :smile:, as adding a new **global** option should indeed mean considering for each and every subcommand whether it is applicable.

SamStudio8:
:+1:

PR #412, opened 2015-06-23
Cram reheader

jkbonfield:
Added samtools reheader support for CRAM files, both to stdout (as in BAM) and also insitu.  Also updated reheader to add @PG lines, more of a useful proof of concept than it being essential (the new @PG could be added explicitly in the new hdr.sam instead if needed) and added tests.

### DIFF
--- a/samtools.1
+++ b/samtools.1
@@ -1254,7 +1254,7 @@ Number of temporary files to use.
 
 .TP \"-------- reheader
 .B reheader
-samtools reheader <in.header.sam> <in.bam>
+samtools reheader [-IP] <in.header.sam> <in.bam>
 
 Replace the header in
 .I in.bam
@@ -1262,6 +1262,24 @@ with the header in
 .I in.header.sam.
 This command is much faster than replacing the header with a
 BAM\(->SAM\(->BAM conversion.
+
+By default this command outputs the BAM or CRAM file to standard
+output (stdout), but for CRAM format files it has the option to
+perform an insitu edit, both reading and writing to the same file.  No
+validity checking is performed on the header, nor that it is suitable
+to use with the sequence data itself.
+
+.B OPTIONS:
+.RS
+.TP 8
+.B -P, --no-PG
+Do not generate an @PG header line.
+.TP 8
+.B -I, --insitu
+Perform the header edit insitu, if possible.  This only works on CRAM
+files and only if there is sufficient room to store the new header.
+The amount of space available will differ for each CRAM file.
+.RE
 
 .TP \"-------- cat
 .B cat
--- a/samtools.1
+++ b/samtools.1
@@ -1265,7 +1265,7 @@ BAM\(->SAM\(->BAM conversion.
 
 By default this command outputs the BAM or CRAM file to standard
 output (stdout), but for CRAM format files it has the option to
-perform an insitu edit, both reading and writing to the same file.  No
+perform an in-situ edit, both reading and writing to the same file.  No
 validity checking is performed on the header, nor that it is suitable
 to use with the sequence data itself.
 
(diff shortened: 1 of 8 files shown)


jmarshall:
```c
#include "cram/cram.h"
```

This isn't going to fly from samtools source code.  This isn't a public htslib header and isn't going to be available when building against an installed htslib rather than a nearby htslib source directory.

To use some CRAM functions from outside htslib, we should come up with a supportable ABI (for now probably the smallest subset of _cram/cram_io.h_ we can get away with!) and invent a new public _htslib/cram.h_ header declaring it.

samtools.1
```diff
@@ -1263,6 +1263,24 @@ with the header in
...
+to use with the sequence data itself.
+
+.B OPTIONS:
+.RS
+.TP 8
+.B -P, --no-PG
+Do not generate an @PG header line.
+.TP 8
+.B -I, --insitu
+Perform the header edit insitu, if possible.  This only works on CRAM
```
jmarshall: `sed` and `perl` call this **in-place** editing, and use the options `-i[suffix]`/`--in-place[=suffix]`… but if you really really like _in situ_, I think it's two words :smile:

jkbonfield: On Tue, Jun 23, 2015 at 08:10:26AM -0700, John Marshall wrote:
> `sed` and `perl` call this **in-place** editing, and use the options `-i[suffix]`/`--in-place[=suffix]`... but if you really really like _in situ_, I think it's two words :smile:

I was wary of using -i as I thought we use it elsewhere already to
indicate input files, but I'm wrong - it's just -o for output.  -i is
used in targetcut (inPen) and stats (insert-size), -I is stats (ID)
and mpileup (skip-indels).  So I'm happy to change that to --in-place
and -i.

James

-- 
James Bonfield (jkb@sanger.ac.uk) | Hora aderat briligi. Nunc et Slythia Tova
                                  | Plurima gyrabant gymbolitare vabo;
  A Staden Package developer:     | Et Borogovorum mimzebant undique formae,
https://sf.net/projects/staden/   | Momiferique omnes exgrabure Rathi. 


-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

jkbonfield:
On Tue, Jun 23, 2015 at 07:54:12AM -0700, John Marshall wrote:
> To use some CRAM functions from outside htslib, we should come up
> with a supportable ABI (for now probably the smallest subset of
> _cram/cram_io.h_ we can get away with!) and invent a new public
> _htslib/cram.h_ header declaring it. 

Hmm, the minimum we need is at least cram_fd and some basic
manipulation functions; reading and writing blocks, slices and
containers.   These are analogous to the various BGZF exports from
htslib.  This is fundamentally necessary for reheader as by design it
has tio read and write data in the file-specific blocks, unless we
move the reheader code into htslib itself which feels incorrect.

I'll have a think about the minimum needed.

On the related note is the whole @PG manipulation.  I added this as it
was a user request (NPG wanted it, and they were also the initial
requesters of making samtools reheader work with CRAMs).  I agree with
adding those too.  However to do so requires a header API!  This old
chestnut which has been lurking for ages.  I ought to review my code
and see just what it supports.

James

-- 
James Bonfield (jkb@sanger.ac.uk) | Hora aderat briligi. Nunc et Slythia Tova
                                  | Plurima gyrabant gymbolitare vabo;
  A Staden Package developer:     | Et Borogovorum mimzebant undique formae,
https://sf.net/projects/staden/   | Momiferique omnes exgrabure Rathi. 


-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

jkbonfield:
The new update still has the @PG header adjustments, but the primary changes here are the use of an external API.  See the comments in the related htslib PR for dicussions.  

A new PR is incoming to add further testing.

PR #413, opened 2015-06-24
Cram reheader 2.1

jkbonfield:
NOTE: contains the merge of output-fmt-option and appropriate fixes to get this working with the reheader code.

This was necessary in order to be able to test both version 2.1 and version 3.0 CRAM file functionality in samtools reheader.  This is vital given the code has two explicit code paths for each format version, but without the output-fmt-option branch there is no way for samtools test harness to control the file format versions being used.


jmarshall:
Test additions now merged as 3c2003eaa90228746e40c94525b7323c0fbe8634.

PR #415, opened 2015-06-30
build improvements

wookietreiber:


jmarshall:
This is the wrong approach.  In particular, I don't particularly want to introduce GNU Makeisms here.  Thanks; note that the maintainers have also been working on the problem you are trying to solve.

PR #419, opened 2015-07-01
Header merge efficiency

smowton:
Hi all,

I was working with some BAM files that had a large number of contigs (they were aligned against parts of the NCBI NT database, which has thousands of short contigs). The input files had disjoint contigs, having been aligned to disjoint subsets of the database. This turned out to make 'samtools merge' prohibitively expensive, spending hours in `trans_tbl_init`.

This patch improves the situation by (a) switching from a quadratic search for `@SQ` lines in `translate->text` to a linear pass over it, (b) switching from using `regexec` et al to plain old `string.h` operations, making a pass over a single file much faster, and (c) allowing the hashtable used to build the global translation table to live between reading each input, avoid quadratic cost of repeatedly building and expanding the table when there are more than a few input files involved.

This is only lightly tested with the particular workload that was causing trouble, but if you think the use case (many small contigs) is interesting, this might be useful to incorporate.


SamStudio8:
Very useful! See also #337, #260 and #203.
Some of the fixes you present may have been replicated by PR #337 which was
recently merged (
https://github.com/legumeinfo/samtools/commit/f0c42fee1d73a12919f09b4eaf3b66b3e22b4a52)
in to develop.

smowton:
There isn't duplication, actually-- that recent fix targets the prettify-header stage which comes after building the header in the first place. As I see it, the four PRs target similar problems but don't overlap:

#203 avoids quadratic-cost prettify-header stage
#260 switches from linear search for a contig to using a hash table, but still does linear search over the plaintext header
#337 further improves prettify-header by avoiding use of the libc regex functons
...and finally here we use a hash table for the plaintext header too, and persist the hashtable introduced in #260 across several files.

SamStudio8:
I didn't say there definitely was duplication, I just wanted to link in
other merge related issues!

smowton:
Sure, no defensiveness intended -- just noting that they are distinct issues

jmarshall:
Thanks for the patch and for pushing us to look at this again.  Rob extended this in #453 and the regexps are now gone from `bam_sort.c`… hopefully we've all finally squashed the algorithmic badness in here!

PR #420, opened 2015-07-02
Samtools cat for CRAM files

jkbonfield:
Note: PR against cram_reheader_2.1.  It could (perhaps) also be against cram_reheader, but either one is necessary as this builds on top of the initial CRAM API added there.


jmarshall:
Reopened as #451.

PR #421, opened 2015-07-02
Allow splitting of statistics on tags with `samtools stats`

SamStudio8:
* Removed several 'global' statistics parameters such as
    `cov_[min|max|step]`, `gcd_bin_size` to a new struct: `stats_info_t`.
* Each `stats_t` holds pointer to `stats_info_t` to avoid passing the
    info structure around to every function that needs access to the
    user specified "global" options.
* `output_stats` accepts new first parameter for target FILE*.
* New `init_structures` function accepts a `stats_t` and initialises it.
* `split_hash` khash structure maps RG IDs to an address in memory
    corresponding to a `stats_t` for that particular read group.
* New `split_stats` accepts a `stats_t` and `split_hash` and creates a
    new `stats_t`, inserting its address in to the hash.
* `all_stats` maintains old behaviour and is written to stdout.
* Other `stats_t` are written to files in the format <bam_fname>_<RG ID>.bamstat


mp15:
It appears we're going to need to revise this onto develop again.

SamStudio8:
@mp15 I have merged this with the latest `develop`?

mp15:
ff94ef1d2e16b27776a20dbc955967c6c41fcb9d Has already been merged with master, so needs to be trimmed off the branch

SamStudio8:
@mp15 Done - Thanks for the `rebase` help!

SamStudio8:
@mp15 Ping

jmarshall:
I've merged this, while taking the liberty of changing the option from `-T`/`--tag-split` to `-S`/`--split` and adding it to the man page.

As a followup, could you please:

* Add some way of specifying an output filename prefix.  At the moment, e.g. `samtools stats -S RG irods:blah/blah/foo.bam` will try to write to iRODS, which isn't going to fly in the Sanger environment.

* Add an indication of which readgroup/whatever this file is to the `# This file was produced…` header in  `output_stats()`.  That information is in the file name when it is created, but it's good practice for the data inside the file to be self-identifying too.

SamStudio8:
@jmarshall Comments addressed in #460

PR #422, opened 2015-07-02
-v option to invert pass/fail

dnephi:
While the bitwise flags can be negated with -f/-F options, one can't negate region selection or invert other criteria. Added a -v option to permit this behavior.


dnephi:
It doesn't seem to be working as expected. I'll write a unit test before requesting an additional pull request.

dnephi:
The problem was that I hadn't updated the get opt call for the function, so it was ignoring the -v option.

Works now! I haven't tested for a lot of contexts, but -v with no fields provides no output (inverting the default printing), and my test example with a bed file worked as expected.

PR #423, opened 2015-07-02
-v option to invert pass/fail for samtools view

dnbaker:
While the bitwise flags can be negated with -f/-F options, one can't negate region selection, bed interval overlap, or mapping quality. 

I had another pull request (442), but I made the mistake of making that request with the wrong branch, which injected code I use personally but didn't want to push to samtools. This new pull request is to use the correct branch.


jmarshall:
Thanks, this is an interesting idea.  What is the use case you have in mind, i.e., what filtering are you doing where this would be useful?

dnephi:
1. A couple use cases:
    1. Looking for regions where reads pile up with low (or 0) mapping quality, since -m only provides a minimum functionality.
    2. Get reads outside of a bed region without having to use bedtools. Since samtools view also performs sam/bam conversion, you can do your selection and a conversion at the same time.

dnbaker:
Not strictly necessary, and it’s been a long time. Closing.

PR #424, opened 2015-07-03
Add support for multiple file comparison in tests

SamStudio8:
* Add `out_map` hash to test_cmd args hash. Keys are the names (or paths
  from `$opts{path}`) of expected output files which map to the path
  of the expected output of that particular output file.
* Add `hskip` to test_cmd args hash. An integer used to splice out
  the desired number of lines from the head when reading the actual
  output file in to the `$out` array, allowing for a match to `$exp`.
* `out_map` does not override or interfere with the `out` argument and
  can also be used alongside it to check both stdout and multiple
  expected output files.
* `rm` automatically called to remove output files, though failures will
  be copied with the `.new` extension as per behaviour for `stdout` and
  `stderr` checks.

**Currently a test will fail on the first error encountered, `stdout` and `stderr` are checked before any expected output files**.

### DIFF
--- a/test/test.pl
+++ b/test/test.pl
@@ -147,10 +147,12 @@ sub cmd
 #       cmd=> command to test
 #       expect_fail=> as per passed()/failed()
 #       want_fail=> consider passed() if cmd() returns non-zero
+#       out_map => map output filenames to their expected result file (can be used alongside out)
+#       hskip => number of header lines to ignore during diff
 sub test_cmd
 {
     my ($opts,%args) = @_;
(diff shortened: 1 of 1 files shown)


jmarshall:
Prerequisite for #421.

mp15:
I now need this for getting test coverage for testing bam2fq

jkbonfield:
Onced over and merged.  We don't need more PRs depending on PRs and any issues could only be in the newly added functionality.

Issue #418, opened 2015-07-01
Usage message for samtools sort unclear


dg13:
Hi,
Very minor thing, but I find the usage message for samtools sort unhelpful. The usage message states "either -O or -T is required", but this doesn't seem to be true:

samtools sort tmp.bam -O bam
[bam_sort] no prefix specified for temporary files (use -T option)

while

samtools sort tmp.bam -T blah
[bam_sort] can't determine output format

so sort requires *both* -O and -T options to be used. Another odd thing is that, even though -o is "Write final output to FILE", if you don't use option -T, the final output is sent to stdout anyway. This is really confusing - it seems like -T is always required, whatever output you want. 
D
ps - using samtools 1.2

SamStudio8:
Duplicates #295

dg13:
OK thanks. There is a separate issue in my post (about -o -T leading to output on stdout rather than to the file specified in -o). Could this be merged also? Should I do this by posting in this issue?
D

====================================
Daniel Gaffney
Career Development Fellow
Wellcome Trust Sanger Institute
Wellcome Trust Genome Campus
Hinxton,
Cambridge, CB10 1SA
UK

http://www.sanger.ac.uk/research/faculty/dgaffney/

Tel: ++44 (0) 122 349 6866
====================================

SamStudio8:
That might be touched on by #356, I'm not entirely sure...

dg13:
I think its a separate issue i.e. behaviour of -o changes depending on whether you use the -T flag or not

====================================
Daniel Gaffney
Career Development Fellow
Wellcome Trust Sanger Institute
Wellcome Trust Genome Campus
Hinxton,
Cambridge, CB10 1SA
UK

http://www.sanger.ac.uk/research/faculty/dgaffney/

Tel: ++44 (0) 122 349 6866
====================================

SamStudio8:
It seems to me that the problem is caused by a check in the argument parsing for sort; where use of either `-O` or `-T` triggers the options to be parsed with a "modern" flag. Otherwise the options are parsed as "legacy" usage, which for `-o` sets `is_stdout = 1`:

`case 'o': if (modern) fnout = optarg; else is_stdout = 1; break;`

Perhaps this is what is meant by "either -O or -T is required [to trigger 'modern' options]".

What happens if you specify `-o` and `-O` without `-T`?

dg13:
Using -o and -O without -T gives:
[...]

Issue #295, opened 2014-09-09
samtools sort documentation misleading regarding -O and -T options


jrandall:
When I give `samtools sort` the `-O bam` option, it exits with error, saying: "[bam_sort] no prefix specified for temporary files (use -T option)".

However, the usage information indicates that "either -O or -T is required" - that doesn't seem right as `-T` seems to be required no matter what? (https://github.com/samtools/samtools/blob/develop/bam_sort.c#L1097)

Cheers,

Josh.

plijnzaad:
I concur, ran into the same problem (samtools 1.2, htslib 1.21), has this been fixed yet? 

(The -T requirement seems unnesssary since samtools could always invent its own prefix based on $TMPDIR or somesuch).

jmarshall:
Fixed by removing the legacy `<out.prefix>` invocation style; this removal enables `-T` to be optional.

PR #425, opened 2015-07-03
Clarify `samtools sort` usage

SamStudio8:
In the hope of closing #418, #295 and finally putting end-user confusion of
`samtools sort` to bed, I present the following minor patch:

* Add `-T PREFIX` to example usage to clarify it is in theory a
  "required option" to use non-legacy functionality.
* Add note below description of `-T` that it is a required option for
  non-legacy functionality.
* If the modern flag is up without setting output format with -O, recall
  `sam_open_mode` with the input file as `fnout` to automatically infer the
  output file format to match the input, thus no longer requiring -O for
  modern usage.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1163,14 +1163,15 @@ int bam_sort_core(int is_by_qname, const char *fn, const char *prefix, size_t ma
 static int sort_usage(FILE *fp, int status)
 {
     fprintf(fp,
-"Usage: samtools sort [options...] [in.bam]\n"
+"Usage: samtools sort -T PREFIX [options...] [in.bam]\n"
 "Options:\n"
 "  -l INT     Set compression level, from 0 (uncompressed) to 9 (best)\n"
 "  -m INT     Set maximum memory per thread; suffix K/M/G recognized [768M]\n"
 "  -n         Sort by read name\n"
 "  -o FILE    Write final output to FILE rather than standard output\n"
-"  -O FORMAT  Write output as FORMAT ('sam'/'bam'/'cram')   (either -O or\n"
-"  -T PREFIX  Write temporary files to PREFIX.nnnn.bam       -T is required)\n"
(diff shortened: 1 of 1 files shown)


jmarshall:
Good stuff… as discussed just now, let's consider also removing the old syntax and adding heuristics to detect potential for overwriting input files (due to mistaken old syntax invocation) and complain instead.

SamStudio8:
@jmarshall Would you still like to "support" the old syntax without explicitly saying so in the documentation? For example, as we know, in the event a user specifies `samtools sort -o <FILE>`, in legacy syntax this invokes a sort operation on FILE and writes to stdout. In "modern" usage this is an error because the output file has been specified without an input file (ignoring the `-O`, `-T` drama).

Could (should?) we *assume* that a user doing this is therefore trying to use legacy syntax and just warn them "assuming -o option as input in absence of actual input" and write to stdout? Or should we dump the new usage printout to the terminal?

SamStudio8:
Following on from that, we could also assume that a user who provides both a value for `-o` and an input, along with another argument is trying to use the `<out.prefix>` argument from legacy syntax and set `-T` for them?

I'm not saying we should and I feel like it would be better in the long run to just say "Hey, you're not using `sort` right!", but I thought I'd check what our intention was.

jmarshall:
> `samtools sort -o <FILE>` [...] In "modern" usage this is an error because the output file has been specified without an input file (ignoring the -O, -T drama).

Ignoring `-T` considerations, `blahblah | samtools sort -o file` is not an error; it is a request to sort standard input and write it to the specified file.  If it were truly necessary we could require this to be written as `blahblah | samtools sort -o file -`, but this is not what users expect (cf 23d901ec10bcf11ea87c35b93827fbc5240260b6).

SamStudio8:
@jmarshall Ah of course, I'd forgotten that you could sort from `stdin`. Should any detected attempt at using old syntax just exit with a warning then?

SamStudio8:
Superseded by #441

Issue #416, opened 2015-06-30
Samtools sort loses RG/PG entries when using tmp files


mktrost:
On the latest development version, when samtools sort uses temporary files, it appears to lose the RG/PG lines of the header.  The lines are maintained if it does not use temporary files.
Samtools version 1.2 keeps the RG/PG even when using temporary files, so it appears to be an issue that was introduced since then.

mp15:
This is caused by the same issue as #411 caused by 55c852f. Currently working on a fix.

mktrost:
Sorry, didn't realize this would be a duplicate.  Thank you for working on it!

mp15:
Fixed by 227c142c288b4757e9bd7907a0b52a9f16ee451b

Issue #411, opened 2015-06-22
samtools merge -c and -p options not working as expected?


mcshane:
samtools merge on a set of bam files gives these readgroups:
```
>samtools merge -u - ~sm15/merge_bug/*.bam | samtools view -H | grep ^@RG
@RG	ID:2822_6	LB:LIB01	SM:SAMPLE01	CN:SC	PL:ILLUMINA	DS:STUDY01
@RG	ID:2822_6-484275CC	LB:LIB01	SM:SAMPLE01	CN:SC	PL:ILLUMINA	DS:STUDY01
@RG	ID:2822_6-27339861	LB:LIB01	SM:SAMPLE01	CN:SC	PL:ILLUMINA	DS:STUDY01
@RG	ID:2822_7	LB:LIB01	SM:SAMPLE01	CN:SC	PL:ILLUMINA	DS:STUDY01
@RG	ID:2822_7-17F89869	LB:LIB01	SM:SAMPLE01	CN:SC	PL:ILLUMINA	DS:STUDY01
@RG	ID:2823_4	LB:LIB02	SM:SAMPLE01	CN:SC	PL:ILLUMINA	DS:STUDY01
@RG	ID:2823_4-7A173AF0	LB:LIB02	SM:SAMPLE01	CN:SC	PL:ILLUMINA	DS:STUDY01
```
Fine. But wanted to use the `-c` option to collapse readgroups with the same ID rather than appending a random string. Using `-c` though removes all `@RG` lines:
```
>samtools merge -cu - ~sm15/merge_bug/*.bam | samtools view -H | grep ^@RG
```

Similar problem with the `-p` option and the `@PG` tags.

kirkmcclure:
The bug was introduced with bam_sort.c commit 55c852f (May 21, 2015) -
````
./samtools merge -cu - test/dat/test_input_1_a.bam test/dat/test_input_1_c.bam
...
@SQ	SN:ref3	LN:4
@PG	ID:bull	PP:donkey
...
````
Previous version of bam_sort.c - e8f27a3 (May 12, 2015) -
````
./samtools merge -cu - test/dat/test_input_1_a.bam test/dat/test_input_1_c.bam
...
@SQ	SN:ref3	LN:4
@RG	ID:fish	PG:donkey
@RG	ID:cow	PU:13_&^&&*(:332
@RG	PU:*9u8jkjjkjd:	ID:colt
@PG	ID:bull	PP:donkey
...
````

[...]

PR #433, opened 2015-07-13
Fix for -c and -p in merge and sort

mp15:
-Add fix so that -c and -p put the first @RG or @PG into the output
-Add testcase to check the -c and -p


mp15:
Fixes #411

bam_sort.c
```diff
@@ -307,14 +307,15 @@ static void trans_tbl_init(bam_hdr_t* out, bam_hdr_t* translate, trans_tbl_t* tb
         regcomp(&rg_id_search, rg_regex.s, REG_EXTENDED|REG_NEWLINE|REG_NOSUB);
         free(rg_regex.s);
         kstring_t transformed_id = { 0, 0, NULL };
-        bool transformed_equals_match;
+        bool transformed_equals_match, not_found_in_output;
```
jkbonfield: I'm a bit unsure of how not_found_in_output is initialised.  It only gets set when rg_override is NULL.  Maybe we need a default value?

I couldn't see how to produce a test case where rg_override is NULL though so I don't know if this is possible.

jkbonfield:
Other than the one minor point, I have no other review comments to make.

mp15:
Fix #416

jmarshall:
Merged as 227c142c288b4757e9bd7907a0b52a9f16ee451b.

PR #435, opened 2015-07-14
Enable travis OS X to test cross platform bugs

mp15:
-Enable travis "osx" platform
-Add brew install xz to "osx"
-Disable sudo so that we deploy onto travis container infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 # Control file for continuous integration testing at http://travis-ci.org/
 
+os:
+  - linux
+  - osx
+
 language: c
 compiler:
   - clang


jmarshall:
1. Even if the travis script futzes around with brew and xz and makes the travis test run work, `make test` will still fail on users' vanilla OS X installations.

    We've discussed this before, and I would still rather recompress the one test file in question with a more widespread compressor instead.  Sadly this bloats the repository rather, so when doing that it would be worth checking whether that test data file needs to be quite so large.

2. Has travis's container infrastructure actually settled down yet?  Last I looked, projects with `sudo: false` appeared to be having all sorts of problems…

mp15:
1. Redone based on feedback.
2. Not sure, so I've removed this for now.

jkbonfield:
The file is large because it's a stress test on a large sequence.  Basically it's approx 1Mb long sequence with approx 60,000 cigar operations.  We want to test the boundaries rather than give it noddy simple cases, so this one single test is designed for that.

The alternative could be to write a generator script that generates both the test data and the expected output, but clearly that will take much longer to code up and test rather than just giving it the output from the previous samtools release to act as a regression test.

It's possible we could hack it a bit to improve gzip too, as gz is frankly rather dumb.  Or we could just revert back to my original implementation of the regression tester for this one file; instead of storing the expected output, store and compare the md5sums of the output to validate (and regenerate it using the previous version should we ever get an md5sum mismatch).

jkbonfield:
Indeed:

```
jkb@seq3a[samtools.../samtools] awk '{print $3,$4,$5,$6}' test/mpileup/expected/1.out|awk '{printf("CHROMOSOME_I\t%d\t%s\t%d\t%s\t%s\n",NR,$1,$2,$3,$4)}'|md5sum
be3acf0e4111038001d29d8fecf7544a  -
jkb@seq3a[samtools.../samtools] md5sum test/mpileup/expected/1.out
be3acf0e4111038001d29d8fecf7544a  test/mpileup/expected/1.out
jkb@seq3a[samtools.../samtools] awk '{print $3,$4,$5,$6}' test/mpileup/expected/1.out|gzip -9|wc -c
596103
```

So it could be stored in a compacted form and regenerated via gzip -d < file | awk... in just 600K - half the size of the current XZ output.

jmarshall:
Excellent, James, thanks for looking into this.  I've implemented that as 598f4a0f1dd9185e7e5b686e6beaf981465e1799.

So all that's needed here is asking travis to switch on OS X builds for us, and to add the `os:`/`‑ linux`/`‑ osx` bit.  And to figure out whether we need or want to do the `sudo: false` thing.

mp15:
I've just dropped the obsolete gzip patch from this pull and rebased to current develop.

.travis.yml
```diff
@@ -20,5 +23,4 @@ before_script:
   # repository $HTSREPO setting) and check out a corresponding branch with the
   # same name, if any, or otherwise the default branch.
   - .travis/clone ${HTSREPO:-git://github.com/samtools/htslib.git} $HTSDIR $TRAVIS_BRANCH
-
```
jmarshall: Unrelated change — accidental?

.travis.yml
```diff
@@ -1,5 +1,8 @@
 # Control file for continuous integration testing at http://travis-ci.org/
 
+os:
+  - linux
+  - osx
 language: c
```
jmarshall: language & compiler are tightly related.  os isn't.  Blank line please.

.travis.yml
mp15: Yes, fixed

.travis.yml
mp15: Done

jmarshall:
Thanks.  I've emailed the travis-ci folks to get this activated for samtools/samtools.

PR #436, opened 2015-07-14
Merge escape regex

mp15:
Add regex_escape to escape extended regex meta chars.
Add regex_escape calls to ID fields embedded into regex searches.

(incorporates fix for -cp to avoid merge conflicts)


dkj:
Looks good to me. Needs rebasing on current #433

mp15:
Done

mp15:
Updated to fix typo, add comment and rebase to current dev.

bam_sort.c
```diff
@@ -201,6 +201,50 @@ static void pretty_header(char** text_in_out, int32_t text_len)
...
+            case '}':
+            case '(':
+            case ')':
+            case '\\':
+            case '*':
+            case '+':
+            case '?':
+            case '|':
+            case '^':
+            case '$':
```
jmarshall: Why no `']'` (and yet the other flavours of closing brackets are there)?

jmarshall:
I would rather like to put my foot down here.  This escaped string is being written to a kstring.  Kstrings manage their own memory.  So there is no need to pay the risk of adding a function that itself calculates how much memory to allocate and crashes itself if the malloc fails.

All this function actually needs to look like is

```c
static void regex_escape_kput(const char* input, kstring_t *str)
{
    const char *s;
    for (s = input; *s; s++) {
        if (strchr(".[{etcetc", *s)) kputc('\\', str);
        kputc(*s, str);
    }
}
```
Isn't that simpler and easier to verify?

jkbonfield:
On Mon, Aug 17, 2015 at 05:43:42AM -0700, John Marshall wrote:
> I would rather like to put my foot down here.  This escaped string is being written to a kstring.  Kstrings manage their own memory.  So there is no need to pay the risk of adding a function that itself calculates how much memory to allocate and crashes itself if the malloc fails.
> 
> All this function actually needs to look like is
> 
> ```c
> static void regex_escape_kput(const char* input, kstring_t *str)
> {
>     const char *s;
>     for (s = input; *s; s++) {
>         if (strchr(".[{etcetc", *s)) kputc('\\', str);
>         kputc(*s, str);
>     }
> }
> ```
> Isn't that simpler and easier to verify?

Completely agreed on using kstring where appropriate as we have a
variable length string system already, we made as well use it.

There is no error checking in your above example, so it is essentially
equivalent to the previous code without checking for malloc.  kputc
can return EOF if the realloc fails and further more this isn't cached
in a state (like it would be for a FILE*), so every kputc needs
checking rather than just checking at the end or the starting malloc.

Hence it may actually be easier to compute a max size, malloc and test
for failure than test everywhere.  (Or easier, preallocate a kstring
with the worst case size so we can avoid all the kputc checks.)

James

-- 
James Bonfield (jkb@sanger.ac.uk) | Hora aderat briligi. Nunc et Slythia Tova
                                  | Plurima gyrabant gymbolitare vabo;
  A Staden Package developer:     | Et Borogovorum mimzebant undique formae,
https://sf.net/projects/staden/   | Momiferique omnes exgrabure Rathi. 


-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

bam_sort.c
dkj: On 17/08/15 13:38, John Marshall wrote:
>
> Why no |']'| (and yet the other flavours of closing brackets are there)?
>
Quizzed Martin about this and confirmed that opening ] didn't cause a 
problem.



-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

jmarshall:
> There is no error checking in your above example, so it is essentially
equivalent to the previous code without checking for malloc.  kputc
can return EOF if the realloc fails and further more this isn't cached
in a state

…shrug… there was me assuming kstrings exploded when they couldn't reallocate their buffers.

What we need in this project is a conversation about adopting a policy like:

* Allocating a gigabyte to analyse some data.  Yeah, check the return value, fail gracefully.
* Allocating 20 bytes for a string.  Failure here can be summarised as either can't happen or signals imminent doom anyway.  So the library routine should just die.  The less graceful failure mode in grossly unlikely situations is more than made up for by the maintenance benefit of not having three times as much code due to error paths on every string manipulation.

jkbonfield:
On Mon, Aug 17, 2015 at 06:02:31AM -0700, John Marshall wrote:
> * Allocating 20 bytes for a string.  Failure here can be summarised as either can't happen or signals imminent doom anyway.  So the library routine should just die.  The less graceful failure mode in grossly unlikely situations is more than made up for by the maintenance benefit of not having three times as much code due to error paths on every string manipulation.

The solution here is just to preallocate the kstring memory (eg via
ks_resize) so that we can then ignore all subsequent return values.

It keeps kstring in use while removing excessive numbers of checks.

-- 
James Bonfield (jkb@sanger.ac.uk) | Hora aderat briligi. Nunc et Slythia Tova
                                  | Plurima gyrabant gymbolitare vabo;
  A Staden Package developer:     | Et Borogovorum mimzebant undique formae,
https://sf.net/projects/staden/   | Momiferique omnes exgrabure Rathi. 


-- 
 The Wellcome Trust Sanger Institute is operated by Genome Research 
 Limited, a charity registered in England with number 1021457 and a 
 company registered in England with number 2742969, whose registered 
 office is 215 Euston Road, London, NW1 2BE.

mp15:
Obsoleted by #453

Issue #356, opened 2015-03-02
samtools sort -o should not require a filename


ekg:
I'd like to use `samtools sort -o -` in a pipeline, but I have to specify a dummy filename as in `samtools sort -o - x` in order for it not to break.

I think the `-o` flag should remove the need for a file name.

ekg:
I think I see what's happening. It needs the name in order to specify temporary files. A smarter way would require more code.

jmarshall:
Which samtools are you using?  In samtools 1.x, this is... more complicated.  In a pipeline in samtools 1.x, you can say

    ... | samtools sort -Obam -Ttmpprefix - | ...

which is supposed to be easier, and eventually at least one of -O and -T may be able to go away (but see also #295 and #349; currently they're both required it seems).

ekg:
@jmarshall I'm using 0.1.19-44428cd.

jmarshall:
The legacy `<out.prefix>` invocation style has now been removed on **develop**.  If you move to current samtools, I think you'll find the `sort` command's options more tasteful than those of 0.1.19.

Issue #349, opened 2015-02-10
New command line args are misleading (sort and likely other commands; ver 1.2)


balwierz:
It seems like samtools don't obey command line options. See the 3 examples below
````
$ samtools sort foo.bam
Usage: samtools sort [options...] [in.bam]
  -l INT     Set compression level, from 0 (uncompressed) to 9 (best)
  -m INT     Set maximum memory per thread; suffix K/M/G recognized [768M]
  -n         Sort by read name
  -o FILE    Write final output to FILE rather than standard output
  -O FORMAT  Write output as FORMAT ('sam'/'bam'/'cram')   (either -O or
  -T PREFIX  Write temporary files to PREFIX.nnnn.bam       -T is required)
  -@ INT     Set number of sorting and compression threads [1]

Legacy usage: samtools sort [options...] <in.bam> <out.prefix>
Options:
  -f         Use <out.prefix> as full final filename rather than prefix
  -o         Write final output to stdout rather than <out.prefix>.bam
  -l,m,n,@   Similar to corresponding options above
```
But this is exactly how I ran it.

Moreover after running
```
$ samtools sort -o bar.bam foo.bam
[E::hts_open] fail to open file 'bar.bam'
[bam_sort_core] fail to open file bar.bam
```

And after running
`$ samtools sort -@ 4 foo.bam -o bar.bam`
there is no result file. But it was sorting using threads, but never the generated the final bar.bam.

All of the above examples are valid according to the usage info displayed by samtools. For now I am sticking to the "legacy" options. Please fix it.

balwierz:
I guess it is because of
```c++
int bam_sort(int argc, char *argv[])
{
 ... [SKIP]
modern = 0;
for (i = 1; i < argc; ++i)
if (argv[i][0] == '-' && strpbrk(argv[i], "OT")) { modern = 1; break; }
...
```
(And it is mentioned to use either -O or -T, but in the wrong place.)
The usage should read:
`Usage: samtools sort <-T PREFIX | -O FORMAT> [options...] [in.bam]`

Loganskier:
I am having the same problem.  
The samtools sort functioned just fine for the following
samtools sort -o DP90_04pes_output.sam -O sam -T DP90_04_tmp -@ 8 DP90_04pe_output.sam

when I ran 

samtools sort -o DP90_24pes_output.sam -O sam -T DP90_24_tmp -@ 8 DP90_24pe_output.sam
[...]

PR #441, opened 2015-07-27
Remove `samtools sort` legacy usage [Fixes #418,#356,#349,#295]

### LINKED ISSUE #418: Usage message for samtools sort unclear

SamStudio8:
* Removes checking for `modern` or `legacy` usage and assumes all usage follows `modern` syntax.
* `-T` is no longer required and is set to `<in.bam>.tmp.NNNN.bam` by default.
* Without `-O`, output format inferred from input with `sam_open_mode`. For `stdin`, bam output is assumed unless specified.
* Updates `samtools.1` and `bam_sort.c` usage syntax.

#### Unambiguous legacy `-o` flag
When `-o` appears to be used as a flag (precedes another flag such as `-n`), the operation aborts:
```
$ samtools sort -o -n test_input_1_b.bam
[bam_sort] -o specifies output filename, got option '-n', aborting...
```

#### Ambiguous legacy `-o` flag
When `-o` appears to be used as a flag but could also be interpreted as valid `modern` usage, `getopt` swallows the argument and prevents it from being used as the input file anyway:
```
$ samtools sort -n -o test_input_1_b.bam
[bam_sort] No input specified, aborting...
```

#### Unambiguous `-o` with PREFIX
Where `-o` is deemed to look like a flag, attempting to specify a prefix without `-T` aborts with warning:
```
$ samtools sort -o -n test_input_1_b.bam prefix
[bam_sort] Use -T to specify a PREFIX
```

#### Ambiguous `-o` with PREFIX
`samtools` attempts to open the `prefix` file as input and will write to `-o`.
```
$ samtools sort -n -o test_input_1_b.bam prefix
[E::hts_open] fail to open file 'prefix'
[bam_sort_core] fail to open file prefix
```
**This has the potential to clobber the input file specified to `-o` but only if `PREFIX` is a file that already exists, which is an unlikely property of prefix**

#### Compatible with `stdin`
Uses `isatty` to detect whether or not a user has failed to specify a filename because they are forgetful or because there is input expected via `stdin`:
```
$ samtools sort -n -o out
[bam_sort] No input specified, aborting...
```
```
$ cat test_input_1_a.sam | samtools sort -n -o out
$ ll out
-rw-r--r--. 1 sam sam 784 Jul 27 16:58 out
```

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1185,33 +1185,22 @@ static int sort_usage(FILE *fp, int status)
 "  -m INT     Set maximum memory per thread; suffix K/M/G recognized [768M]\n"
 "  -n         Sort by read name\n"
 "  -o FILE    Write final output to FILE rather than standard output\n"
-"  -O FORMAT  Write output as FORMAT ('sam'/'bam'/'cram')   (either -O or\n"
-"  -T PREFIX  Write temporary files to PREFIX.nnnn.bam       -T is required)\n"
-"  -@ INT     Set number of sorting and compression threads [1]\n"
-"\n"
-"Legacy usage: samtools sort [options...] <in.bam> <out.prefix>\n"
-"Options:\n"
-"  -f         Use <out.prefix> as full final filename rather than prefix\n"
-"  -o         Write final output to stdout rather than <out.prefix>.bam\n"
-"  -l,m,n,@   Similar to corresponding options above\n");
+"  -O FORMAT  Write output as FORMAT ('sam'/'bam'/'cram')\n"
+"  -T PREFIX  Write temporary files to PREFIX.nnnn.bam\n"
+"  -@ INT     Set number of sorting and compression threads [1]\n");
     return status;
 }
 
 int bam_sort(int argc, char *argv[])
 {
     size_t max_mem = 768<<20; // 512MB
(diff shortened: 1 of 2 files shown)


SamStudio8:
Any feedback, @jmarshall ?

SamStudio8:
@jmarshall Just squashed this to tidy up the commit log. It was okayed at the merge party but I'm happy to make any changes if it still needs work.

SamStudio8:
@jmarshall I believe this should be in the 1.3 milestone if we still intend to deprecate legacy usage in this release!

jmarshall:
Yep, fear not, it's on the short list.  At this point, I think everything marked **1.3-maybe** will be making the cut.

PR #443, opened 2015-07-28
Handle malloc failures in bam_sort

mp15:
If a malloc fails in pretty_header, print an error and return true.
If a malloc fails in trans_tbl_init, print and error and die.


mp15:
abandoned in favour of @daviesrob code

PR #446, opened 2015-08-04
Misc code cleanups

mp15:
Replace != with < in for loops in errmod.
Replace 0's with NULL in bam_plcmd.c
Replace exit(1)'s with exit(EXIT_FAILURE)'s in bam_plcmd.c.

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -25,6 +25,7 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include <math.h>
 #include <stdio.h>
+#include <stdlib.h>
 #include <unistd.h>
 #include <ctype.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


bam_plcmd.c
```diff
@@ -276,13 +276,13 @@ static void group_smpl(mplp_pileup_t *m, bam_sample_t *sm, kstring_t *buf,
             const bam_pileup1_t *p = plp[i] + j;
             uint8_t *q;
             int id = -1;
-            q = ignore_rg? 0 : bam_aux_get(p->b, "RG");
+            q = ignore_rg ? NULL : bam_aux_get(p->b, "RG");
```
jmarshall: Okay re `NULL`, but the space before `?` is a stylistic choice.  In samtools the existing style is 80% without; in bam_plcmd.c it's exclusively without.  So please leave the existing style here. &lt;/pedantic-picard-hat&gt;

bam_plcmd.c
```diff
@@ -276,13 +276,13 @@ static void group_smpl(mplp_pileup_t *m, bam_sample_t *sm, kstring_t *buf,
...
             if (q) id = bam_smpl_rg2smid(sm, fn[i], (char*)q+1, buf);
             if (id < 0) id = bam_smpl_rg2smid(sm, fn[i], 0, buf);
             if (id < 0 || id >= m->n) {
                 assert(q); // otherwise a bug
                 fprintf(stderr, "[%s] Read group %s used in file %s but absent from the header or an alignment missing read group.\n", __func__, (char*)q+1, fn[i]);
-                exit(1);
+                exit(EXIT_FAILURE);
```
jmarshall: `EXIT_FAILURE` is supplied by stdlib.h.  Presumably bam_plcmd.c gets it by accident from some header; but please add also a `#include` to bam_plcmd.c to make sure.

bam_plcmd.c
mp15: Fixed.

bam_plcmd.c
mp15: Done

jmarshall:
Thanks, though let's not get distracted from the significant stuff at this stage.  Cherry-picked as b8ff342fd5846c5df6129d67b154aa8c1a157d8f.

PR #448, opened 2015-08-05
mpileup output updates for VCF4.3

mcshane:
* the unseen allele is to be specified as <*> rather then X or <X>
      as has been the case in mpileup for a while.
      See samtools/hts-specs@4a917457cb9be262bbe22634a90ece2e0dd08f48
      Note that bcftools call supports X, <X> and <*>,
      see samtools/bcftools@802ff309822b8648cc3b7022e3a72e3312e03279
    
    * add options to output AD,ADF,ADR (samtools/hts-specs#78)
    
    * deprecate DV,DP4,DPR annotations as largely superseded by
      the new AD tags


jmarshall:
The (still-functional albeit deprecated) format `DPR`'s value has changed.  Is this intended?

mcshane:
No, not intended. Is it one of the test outputs or can you point me to an example?

jmarshall:
It's trivially demonstrable with the test suite files so I guess it's probably across the board:

    samtools mpileup -r 17:200-400 -vu -t DPR test/mpileup/mpileup.1.bam

The change reduces these from `0,7,0` or `0,13,0` to `0,6,0` or `0,8,0` respectively etc on every record.

mcshane:
Thanks. Good spot. Fixed now.

jmarshall:
It appears the same bug exists in @pd3's bcftools…

mcshane:
yes. i'll let him know once he is back from holiday

jmarshall:
The plan was that the notification email just did that…

PR #459, opened 2015-08-25
Inverse video mode for tview and also changed some colours.

### LINKED ISSUE #457: Colorblind-friendlier output for tview
Many colorblind users can see the colors used in tview, but not when they are on so few pixels at a time.

Could you add an option to tview to color the background instead of the foreground?

jkbonfield:
For consideration.  It does change the default colour set, but doesn't enable reverse video until 'v' is pressed.

Fixes #457

### DIFF
--- a/bam_tview_curses.c
+++ b/bam_tview_curses.c
@@ -110,6 +110,33 @@ static void curses_clear(struct AbstractTview* tv)
     clear();
     }
 
+static int curses_init_colors(int inverse)
+{
+    if (inverse) {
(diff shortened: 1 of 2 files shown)


bam_tview_curses.c
```diff
@@ -110,6 +110,33 @@ static void curses_clear(struct AbstractTview* tv)
     clear();
     }
 
+static int curses_init_colors(int inverse)
+{
```
jkbonfield: Yes yes I know... the indentation is sane.  Sorry for not following the local code conventions. :-)

PR #460, opened 2015-08-25
Add stats split prefix support and split indicator

SamStudio8:
Addressing @jmarshall's comments to #421:

* Add `-P`, `--split-prefix` to `samtools stats` to prepend
  filenames created by `-S` or `--split` with a path or string.
* Add an additional comment line at the head of `samtools stats`
  output to clarify whether the statistics pertain to all or
  a tagged subset of reads.
* Moved `split_tag` to `stats_info_t` to allow simpler access to
  its value during `output_stats`.

I've incremented arguments to `tail` in `test.pl` to "ignore" this additional output line.

### DIFF
--- a/samtools.1
+++ b/samtools.1
@@ -652,6 +652,11 @@ is the input filename and
 has been encountered as the specified tagged field's value in one or more
 alignment records.
 .TP
+.BI "-P, --split-prefix " STR
+A path or string prefix to prepend to filenames output when creating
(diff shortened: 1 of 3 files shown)


SamStudio8:
I should add, the "split indicator" appears thus:

```
# This file contains statistics only for reads with tag: RG=s1_a_1
```
or
```
# This file contains statistics for all reads.
```

PR #461, opened 2015-08-26
Fixes #450

### LINKED ISSUE #450: Fallout from the output-fmt-options merge
PR #395 has been merged, but there are a few aspects of the additions that need further tidying up:

- The following don't use global options, though at least some probably should:
  - [ ] bam_cat.c
  - [ ] bam_index.c
  - [ ] bam_md.c
  - [ ] bam_rmdup.c
  - [ ] padding.c
- Need to integrate existing `ref.fasta` arguments or `-f`/etc options with the new `--reference` in
  - [ ] bam_md.c (once it uses global options!)
  - [ ] bam_plcmd.c
  - [ ] bam_tview.c
  - [ ] cut_target.c
  - [ ] sam_view.c
- [ ] Need to resurrect `sam_open_mode()` or an equivalent in bam_mate.c and bam_sort.c

jkbonfield:
At least partially.

Added global I/O options to calmd (mostly pointless), depad (now also in the standard list of sub-commands) and rmdup.  Rmdup now actually "works" too due to  finally implementing bam_get_library().

Made the --reference and other options mostly do the same thing.  Mpileup still treats these differently (as it changes the output), while some other tools process them as a separate argument permitting them to both be specified and differ, but accepting --reference as a default for the reference option.  The reason is that --reference MUST much the reference used when encoding a cram, but we may wish to use a subtly different reference for the actual tool (SNP comparison, etc).

Samtools cat and index are unchanged as they operate on low level file blocks and so the global options are inappropriate.

NOTE: I haven't yet done anything with sam_open_mode().

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,14 +146,14 @@ bam_import.o: bam_import.c $(HTSDIR)/htslib/kstring.h $(bam_h) $(HTSDIR)/htslib/
 bam_index.o: bam_index.c $(htslib_hts_h) $(htslib_sam_h) $(HTSDIR)/htslib/khash.h
 bam_lpileup.o: bam_lpileup.c $(bam_plbuf_h) $(bam_lpileup_h) $(HTSDIR)/htslib/ksort.h
 bam_mate.o: bam_mate.c $(sam_opts_h) $(HTSDIR)/htslib/kstring.h $(htslib_sam_h)
-bam_md.o: bam_md.c $(htslib_faidx_h) $(htslib_sam_h) kprobaln.h
+bam_md.o: bam_md.c $(htslib_faidx_h) $(htslib_sam_h) kprobaln.h $(sam_opts_h)
 bam_pileup.o: bam_pileup.c $(sam_h)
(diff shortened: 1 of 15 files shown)


jkbonfield:
Oops, forgot that. I'll work on it.

jmarshall:
We could of course consider that in future, but we are most certainly not going to devote any brain cycles whatsoever to it before the upcoming release!

jkbonfield:
Gah, noticed rmdup still not working quite right.  Hold off on this PR for now.
I had incoming code for sam_open_mode() too.

jkbonfield:
I believe I've now caught all the places we needed to add sam_open_mode(), fixing some rather broken regressions. Oops.  Also fixed rmdup so it actually used getopt_long.  It's been a long day.

TODO: our test harness really needs to be able to check the file formats and to repeat all commands in all formats.

Anyway, I'm happy with this PR now.  The only change I may wish to make is samtools view -o to auto-detect.  I didn't add it as it didn't previously do that either, but it's a bit of an outlier compared to other tools so perhaps this should be brought into line.

PR #464, opened 2015-09-07
Rename bamshuf to collate

mp15:
Now that bamshuf can handle CRAM files we have renamed it so that it has a more appropriate name.

### DIFF
--- a/bamshuf.c
+++ b/bamshuf.c
@@ -1,4 +1,4 @@
-/*  bamshuf.c -- bamshuf subcommand.
+/*  bamshuf.c -- collate subcommand.
 
     Copyright (C) 2012 Broad Institute.
     Copyright (C) 2013 Genome Research Ltd.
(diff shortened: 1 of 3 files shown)


jmarshall:
I thought _collate_ meant this too, but AFAICT it is actually about [sort order](https://en.wikipedia.org/wiki/Collation).  SQL uses `group by`, of course.  Do we have evidence that this usage of collate is common outside our (samtools-core-, Sanger-, or UK-) enclave?

jmarshall:
Documentation or otherwise needs to make clear that this is doing some nice randomisation as well as just plain collation.

daviesrob:
I discovered the other day that GNU sort does something similar to this, which it calls 'random-sort'.  This is what the info page says about it:

```
`-R'
`--random-sort'
`--sort=random'
     Sort by hashing the input keys and then sorting the hash values.
     Choose the hash function at random, ensuring that it is free of
     collisions so that differing keys have differing hash values.
     This is like a random permutation of the inputs (*note shuf
     invocation::), except that keys with the same value sort together.
```

bamshuf.c
```diff
@@ -1,4 +1,4 @@
-/*  bamshuf.c -- bamshuf subcommand.
```
mp15: Oops shouldn't change file name here

jmarshall:
Merged with trivial doco changes as 97c9883e139a127a8b36353b0e7ec12c8319ef38.

PR #466, opened 2015-09-13
Added code to fix #343, thanks @cyrusmaher

### LINKED ISSUE #343: For mpileup, -O option gives positions pre-BQ filter
For samtools mpileup, the --output-BP/-O option returns the positions of bases that are otherwise filtered in the output using --min-BQ > 0. This means it's no longer possible to match the position with the base. One can set --min-BQ to zero and then manually apply a BQ filter after, but hopefully it would be a simple tweak to match up the bases and the positions in the output.

Thanks for all your hard work...

DaGaMs:
mpileup outputs more read base positions than reads when the `-O` option is used in conjunction with `-Q`. This pull request should fix this problem.

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -637,9 +637,8 @@ static int mpileup(mplp_conf_t *conf, int n, char **fn)
                             int c = bam_get_qual(p->b)[p->qpos];
                             if ( c < conf->min_baseQ ) continue;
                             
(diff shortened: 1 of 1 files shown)


jkbonfield:
Looks good.   I merged the two commits, rebased and pulled it in.  Thanks for the bug fix.

PR #467, opened 2015-09-16
Fix warnings from adding HTS_RESULT_USED to htslib prototypes (rebased).

daviesrob:
Companion to pull request samtools/htslib#271. Add checks to return values from
functions now marked with HTS_RESULT_USED in htslib. Also fix some other
obvious places where errors were not handled correctly. Try to improve
clean-up code so that memory is not leaked and files are closed, even when
exiting due to an error.


jmarshall:
Thanks.  Rebased and merged as f6db7827764c375f8ca30d026cc4e8e2b2fdab7a.

PR #469, opened 2015-09-18
Fix how stats handles Supp reads on alternate haplotypes

mp15:
Fix how stats handles Supp reads on alternate haplotypes by treating them similar to reads flagged secondary.
Fix typo in getopt_long case statement where it's case 1 rather than case '1'

### DIFF
--- a/stats.c
+++ b/stats.c
@@ -37,6 +37,7 @@ DEALINGS IN THE SOFTWARE.  */
 
 */
 
+#include <sys/stat.h>
 #include <unistd.h> // for isatty()
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


jmarshall:
Get rid of f2e0a75d54ad2e0fdf3e2682246fea483703cab8, as noted previously.

jmarshall:
Please explain how this has been tested.

mp15:
This is an important fix but it needs work. This fix was giving the improvement I wanted but due to a bug it was doing it by doing the opposite of what I intended (aka excluding the SUPP's on primary contigs not alt). I will work on this further in the near future.

PR #471, opened 2015-09-25
Make samtools merge use CRAM_OPT_SHARED_REF to share references between file handles

jkbonfield:
NB. This requires that all files being merged have the same set of
references in the same order. (Is this already an assumption used in
merging?)


jkbonfield:
NB: Rebased and forced push.

bam_sort.c
```diff
@@ -1227,6 +1229,12 @@ int bam_merge_core2(int by_qname, const char *out, const char *mode,
         if ((translation_tbl+i)->lost_coord_sort && !by_qname) {
             fprintf(stderr, "[bam_merge_core] Order of targets in file %s caused coordinate sort to be lost\n", fn[i]);
         }
+
+        if (!refs && cram_get_refs(fp[i]))
+            refs = cram_get_refs(fp[i]);
+
+        if (refs && hts_set_opt(fp[i], CRAM_OPT_SHARED_REF, refs))
+            return -1;  // FIXME: memory leak
```
jmarshall: This function now has `goto fail` style cleanup, so you should be able to fix the FIXME without much ado.

bam_sort.c
```diff
@@ -42,6 +42,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include "htslib/klist.h"
 #include "htslib/kstring.h"
 #include "htslib/sam.h"
+#include "htslib/cram.h"
```
jmarshall: And in the makefile.

bam_sort.c
jkbonfield: I was just copying the coding style at the time (there are LOTS of that same comment in my version, ugh).

Will merge with upstream and fix this additional one (+makefile).

jkbonfield:
I see there are still a mixture of goto fail vs return -1 in the bam_merge_core code, but these are outside the lines touched by this PR.  I'm not sure why specifically some were deemed to be immediate returns and others warranted tidying up (e2cab6fc7d6f055093498bff95eb1400d3e507c2), but that's perhaps a topic for another issue and not related to this specific change.

We still have the open / unanswered question of whether this whole PR is even valid.  I don't know whether we can assume sanity with regards to merging BAMs and @SQ lines matching.  I can't see how it works if they don't, but I've learn enough to expect the unexpected!  If it's permissable to merge files with different headers then all bets are off when it comes to optimisation.  We could either scan the headers to make sure they're all the same before entering the efficient/shared mode, or just reject the PR if that is non-trivial. 

(The people that initially asked for this change have, to my knowledge, never used it.)

mp15:
I have encountered a similar problem whilst attempting to call a couple of hundred CRAMs with mpileup (ending up with hundreds of file handles). Could we have another look at merging this please?

bam_sort.c
mp15: Looks to have been addressed

mp15:
I've taken the liberty of adding a commit to add goto fails to the entire function. If we're going to fix one return -1 we might as well do the lot.

jkbonfield:
This has been hanging around like a bad smell for some time.  I can rebase it again if people want it merged.

What's the consensus from the mpileup users / authors, specifically the "NB. This requires that all files being merged have the same set of references in the same order. (Is this already an assumption used in
merging?)" comment.  @mcshane @pd3 

The samtools man page seems to imply that merge works with arbitrary header orders, which makes shared references non-workable in the current implementation.  Is this true?

daviesrob:
Merge has code to build a translation table for `@SQ` entries in `trans_tbl_add_sq()`, so in theory it might be able to handle inputs with references in different orders.  In practice, I suspect trying this will not work because data would be added to the heap used for merging at the wrong time.  Files that skip references, but otherwise follow a global ordering probably will work, though.

As `trans_tbl_add_sq()` goes through all of the headers already, I think the best solution would be to make it check that the ordering is consistent.  It could then turn off this optimisation if it discovers that the ordering in each file is not the same.

jkbonfield:
That sounds like a pragmatic middle ground.

jkbonfield:
I've rebased this ancient PR and tested it with out of order data.  It did have one issue, but as suggested above I've added a commit to use the existing out-of-order check to avoid sharing the reference which resolves this.  Please don't squash it as we ought to record the time differences between the initial commits (now squashed together) and the updates.

There are still existing out of order issues, but they're nothing to do with this PR and best fixed elsewhere.  For details, my analysis:

---

3 files with alignments in the order ref1, ref2, ref3, but SQ headers different: {ref1, ref2, ref3}, {ref2, ref3, ref1} and {ref3, ref1, ref2}.

I'm testing merge and mpileup on the same file 3 times vs the 3 different copies of the file with rotated headers to see how they compare.

CRAMs created by directly converting sam to cram:

    samtools view --write-index -T ref.fa -o _1.cram _1.sam

Test: merge SAM
    samtools merge _1.sam _1.sam _1.sam
    samtools merge _1.sam _2.sam _3.sam

    PASS: both outputs match

Test: merge CRAM
    samtools merge _1.cram _1.cram _1.cram
    samtools merge _1.cram _2.cram _3.cram

    PASS: both outputs match, despite warning
    "[bam_merge_core] Order of targets in file _2.cram caused coordinate sort to be lost"

Test: mpileup SAM
    samtools mpileup _1.sam _1.sam _1.sam
    samtools mpileup _1.sam _2.sam _3.sam

    PASS: both outputs match

Test: mpileup CRAM
    samtools mpileup _1.cram _1.cram _1.cram
    samtools mpileup _1.cram _2.cram _3.cram

    FAIL: outputs differ, plus warning
    "[E::bam_plp_push] The input is not sorted (chromosomes out of order)"


If we sort the CRAM files such that the order of records matches the order of the header, things start to fail:

    samtools sort --write-index --reference ref.fa -o _1s.cram _1.sam

Test: merge sorted SAM/BAM/CRAM
    samtools merge _1s.sam _1s.sam _1s.sam
    samtools merge _1s.sam _2s.sam _3s.sam

    All 3 FAIL

Test: mpileup sorted SAM
    samtools mpileup _1s.sam _1s.sam _1s.sam
    samtools mpileup _1s.sam _2s.sam _3s.sam

    FAIL: hard error
    "[E::bam_plp_push] The input is not sorted (chromosomes out of order)"

Test: mpileup sorted BAM
    samtools mpileup _1s.bam _1s.bam _1s.bam
    samtools mpileup _1s.bam _2s.bam _3s.bam

    FAIL: no warnings, but outputs differ in order with the same data appearing in all 3 references.
    It does output the exact same number of base calls though, so nothing is lost, just misassigned.

Test: mpileup sorted CRAM
    samtools mpileup _1s.bam _2s.bam _3s.bam
    samtools mpileup _1s.cram _2s.cram _3s.cram

    FAIL: but identical to BAM output, so CRAM isn't the issue here.


So we have a bug in that mpileup doesn't spot sort orders differing.  It could in theory correct this, but it's a different issue and not this PR.

bam_sort.c
```diff
@@ -1103,19 +1105,17 @@ int bam_merge_core2(int by_qname, char* sort_tag, const char *out, const char *m
...
-        // Potential future improvement is to share headers between CRAM files for
-        // samtools sort (where all headers are identical.
-        // Eg:
-        //
-        // if (i > 1) {
-        //     sam_hdr_free(cram_fd_get_header(fp[i]->fp.cram));
-        //     cram_fd_set_header(fp[i]->fp.cram, cram_fd_get_header(fp[0]->fp.cram));
-        //     sam_hdr_incr_ref(cram_fd_get_header(fp[0]->fp.cram));
-        // }
+        if (!refs && cram_get_refs(fp[i]))
```
valeriuo, line 1114: `if (!refs)` should be enough.

bam_sort.c
valeriuo: +1 for this

valeriuo (approved):

bam_sort.c
jkbonfield: Agreed.  Not sure what I was thinking there.

jkbonfield:
Squashed some more changes into the last commit.  It now calls `goto fail` a lot more in the bam_merge_core2 function, fixing many of the historic lackings as well as a new one added by me in this PR.  Also tidied up the refs setting code.  I think this addresses all comments to date.

valeriuo:
Merged as 0e4de419e4c34bb7fc01b8a47dc8f7878741bbbe and 589b8d2898e8a0ad063ad82029b6abdbd364f12b

PR #472, opened 2015-09-25
Threads=@

jkbonfield:
Made --threads long option versions of -@ to be consistent with htslib#272.

This has also lead to some reformatting of the usage help to keep things tabular.


jmarshall:
Thanks.  Applied in rebased and squashed form as 803143bf2e7004349d1797ca4d17b0b1b355f98b.

PR #483, opened 2015-11-09
Update merge -cp usage

### LINKED ISSUE #482: minor: samtools merge usage message on '-p' option
The documentation of the -p option to samtools merge is obscure:
> combine PG tags with colliding IDs rather than amending them

When I do a little test run it turns out that what actually happens is

> keep the PG tags of just the first of the files being merged, rather than concatenating those of all

(The fact that in the default case the IDs are made artificially unique by appending a random string is trivial). Or am I overlooking something?

mp15:
Update merge -cp usage to clarify we use the first line with a given ID rather than merging the actual lines (fixes #482)


jmarshall:
Thanks.  Somewhat reworded (in particular, fixed RG/PG typos) and applied as fe1868b084cff67a3f5afb036a6b6ba4942dca7a.

PR #486, opened 2015-11-13
document "*" region query for unmapped read pairs

mcshane:
Not sure of the wording, but we should document this previously undocumented feature. Not sure how many times I've told people about this.


jmarshall:
How about this wording?

mcshane:
Much better :+1:

PR #488, opened 2015-11-19
add --md (min_depth) option to ignore the position without sufficient…

shengqh:
For somatic mutation call, only the position with sufficient reads covered is useful. With this new option "--md (min_depth)", the mpileup will ignore the position without sufficient reads covered and then accelerate the analysis followed by,


valeriuo:
Is this feature something we want to add to the next release? @pd3

pd3:
Thank you for the pull request. I don't think we want to add and maintain this feature, piping through `awk` is simple enough and gives the same functionality.

Sorry it took us so long to respond.

PR #490, opened 2015-11-23
Use O_EXCL when writing temporary merge files

### LINKED ISSUE #432: `samtools sort` clobbers temporary files if "misusing" `-T`
Providing a duplicate prefix, or incorrectly specifying a directory for `-T` when performing simultaneous execution of `samtools sort` causes temporary files to be clobbered, leading to incorrect output and errors.

### Reproduce
Execute two sorts providing a duplicate prefix or duplicate directory path to `-T`:
```
samtools sort -m1K -O sam -T tmp/ -o output1.sam input1.sam
samtools sort -m1K -O sam -T tmp/ -o output2.sam input2.sam
```
### Result
* output1.sam contains input2.sam's header along with read groups appearing in input2.sam and potentially some from input1.sam.
* output2.sam is not created as the job aborts.

### Behaviour
* Job A begins executing, creating temporary files in `tmp/` with the name `.0000.tmp` to `.NNNN.tmp`.
* Job B begins, chasing Job A, clobbering temporary files in `tmp/` from `.0000.tmp` to `.MMMM.tmp`.
* Job A completes sorting temporary files and merges `.0000.tmp` to `.NNNN.tmp`. Thus incorrectly including read groups from Job B in place of the desired read groups from Job A where N <= M).
* Job A completes merging and deletes temporary files.
* Job B crashes attempting to read `.0000.tmp` as it no longer exists.

### Solution
* `samtools sort` should attempt to check whether a `.tmp` file exists already and either issue a sensible warning or abort.

SamStudio8:
* Append 'x' to the open mode string for temporary files handled by
  merge workers to prohibit a second sort clobbering the temporary
  files of a first sort.
* Fixes #432.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1554,7 +1554,7 @@ static void *worker(void *data)
     ks_mergesort(sort, w->buf_len, w->buf, 0);
     name = (char*)calloc(strlen(w->prefix) + 20, 1);
     sprintf(name, "%s.%.4d.bam", w->prefix, w->index);
-    write_buffer(name, "wb1", w->buf_len, w->buf, w->h, 0, NULL);
+    write_buffer(name, "wb1x", w->buf_len, w->buf, w->h, 0, NULL);
 
 // Consider using CRAM temporary files if the final output is CRAM.
 // Typically it is comparable speed while being smaller.


jkbonfield:
Note, the total lack of error checking in samtools sort means that this doesn't fix it.  Having failed to open, it happily then fails to write all the records followed by merging in the temporary BAM files at the end.  So it still gives gibberish.

See #530 for a more complete implementation.

SamStudio8:
@jkbonfield Woops, you're right. I tested this previously: https://github.com/samtools/samtools/issues/432#issuecomment-158959036 and was just happy to see some form of error printed (rather to quickly alert the user that *something* bad has happened and that all your science is ruined, as opposed to actually trying to prevent it from happening on the first input).

Closing in favour of #530.

PR #492, opened 2015-11-24
Don't need sudo in TravisCI so can use container-based infrastructure

peterjc:
See https://docs.travis-ci.com/user/workers/container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 # Control file for continuous integration testing at http://travis-ci.org/
 
+sudo: false
+
 os:
   - linux
   - osx


jmarshall:
See also discussion in #435.  Personally I'd rather just wait for Travis to obsolete their old infrastructure and change their default than annotate our config files…

peterjc:
If you've heard word that TravisCI is going to change the default then fair enough. I've been using ``sudo: false`` in several projects without problems.

(The only potential snag is the other way round - projects still using ``sudo apt-get install ...`` or similar if the defaults change)

valeriuo:
The default environment is container-based as of August 2017.

PR #498, opened 2015-12-08
Prevents unintended free in finish_merged_header when n_targets == 0

### LINKED ISSUE #495: samtools sort crashes
samtools has been very shaky recently - it crashes when sorting almost any BAM file. Here is how it goes:

samtools sort SATP_86.bam -o foo
[bam_sort_core] merging from 32 files...
*** Error in `samtools': double free or corruption (!prev): 0x0000000032b611b0 ***
Aborted (core dumped)

I've had similar problems for a month or two. I update both samtools and htslib regularly and compile them "as is". Version 1.2 does not have this issue.

daviesrob:
finish_merged_header reallocs the target_name and target_len arrays to
be exactly n_targets in size.  This caused a problem when n_targets == 0
as realloc(ptr, 0) frees ptr.  To stop this from happening, make sure we
always allocate at least one element.

Fixes #495

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -860,6 +860,7 @@ static bam_hdr_t * finish_merged_header(merged_header_t *merged_hdr) {
     char      *text;
     char     **target_name;
     uint32_t  *target_len;
+    size_t     targets_sz = merged_hdr->n_targets ? merged_hdr->n_targets : 1;
     bam_hdr_t *hdr;
 
     // Check output text size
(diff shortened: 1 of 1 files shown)


jmarshall:
Thanks for tracking this down!  I think the bug is more about the interplay between those lines and the later error-handling lines

```c
hdr->target_name = target_name ? target_name : merged_hdr->target_name;
hdr->target_len  = target_len ? target_len : merged_hdr->target_len;
```

which are wrong when `realloc(merged_hdr->target_name,0)` frees `merged_hdr->target_name` and returns NULL.  (Which is a glibc behaviour not shared by BSD libc btw…)  So I fixed it differently in 90dcd90135c7fd134e10b243f646bfb24b600cc0.

PR #504, opened 2015-12-18
Fixed typo in README.

blaiseli:
There was a dot instead of a space in the `aclocal` command.


jmarshall:
Actually that's intended as an ellipsis, thus an abbreviated form of

    aclocal -I …wherever/you/installed…/autoconf-archive/m4

Issue #364, opened 2015-03-16
More powerful samtools view flag filtering


jmarshall:
From a user:

> Please can you add functionality to allow removal of neither end mapped read pairs using samtools view? E.g by adding an option to filter reads for which all bits are set as well as the current -F for any bits set.

We should be able to find some way to do this by adding optional `+` or `-` prefixes to `-f`/`-F`, akin to find(1)'s `-flags` and `-perm`.

jkbonfield:
Fixed by merging PR #505.

PR #505, opened 2015-12-18
Option -G to exclude reads based on the presence of a full set of bits in the SAM flag.

blaiseli:
I wanted to be able to exclude reads having a given combination of bits set in he SAM flag.
Did I miss how to do it with -f and -F ?

Anyway, I tried to add a new option for that. For instance, to exclude reads where both mates are unmapped, use `samtools view -G 12`

I was unable to compile samtools. I had managed with the non-git version 1.1, but the structure of the `sam_view.c` file has changed in the git version, so the present pull request is untested code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ committed to this repository, so building the code from a Git repository
 requires some extra steps:
 
 ```sh
-aclocal -I.../autoconf-archive/m4   # Cache configure subroutines
+aclocal -I ../autoconf-archive/m4   # Cache configure subroutines
 autoconf       # Generate the configure script, if needed
 ./configure    # Optional, needed for choosing optional functionality
 make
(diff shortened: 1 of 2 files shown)


jrandall:
Related to user request #364

jkbonfield:
I just got around to looking at this and fixed a bug in the bam2fq (it needed a check for ```b->core.flag&(state->flag_alloff) != 0``` too), but otherwise am happy with it.  I'll update the samtools.1 man page too.

One thing though is the usage phrasing is convoluted.  Currently we have:

  -f INT   only include reads with all bits set in INT set in FLAG [0]
  -F INT   only include reads with none of the bits set in INT set in FLAG [0]
  -G INT   only include reads with not all the bits set in INT set in FLAG [0]

I propose:

  -f INT   only include reads with all of the bits set in INT set in FLAG [0]
  -F INT   only include reads with all of the bits unset in INT set in FLAG [0]
  -G INT   only include reads with any of the bits unset in INT set in FLAG [0]

The man page will also need to clarify this with an "and not zero" statement.  Any comments on the above wording?  Does it actually mean the same thing? (I think so, but the office doesn't have consensus on this point!)

jkbonfield:
Rebased and merged as b1b6717e7017e5808560b9f808726049f7485f41

Also fixes #364

PR #507, opened 2015-12-20
samtools sort: don't free BAM-related buffers

nathanweeks:
samtools sort optimization for sorting large bam files: don't explicitly free
BAM-related buffers before exiting.

Profiling a samtools sort of a 13GB BAM file that was constructed to minimize
I/O (allocating enough memory to avoid temporary file creation, writing output
to /dev/null) revealed that ~18% of the run-time was spent freeing memory
at the end of bam_sort_core_ext(). Specifically, the comman invoked was:

   samtools sort -@ 16 -l 0 -m 3G -o /dev/null SRR1611184.bam

As "samtools sort" exits after calling bam_sort_core_ext(), it isn't necessary
to explicitly free the allocated memory. This patch adds a boolean argument to
bam_sort_core_ext() that determines whether or not the allocated memory will
be freed.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1620,6 +1620,7 @@ static int sort_blocks(int n_files, size_t k, bam1_p *buf, const char *prefix, c
   @param  max_mem  approxiate maximum memory (very inaccurate)
   @param  in_fmt   input file format options
   @param  out_fmt  output file format and options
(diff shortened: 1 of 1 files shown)


daviesrob:
Replaced by #593 (now merged as d39ed264)

PR #509, opened 2016-01-12
faidx with long opts + output filename + line length

lindenb:
some Minor changes in faidx:
- added support for long_opt
- added option -o|--output (default:stdout)
- added option -n|--length to specify the length of the lines in the fasta output


whitwham:
Rebased and merged via 1e23e1c

tseemann:
@lindenb thanks for this!  `fadix` needed some "TLC".

I am surprised that `putchar()` was used. I guess it's a macro/inline and uses stdio buffering so shouldn't be too inefficient.

PR #516, opened 2016-01-27
Extended view -x to handle tag-lists and also added a whitelist equivalent.

jkbonfield:
Both have long options now too.

NOTE: We may wish to consider making some functions from htslib public.  See the aux_type2size and skip_aux functions below.  These exist in htslib and are needed in samtools, but are duplicated here until we have a discussion on what public API we wish to expose (these or a derivative?).

Example uses:

  samtools view -x BD -x BI -x BQ in.bam
  samtools view -x{BD,BI,BQ} in.bam;               # equivalent via shell
  samtools view -x BD,BI,BQ in.bam;                # equivalent
  samtools view --tag-blacklist=BD,BI,BQ in.bam;   # equivalent

  samtools view --tag-whitelist BD,BI,BQ;          # opposite of above

### DIFF
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -362,7 +362,19 @@ in octal by beginning with `0' (i.e. /^0[0-7]+/), as a decimal number
 not beginning with '0' or as a comma-separated list of flag names.
 .TP
 .BI "-x " STR ", --remove-tag " STR
-Read tag to exclude from output (repeatable) [null]
+Read tag(s) to exclude from output (repeatable) [null].  This can be a
+single tag or a comma separated list.  Alternatively the option itself
+can be repeated multiple times.
+
+If the list starts with a `^' then it is negated and treated as a
+request to remove all tags except those in \fISTR\fR. The list may be
+empty, so \fB-x ^\fR will remove all tags.
+.TP
+.BI --keep-tag " STR
+This keeps \fIonly\fR tags listed in \fISTR\fR and is directly equivalent
+to \fB--remove-tag ^\fR\fISTR\fR.  Specifying an empty list will remove
+all tags.  If both \fB--keep-tag\fR and \fB--remove-tag\fR are
+specified then \fB--keep-tag\fR has precedence.
 .TP
 .BR -B ", " --remove-B
 Collapse the backward CIGAR operation.
--- a/sam_view.c
+++ b/sam_view.c
@@ -45,6 +45,11 @@ DEALINGS IN THE SOFTWARE.  */
 KHASH_SET_INIT_STR(str)
 
 typedef khash_t(str) *strhash_t;
+KHASH_SET_INIT_STR(rg)
+KHASH_SET_INIT_INT(aux_exists)
+
+typedef khash_t(rg) *rghash_t;
+typedef khash_t(aux_exists) *auxhash_t;
 
 // This structure contains the settings for a samview run
 typedef struct samview_settings {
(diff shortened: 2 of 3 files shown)


jkbonfield:
Looking at bcftools it uses --include and --exclude regularly, so it's prefereable to use --include-tags and --exclude-tags instead of --tag-whitelist and --tag-blacklist.
                                                                                                               
The short option though is less clear.  Eg bcftools view has -i / --include and -e / --exclude, but also -p, --phased and -P, --exclude-phased, plus -u, --uncalled and -U, --exclude-uncalled.

That implies it prefers lower-case for include and upper-case for exclude, with the include word removed.  For samtools that would be -x, --tags and -X, --exclude-tags, however we already have used -x for exclude tags.  Basically I wouldn't start from here, but we're here so we're stuffed.  How about -x, --exclude-tags and --include-tags (no shorthand)?

jkbonfield:
Rebase and corrected the many conflicts.

In the intervening 5 years the whole debate about `--tag-blacklist` vs `--exclude-tags` is now a moot point as we added a long opt of `--remove-tag`.  So I renamed the whitelist option as `--keep-tag`.

I also decided to include the bcftools style `^LIST` option for brevity given there is no short option for `--keep-tag`.

doc/samtools-view.1
```diff
@@ -362,7 +362,16 @@ in octal by beginning with `0' (i.e. /^0[0-7]+/), as a decimal number
...
-Read tag to exclude from output (repeatable) [null]
+Read tag(s) to exclude from output (repeatable) [null].  This can be a
+single tag or a comma separated list.  Alternatively the option itself
+can be repeated multiple times.
+
+If the list starts with a `^' then it is negated and treated as a
+request to remove all tags except those in \fISTR\fR.
+.TP
+.BI --keep-tag " STR
+This is keeps \fIonly\fR tags listed in STR and is directly equivalent
```
jmarshall, line 373:
```suggestion
This keeps \fIonly\fR tags listed in \fISTR\fR and is directly equivalent
```

jmarshall:
In the intervening 5 years, another whole debate about names like `--tag-blacklist` has lead to a welcome raised level of consciousness in our field (along with a few overshoots, of which this is not one IMHO). It would probably be worth revisiting this PR's `samview_settings_t` field names too — and then they likely wouldn't need a comment. :smile:

jkbonfield:
A fair comment.  I'll do some variable hackery.  Hopefully without a trillion force pushes this time to correct stupid errors added at the last minute. :)

jkbonfield:
I can confirm this fixes #871 by `-x ^` or `--keep-tag ""`.

sam_view.c
```diff
@@ -68,13 +73,58 @@ typedef struct samview_settings {
...
+    case 'Z':
+    case 'H':
+        while (*s) ++s;
+        return s + 1;
+    case 'B':
+        size = aux_type2size(*s); ++s;
+        memcpy(&n, s, 4); s += 4;
+        return s + size * n;
+    case 0:
+        abort();
```
daviesrob, line 121: This is a bit of a harsh way of telling the user they have an invalid file.  Printing a suitable error message (preferably saying which read caused the problem) and calling exit would be nicer.

sam_view.c
```diff
@@ -68,13 +73,58 @@ typedef struct samview_settings {
...
+        return 8;
+    case 'Z': case 'H': case 'B':
+        return type;
+    default:
+        return 0;
+    }
+}
+
+// Copied from htslib/sam.c.
+static inline uint8_t *skip_aux(uint8_t *s)
```
daviesrob, line 107: The [version of this in HTSlib](https://github.com/samtools/htslib/blob/4811c2d8/sam.c#L4093) is careful not to fall off the end of the BAM record.  Was this from an earlier, less safe version?

The comment is correct, an HTSlib interface to iterate through the aux tags would be useful.  Or for a future update we could consider adding something to do the aux tag editing part of this PR as a handy higher-level function.

sam_view.c
```diff
@@ -286,6 +361,31 @@ static inline void change_flag(bam1_t *b, samview_settings_t *settings)
...
 }
 
+int parse_aux_list(auxhash_t *h, char *optarg) {
+    if (!*h)
+        *h = kh_init(aux_exists);
+
+    while (strlen(optarg) >= 2) {
+        int x = optarg[0]<<8 | optarg[1];
+        int ret = 0;
+        kh_put(aux_exists, *h, x, &ret);
```
daviesrob, line 371: Ideally `ret` should be checked here, although you probably won't get far if you're out of memory this early...

doc/samtools-view.1
```diff
@@ -362,7 +362,16 @@ in octal by beginning with `0' (i.e. /^0[0-7]+/), as a decimal number
...
 .TP
 .BI "-x " STR ", --remove-tag " STR
-Read tag to exclude from output (repeatable) [null]
+Read tag(s) to exclude from output (repeatable) [null].  This can be a
+single tag or a comma separated list.  Alternatively the option itself
+can be repeated multiple times.
+
+If the list starts with a `^' then it is negated and treated as a
+request to remove all tags except those in \fISTR\fR.
+.TP
```
daviesrob, line 371: It might be good to specifically document that `-x ^` or `--keep-tag ''` will remove all tags.  It may not be obvious to everyone that this is how you create an empty "keep" list.

It would also be good to say that `--keep-tag` overrides `--remove-tag` if you supply both.

sam_view.c
jkbonfield: This just comes from samtools/htslib@30b4033c and is just a demonstration of how ancient this PR is!

I'll copy the code from the latest htslib develop instead.

sam_view.c
jkbonfield: See above - it's just that Htslib has moved on in the last 6 years since I copied these functions over.

I could create an htslib PR to make them public, but I don't particularly wish to slow up merging even more than it's already taken.

sam_view.c
jkbonfield: Added, although I'm not bothering to report an error.  Frankly if your system is so stressed out it can't even parse command line arguments correctly then I think you have bigger problems than worrying about the lack of a nice message! (Plus it'll be *really* obvious in a myriad of other failures.)

jmarshall:
(I don't know if Git still has the historical limitation whereby if multiple branches point to the same place as `origin/HEAD` then new clones don't know which branch (e.g. `develop` or `tag_whitelist`) to check out and choose randomly… but as it's in the main **samtools** repository it'd be more than usually good to delete the `tag_whitelist` branch now that this has been merged.)

Issue #517, opened 2016-01-28
fastq segfault with -O option


mcshane:
```
samtools fastq -O test/dat/view.001.sam
Segmentation fault
```
Looks like it came in with this change (b73d36e9281a9c23829f0cee22f9a9af184d1899)
```
-    if (state->use_oq) oq = bam_aux_get(b, "OQ");
+    if (state->use_oq) oq = bam_aux_get(b, "OQ") + 1;
```
If there is no OQ on a line, then later `if(oq)` statements are true because of the `+1`.

PR #519, opened 2016-02-01
Fix -O option for reads without OQ tag #517

mp15:
When using the -O option the +1 meant our NULL check was failing. Fix to only populate oq variable where there is an OQ tag present.


jmarshall:
Updated to avoid calling `bam_aux_get` twice in the usual case.

PR #524, opened 2016-02-05
Added bz2 and lzma to Makefile (mandatory).

jkbonfield:
See also https://github.com/samtools/htslib/pull/331

If people want to avoid requiring these, eg building samtools without
CRAM support, they'll need to use the configure script.

TODO: Get configure script working via some form of pkg-config
approach so it knows which libraries to link against.  This is a
larger issue (it doesn't work for curl or irods either) and so
not addressed in this PR.


mp15:
This is failing on Mac due to lack of lzma.h by default might need to add something to deal with that

jkbonfield:
I don't think this failure is anything to do with Travis MacOS X lacking lzma in the current configuration (besides autoconf is meant to spot and avoid that for us so it wouldn't be a failure).  

Rather it's the standard problem we've had all along.  Samtools/configure branch requires htslib/configure branch, but one of our travis jobs tries samtools/configure against htslib/develop.  Every time I've submitted paired pull requests like this they fail, for obvious reasons.  Typically I just check one of the jobs passed (1 out of 2 above did) and know by experience that that means it's essentially working once the matched PR is merged in.

jmarshall:
Seems a lot like it to me (and the travis builds don't currently run configure)…

> cram/cram_io.c:60:10: fatal error: 'lzma.h' file not found
> #include \<lzma.h\>

The **push** build gets the companion **htslib/_configure_** branch (and failed as above).  The **pr** build repurposes the relevant travis environment variable and gets **develop** (and succeeded as neither htslib or the samtools test suite has any lzma).  Good work, Travis!

jkbonfield:
Oh so this is actually the reverse of the normal error - where we get one succeed and one failure due to using the same named branch and using develop - only here it fails on the identical branch and succeeds on the develop branch? That's kind of obfuscated if so.   The cry wolf mentality just makes me expect one pass and one fail as the same thing as success.   Is there any reason to do both instead of the appropriate name branch? (Ie _foo_ if exists, _develop_ if not?)  I don't know how configurable travis is.

Either way it's not a macosx issue, but it is a travis config one I guess.

jmarshall:
Feel free to reread _repurposes the relevant travis environment variable_, dive into the relevant Travis documentation, and file a PR that improves our travis setup.

The **Mac OS X issue** here is that OS X does not typically come with LZMA development files, and they're not just an `apt-get` away.  Travis failing is just a demonstration of this.  What we have here is a dependency on something that's not particularly common on this platform.

jkbonfield:
So how do we resolve the macos x issue?  CRAM natively has support for lzma, so it is a strong recommendation that htslib on macs has lzma.  Ie it's default and you have to actively disable it.

Is this basically pointless as Makefile will be retired?  I don't know what your plans are regarding Makefile vs configure.

jkbonfield:
Likewise the whole issue of our travis setup not using configure script.  Why is this?  Is it deliberate policy, meaning a pull request to fix travis would be rejected, or just history?

jmarshall:
> Likewise the whole issue of our travis setup not using configure script.

This is basically just history, as our `.travis.yml` predates our `configure.ac`.

jkbonfield:
Ok, I'll try and delve into the murky world of travis and see if I can get it using configure, although I just realised this is samtools and not htslib issue so sorry for the query in the wrong spot.   (I'll also look at figuring out a more appropriate --with option at the same time.)

jkbonfield:
Deprecated.

PR #530, opened 2016-02-10
Made samtools sort notice when it clobbers other sorts.

### LINKED ISSUE #523: Default samtools sort temp-prefix leads to data corruption when reading from stdin
Usually I like to pipe alignments until I have a sorted bam file. Like this:
``` 
bwa mem ${assembly} ${fastqFile} | samtools view -bSF 4 - | samtools sort -o ${outputBase}.bam -
```

However with v1.3 this fails when creating multiple bam files in the same directory, because samtools sort seems to use the input file name '-' as PREFIX for the temporary files.  So different parallel samtools instances run into conflict with each other since all temporary files are named
```
-.tmp.0001.bam
-.tmp.0002.bam
-.tmp.0003.bam
```

Imho in case of - as input, samtools should rather use the output name (-o) as prefix or if this one is not defined either (and the users just want to write to stdout) `sort` should require the user to specify a prefix with -T.

I think to recall that this was working correctly with earlier versions of samtools.

jkbonfield:
The temporary filenames still aren't unique; the same filenames
generated when opening and then again when unlinking, so fixing this
properly requires larger code reorganisation.

Instead I made the minimal change to open with exclusive mode,
followed by the inevitable code to actually check the open, writes and
closes (none of it present before - gah!).

Fixes #523

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1617,6 +1617,8 @@ typedef struct {
     int error;
 } worker_t;
 
+// Returns 0 for success
+//        -1 for failure
 static int write_buffer(const char *fn, const char *mode, size_t l, bam1_p *buf, const bam_hdr_t *h, int n_threads, const htsFormat *fmt)
 {
     size_t i;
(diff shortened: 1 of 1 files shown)


jmarshall:
Rebased to current develop post-PR-#467, which addressed many of the same missing error checks.  (Previously **sort_checks** was e5ce135bf3a9ac262c88b062f4f315b7a34ebd21.)

PR #534, opened 2016-02-12
[WIP] Adding option for computing shared k-mers with contaminants

holtgrewe:
This is useful for screening aligned BAM files for shared k-mers (10 turned out to work well for me) with Illumina adapters, for example.

The observed/expected count could be improved. I'll think about how to get a better approximation here but I don't know whether I will succeed.

Please tell me what you think about this.


jkbonfield:
I haven't studied it in detail, but we agreed at our last samtools meeting that the basic premise of the work is justified and it is an appropriate option to add to stats.

The brief glance I've given of the code looks OK.   Please let us know when it's no longer considered a work in progress.

Thanks.

jkbonfield:
Sorry it's been a while - we've just been going through some of our old PRs tidying things up and this has been labelled WIP for some time.  Is it something that is still required, and if so is there a new version?  (If not we'll close the PR in a week or two.)

Thanks.

holtgrewe:
I think I'll not be able to continue working on this. Sorry.

PR #536, opened 2016-02-15
Add check on hts_close to quickcheck

mp15:
Add check on hts_close to quickcheck command, we are missing some errors here.
Also add expected fail test to test check for EOF block on CRAM.


jmarshall:
\<broken_record>Did you test this?\</broken_record>

jmarshall:
Thanks.  Added also to quickcheck/all test.

PR #537, opened 2016-02-16
Added mpileup -a and -aa options; fixes #496.

### LINKED ISSUE #496: Suggestion: optional output for zero-coverage positions in samtools mpileup
At the moment, mpileup skips positions which don't have any reads mapping to them.

However, if you want to use mpileup output downstream in custom scripts, having lines output for positions with zero coverage becomes important. E.g. if you want to compare separate mpileup outputs for different mappings to the same reference.

Could you implement a -a and -aa option like you have for samtools depth?

For zero-coverage positions, the base and quality columns should ideally consist of empty tabs, or some other placeholder such that the number of columns in each line is kept consistent.

jkbonfield:
Also fixed the tests for c1#pad1.bam and c1#pad2.bam; they should have
been known failures instead of know passes (see issues
samtools/htslib#57 and samtools/htslib#59).

[NEWS]
Mpileup has gained -a and -aa options to output all coordinates
including those with zero coverage.  Single -a limits this to
reference sequences containing some coverage with double -a (-aa or -a
-a) also displays references with zero reads mapped to them.


jkbonfield:
Sorry the commit message is linking to samtools 57/59 instead of htslib; minor glitch.

jkbonfield:
I should add, this changes the syntax of mpileup text output to be more self consistent, but therefore is **incompatible** with the previous output if you attempt to parse it in certain ways.

Specifically, mpileup sometimes outputted "0\t*\t*" for zero coverage (zero coverage in one file but coverage in another) and other times "0\t\t" (zero coverage due to failing the minimum base quality check).

Given the rather arbitrary mix of styles and the fact that "0\t\t" is problematic to parse in some tools, I made it all self consistent.   YOU HAVE BEEN WARNED! :-)

jmarshall:
Rebased, refactored, and a couple of bugs fixed (`\n` printed to stdout instead of `pileup_fp`; combination of `-a -g` crashed).

PR #539, opened 2016-02-17
Fixed bam_get_library() crash.

### LINKED ISSUE #538: bam_get_library segfaults if read groups lack LB tags
On line 114 of bam.c (https://github.com/samtools/samtools/blob/74a3e0a744f91bc2b1f65a4144a509d6dd791631/bam.c#L114), `bam_get_library()` looks for `LB` tags in `@RG` lines. That's all well and good if they're there, but apparently some people are adding read groups that don't contain all of the normal tags (cf., here http://seqanswers.com/forums/showthread.php?t=66333, which was subsequently posted to the samtools-help list here https://sourceforge.net/p/samtools/mailman/message/34857333/).

I presume that a simple `if(LB) {` around line 112 would solve this, though it'd probably be good to print a warning to alert the user that the read group should really have an `LB` tag.

jkbonfield:
This was triggerable from samtools rmdup or samtools view -l when an
@RG header was present without an LB: record.

I also fixed the samtools view -l logic.  The man page states that it
only shows reads in that library, while the implementation was to only
show reads in that library or those where it could not determine the
library.

Fixes #538

### DIFF
--- a/bam.c
+++ b/bam.c
@@ -105,6 +105,9 @@ const char *bam_get_library(bam_header_t *h, const bam1_t *b)
             last = *cp++;
         }
 
+        if (!ID || !LB)
+            continue;
+
         // Check it's the correct ID
         if (strncmp(rg, ID, strlen(rg)) != 0 || ID[strlen(rg)] != '\t')
             continue;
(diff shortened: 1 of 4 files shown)


dpryan79:
:+1: Thanks for the quick fix @jkbonfield!

PR #540, opened 2016-02-17
Convert quickcheck to hts_check_EOF API to check EOF on CRAM

mp15:
Convert quickcheck to hts_check_EOF API to check EOF on CRAM files and avoid using BGZF internal APIs.
Depends on samtools/htslib#339


mp15:
Updated with a more diverse set of tests as suggested by @jkbonfield

bam_quickcheck.c
```diff
@@ -26,7 +26,6 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include <htslib/hts.h>
 #include <htslib/sam.h>
-#include <htslib/bgzf.h>
```
jmarshall: Also remove from bam_quickcheck.o dependency in the Makefile.

jmarshall:
Please rebase, and:

* lift the _check EOF on formats that support this_ block up a level so it operates on non-`sequence_data` files too.
* Handle all the codes returned by `hts_check_EOF()`; in particular, at the moment for _foo.v2.0.cram_ (code 3) it prints "foo.v2.0.cram has good EOF block" which is at best misleading.
* Apply the makefile tweak noted.
* While you're at it, fix the `bam_hdr_t *header` memory leak just above here, and handle `sam_hdr_read()` returning NULL.

mp15:
Updated as recommended

PR #545, opened 2016-03-14
Add MC support to fixmates

### LINKED ISSUE #170: fixmate should fix / optionally add MC tags
**MC** is a recent addition to the list of well-known tags that provides the mate's CIGAR string.  (See samtools/hts-specs#19)

Fixmate should also fix this new tag if necessary, and perhaps have an option to add it.

mp15:
Add mate cigar tag support to fixmate and fix tests. (Fixes #170)
bam_cigar_to_string may be a candidate to move to htslib.


jmarshall:
Should this be enabled by a command-line option?

What about unmapped reads or otherwise `*` CIGARs?

Change the `bam_cigar_to_string` signature to something like `void bam_format_cigar(const bam1_t *b, kstring_t *str)`, and 1) it'll be more of a candidate for htslib; 2) you really won't need `restrict`, which apparently not-so-outdated compilers dislike :disappointed:; 3) you won't leak `mc`; 4) ~1 less memory allocation for every alignment record.

jkbonfield:
We were discussing the validity of MC:Z:* today in the office, with respects to a recent biobambam bug producing MC:Z: strings.  Is it better to not have MC tag at all or to add MC:Z:*?  It's a tricky one.  Technically * is correct, but it's probably over the top.  If we do want to do this then please consider only writing out * if at least one read is mapped (NB: may have more than 2 reads in total with this name), as it's pointless to add reams of MC:Z:* for all the unmapped read pairs.

Would the programs out there parsing MC tags choke on a *?  Hopefully not, but let's test them anyway.  I'd like an explicit check for MC * in the test data set too.  I only glanced over it quickly in github so I may have missed it in the full file, but if it's not there please add one.

jmarshall:
Actually I'm in two minds about controlling this via an option: if we think MC is useful, there is a case for just adding it (where appropriate, e.g., not on those reams of both-unmapped read pairs) unconditionally.

mp15:
I've updated this to modify the existing test case to add the situation where one read is mapped and it's mate is not. I've also converted the cigar to string code to John's suggestion of `void bam_format_cigar(const bam1_t *b, kstring_t *str)`.

bam_mate.c
```diff
@@ -154,16 +154,41 @@ static bool plausibly_properly_paired(bam1_t* a, bam1_t* b)
         return false;
 }
 
-static void sync_mq(bam1_t* src, bam1_t* dest)
+static void bam_format_cigar(const bam1_t* b, kstring_t* str)
+{
+    const uint32_t *cigar = bam_get_cigar(b);
+    uint16_t i;
+
+    for (i = 0; i < b->core.n_cigar; ++i) {
```
mp15: Should this return * for unmapped?

mp15:
Modified based on feedback and rebased to current trunk

bam_mate.c
```diff
@@ -155,9 +155,27 @@ static bool plausibly_properly_paired(bam1_t* a, bam1_t* b)
...
+    if (b->core.n_cigar == 0) {
+        kputc('*', str);
+        return;
+    }
+
+    const uint32_t *cigar = bam_get_cigar(b);
+    uint16_t i;
+
+    for (i = 0; i < b->core.n_cigar; ++i) {
+        kputw(bam_cigar_oplen(cigar[i]), str);
```
jkbonfield: Kstring since https://github.com/samtools/htslib/commit/67234cb0115001eb92909726fc7259bf1806377a now returns error codes, which means  bam_format_cigar() should check the return value from kputw and kputc.  If feels a bit churlish to point this out given this also means sync_mq_mc should return a value, which in turn means sync_mate should return a value too, and they aren't entirely new functions added by this PR.  However as other changes  would be useful I'll mention this here too. :-)

bam_mate.c
```diff
@@ -155,9 +155,27 @@ static bool plausibly_properly_paired(bam1_t* a, bam1_t* b)
...
+static void bam_format_cigar(const bam1_t* b, kstring_t* str)
+{
+    // An empty cigar is a special case return "*" rather than ""
+    if (b->core.n_cigar == 0) {
+        kputc('*', str);
+        return;
+    }
+
+    const uint32_t *cigar = bam_get_cigar(b);
+    uint16_t i;
```
jkbonfield: This should be uint32_t as although uint16_t is sufficient for current BAMs, we're already bumping into this limit (and CRAM can avoid the 16-bit size limit).

bam_mate.c
```diff
@@ -166,6 +184,18 @@ static void sync_mq(bam1_t* src, bam1_t* dest)
 
         bam_aux_append(dest, "MQ", 'i', sizeof(uint32_t), (uint8_t*)&mq);
     }
+    // Always copy Mate Cigar it can have meaning even in unmapped reads (hard clipping etc)
```
jkbonfield: I'm not convinced on the usefulness of always adding this.  It may be helpful when one read is mapped and the other is "placed" (ie it is unmapped, but has chr/pos along side the mate), but I don't see this as being useful when both ends are unmapped, with chr \* and pos 0.

Right now it appears that such data gets MC:Z:\* added.  Is that deliberate?

mcshane:
@mp15 Where are we at with this PR? Do we want it for 1.4?

mp15:
Updated the PR based on @jkbonfield feedback, rebased and added error handling.

bam_mate.c
```diff
@@ -156,24 +156,28 @@ static bool plausibly_properly_paired(bam1_t* a, bam1_t* b)
         return false;
 }
 
-static void bam_format_cigar(const bam1_t* b, kstring_t* str)
+// Returns true on error, false on success
```
jkbonfield: I dislike the use of bool here, in particular true being error and false being success.

It looks like you're trying to return 0 (false) for success with non-zero being error in the traditional unix way, in which case it should be return 0 and not return false.  Using a bool just obfuscates things.

If you don't object I'll fix this and push back to the PR for you to comment on.

mp15: I don't object but for some reason Github thinks you're commenting on an outdated changeset? No clue how that is even possible, I haven't touched it for a week.

mp15:
Good to merge now then?

Issue #551, opened 2016-04-01
samtools view -U


zfrenchee:
I'm unclear on how samtools view's -U option works. Is it necessary for a -o option to also be specified for -U to work? Why is it that this command fails? (and by fails I mean it does not filter out mitochondrial reads).

```
samtools view -U output.bam -b input.bam chrM > out.txt
```
out.txt. contains all the mitochondrial reads, but output.bam is empty. Why is that? 

I think the docs could be written a little better to explain the -U flag, which I would be happy to do once I understand it.

mp15:
You command does not do what you expect because of the way you specified your regions. In this mode of use chrM isn't a filter it instead an index seek. When we do index seeks we do not process the region from outside the region specified because that would take far too long. This means that when you use the -U command it only outputs reads within the regions specified for your iterator to walk over.

I suspect if you specified chrM as a bed file you will get the result you were looking for. BED files are processed in a streaming fashion and thus treated as a filter. The reason for this is that the expected use case was to use index seeks in the command line to specify locations in the genome to process (usually for parallelism) and BED files to specify subsets within the regions to process (such as exome baits).

jmarshall:
The man page documentation says

> -U _FILE_
> Write alignments that are _not_ selected by the various filter options to _FILE_.  When  this  option  is used,  all alignments (**or all alignments intersecting the _regions_ specified**) are  written  to  either the output file or this file, but never both.

(emphasis added)

Do you have a suggestion as to how this could be improved?

zfrenchee:
I think that a fair reading of that documentation, for a newbie like myself, is that specifying chrM in the command and -U should filter out chrM reads from the bam file. If that's only feasible by specifying regions in a BED file, then I think this warrants mentioning in the documentation, but if you don't feel the same, then perhaps at least this issue will prove useful for future newbies. 
Thanks both for helping me work through this.

PR #552, opened 2016-04-01
Whitespace fix in sam_view.c

zfrenchee:


mp15:
Patch appears to be a whitespace fix for sam_view.c, looks good to me.

zfrenchee:
Sorry for not giving a better description here, I just used github's built in "change this file" UI. 

If you could help me with #551 that's what I was working on when I decided to fix this.

jmarshall:
There is indeed an indentation problem here, but this is not the right fix.  Thanks, I've taken care of it.

PR #556, opened 2016-04-06
Fix bug caused by difference between BAM and SAM API

### LINKED ISSUE #532: samtools fasta/q generating empty files if option -s is used
Using samtools version 1.3

```
$ samtools view -c foo.bam 
11200
$ samtools fasta -1 foo_1.fa -2 foo_2.fa foo.bam 
[M::bam2fq_mainloop] processed 11200 reads
$ grep "^>" foo_1.fa | wc -l
5742
$ grep "^>" foo_2.fa | wc -l
5458
$ echo "5742 + 5458" | bc
11200
```

OK, so all my reads are there.

Now:
```
$ samtools fasta -1 foo_1.fa -2 foo_2.fa -s foo_s.fa foo.bam 
[M::bam2fq_mainloop_singletontrack] discarded 0 singletons
[M::bam2fq_mainloop_singletontrack] processed 0 reads
$ grep "^>" foo_2.fa | wc -l
0
$ grep "^>" foo_1.fa | wc -l
0
$ grep "^>" foo_s.fa | wc -l
0
```

What happened to my reads ?

mp15:
BAM returns different values to SAM and there was a silly mistake in the code that checked that value. (Fixes #532)

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -928,7 +928,7 @@ static bool bam2fq_mainloop_singletontrack(bam2fq_state_t *state)
 
     bool valid = true;
     while (true) {
-        at_eof = sam_read1(state->fp, state->h, b);
+        at_eof = sam_read1(state->fp, state->h, b) < 0;
 
         if (!at_eof && filter_it_out(b, state)) continue;
         if (!at_eof) ++n_reads;


jmarshall:
I will merge this (thanks!), but not with the message _Fix bug caused by difference between BAM and SAM API_ because that is not really true.  All of `sam_read1(SAM/BAM/CRAM)` return >=0 for success or <0 for failure; to be sure this is not documented for this function,¹ but it follows from the general negative ⇒ failure convention.

Change that first line to _Check sam_read1() return value correctly_ or similar and this will be fine.

Also it's a pity we failed to test `samtools bam2fq -s` with a BAM file… :smile:

¹ which we can improve later; let's not alter this for 1.3.1.

jmarshall:
Thanks, merged as 6994f30af178de66c7ad3900186b714cbd589b06.

PR #557, opened 2016-04-08
bam_split now adds @PG header to bam files

jenniferliddle:


jkbonfield:
For tidiness it should probably ```free(arg_list)``` at the end, but it's not a major issue as it is exiting anyway.  (It'll simply silence memory checkers, making it easier to spot more serious problems.)

jmarshall:
Thanks.  Rebased, simplified (the test shenanigans), and applied as 324d66211ee564d7c360edfd8a812949aed1c658.

PR #560, opened 2016-04-12
Implemented tab completion

george-hall-wtsi:
I have added a script which determines all Samtools subcommands and the
options available for these subcommands. It then generates a file which,
when sourced in Bash, allows for tab completion of all subcommands, and,
when appropriate, the long options for the relevant subcommand.

Not sure if something should be added to README/INSTALL/? to give instructions on how to configure the completions.


george-hall:
I have changed the implementation - options are now dynamically generated.

misc/samtools_tab_completion
```diff
@@ -0,0 +1,81 @@
...
+	COMPREPLY=()
+	cur="${COMP_WORDS[COMP_CWORD]}"
+
+	if [[ $COMP_CWORD == 1 ]] ; then
+
+		# If on the first word, generate possible subcommands, and tab complete those
+
+		opts=$($SAMTOOLS_BIN 2>&1 | awk '{if ($1 != "" && $1 != "Program:" && \
+			$1 != "Usage:" && $1 != "Version:" && $1 != "Commands:" && \
+			$1 != "--") print $1}' | xargs -n 1 -I @ printf @" ")
```
jmarshall: * It's not unintentional that `samtools 2>&1 | grep '^   '` (three spaces) gives just the subcommands. My more rudimentary completion script uses `samtools 2>&1 | sed -n '/^    /s/ *\([^ ]*\).*/\1/p'` to get the list of subcommand names.
* What is `xargs` for here?  The `$(…)` will return the words simply space-separated, which is all you need, no?

misc/samtools_tab_completion
```diff
@@ -0,0 +1,81 @@
...
+
+_samtools_options() 
+{
+
+	SAMTOOLS_BIN="${COMP_WORDS[0]}"
+	local cur current_sub opts
+	COMPREPLY=()
+	cur="${COMP_WORDS[COMP_CWORD]}"
+
+	if [[ $COMP_CWORD == 1 ]] ; then
```
jmarshall: `-eq`

misc/samtools_tab_completion
```diff
@@ -0,0 +1,81 @@
...
+			COMPREPLY=($(compgen -W "${opts}" -- ${cur}))
+			return 0
+		fi
+		
+	fi
+
+	# If we have not returned by now, assume that the user is looking for files
+	compopt -o filenames # Don't append a space
+	COMPREPLY=($(compgen -f "${cur}"))
+	return 0
```
jmarshall: You probably don't need this bit if you use `complete -F _samtools_options -o default samtools`.

jmarshall:
Cool.  Will the same completion function also work with `bcftools`?

Within reason, we can of course tweak the usage displays if necessary to make it easier to grep out the interesting bits.

george-hall:
Updated to incorporate your first two comments. 

I haven't done anything about your last comment yet, because I found that 
`samtools /<tab><tab>` would display the contents of /. Not sure if you want this to be possible or not? My current code (unintentionally) doesn't seem to allow this, but admittedly is slightly more messy than your suggestion of just using `-o default`. 

The same function seems to work well with `bcftools`. It seems that no tweaks are necessary, other than adding the line `complete -F _samtools_options bcftools` at the end of the file.

dmckean:
I attempted to alter samtools_tab_completion to offer mac-compatible completions in #740 . Does this preserve functionality in a bash 4 shell?

PR #571, opened 2016-05-10
Correctly ignore bed header lines

dnbaker:
Not a big issue, but bedcov shouldn't be throwing errors on header lines.


jmarshall:
What about [`browser` and `track` headers](https://genome.ucsc.edu/goldenPath/help/customTrack.html)?

dnbh:
They're not .bed files, so I wouldn't expect them to need support.

I think it'd be better to require a command-line flag for doing so. If it's important enough, I can add that.

PR #453, opened 2015-08-18
Improved header merge

daviesrob:
Builds on #419, extending the use of hash tables to all record types and removing all use of regular expression searches from bam_sort.c.  This should avoid quadratic behaviour when merging headers.

PR #583, opened 2016-05-19
.gitignore a binary

pjvandehaar:


jmarshall:
Thanks, but _test_pretty_header_ disappeared in dce3c288c3e31249a0183c52d489cad42c74a78d, as part of a header merge rewrite (PR #453).

I think you will find you have no _test/merge/test_pretty_header.c_ and if you still have a _test/merge/test_pretty_header_ executable you should just delete it by hand.

pjvandehaar:
Thanks!  That makes sense.

PR #579, opened 2016-05-18
Fixed the `-a` option for zero coverage positions

mfoll:
Fixes #374 after my [comment](https://github.com/samtools/samtools/issues/374#issuecomment-219427442).
The test in `test/mpileup/xx#depth1.sam` was unable to detect the bug because the two reads are on the same contig `xx`. Putting the second read to contig `xn` reveals the bug:
```
$ cat test/mpileup/xx#depth3.sam 
@HD	VN:1.4	SO:coordinate
@SQ	SN:xp	LN:20	AS:?	SP:?	UR:?	M5:bbf4de6d8497a119dda6e074521643dc
@SQ	SN:xx	LN:20	AS:?	SP:?	UR:?	M5:bbf4de6d8497a119dda6e074521643dc
@SQ	SN:xn	LN:20	AS:?	SP:?	UR:?	M5:bbf4de6d8497a119dda6e074521643dc
a1	16	xx	6	1	3M	*	0	0	*	*
b1	16	xn	13	1	3M	*	0	0	*	*
```
The current version returns:
```
$ ./samtools depth -a test/mpileup/xx#depth3.sam 
xx	1	0
xx	2	0
xx	3	0
xx	4	0
xx	5	0
xx	6	1
xx	7	1
xx	8	1
xn	1	0
xn	2	0
xn	3	0
xn	4	0
xn	5	0
xn	6	0
xn	7	0
xn	8	0
xn	9	0
xn	10	0
xn	11	0
xn	12	0
xn	13	1
xn	14	1
xn	15	1
xn	16	0
xn	17	0
xn	18	0
xn	19	0
xn	20	0
```
where contig `xx` is cut after it's last read in position 8. After the change I made in the code we have the correct output:
```
$ ./samtools depth -a test/mpileup/xx#depth3.sam 
xx	1	0
xx	2	0
xx	3	0
xx	4	0
xx	5	0
xx	6	1
xx	7	1
xx	8	1
xx	9	0
xx	10	0
xx	11	0
xx	12	0
xx	13	0
xx	14	0
xx	15	0
xx	16	0
xx	17	0
xx	18	0
xx	19	0
xx	20	0
xn	1	0
xn	2	0
xn	3	0
xn	4	0
xn	5	0
xn	6	0
xn	7	0
xn	8	0
xn	9	0
xn	10	0
xn	11	0
xn	12	0
xn	13	1
xn	14	1
xn	15	1
xn	16	0
xn	17	0
xn	18	0
xn	19	0
xn	20	0
```

PR #586, opened 2016-05-25
(Samtools#374)  Improved fix of mpileup/depth -a/-aa

jkbonfield:
This includes and improves on https://github.com/samtools/samtools/pull/579 by @mfoll.  In incorporating and testing this PR I found additional bugs, plus the original PR only fixed depth.  Mpileup had the same issues.

There is a known failure in the tests for both depth and pileup, which is to do with using a single -a in conjunction with a BED file.  In certain circumstances it can over egg the pudding a bit and include zero coverage output, normally associated with -aa.  I think it gets a bit blurry as to what is right and wrong when using -r/-b in conjunction with -a/-aa though so I don't see this as a major concern.

Conceptually we could argue that -r/-b are just filters so if we produced a new bam file with those regions then -a/-aa on that should be identical to -a/-aa with the -r/-b on the original.   Hence I'm flagging it as a failure currently.   It's simply a rats nest to fix and probably extremely low priority.


jmarshall:
Thanks, merged without the merge commit as 9c5fd9813b3cd2bea176a59dbc67507969a1c7fd and previous.

Please try to avoid pointless merge commits :smile:

PR #589, opened 2016-06-08
[plot-bamstats] Insert Size Y axis in Log10 scale parameter

antonkratz:
Added command line parameter parsing and Gnuplot code to plot the Y axis of the Insert Size plot optionally in log10 scale.


jmarshall:
Nice, thanks.  We'll probably change the long option to `--log-y` or so, so that it can control log scaling of other graphs too where that becomes useful.

jmarshall:
Thanks; merged as b25695bc25268aebe3372696c79fb38a634c2b1a.

Issue #494, opened 2015-11-26
memory issue with samtools sort


silvia-s:
I have asked samtools sort to use several combinations of threads and max memory per thread, but in each case it uses much more memory than requested and goes in the swap... Why is that? 
Here's some of the trials I've done (shown as stdout of "top"):

32060 silvia     18   0 16.3g  16g  892 R 172.1 68.9  55:39.99 /usr/local/genetics/samtools-1.2/samtools sort -n -@ 2 -m 5G [ARGS]

12400 silvia     18   0 12.6g  11g  648 R 35.6 50.7  16:46.94 /usr/local/genetics/samtools-1.2/samtools sort -n -@ 4 -m 2G [ARGS]

12438 silvia     23   0 6996m 6.5g  636 S 399.8 27.5  23:36.72 /usr/local/genetics/samtools-1.2/samtools sort -n -@ 4 -m 1G [ARGS]

Am I doing something wrong or is this a bug?
Thanks for your help!

daviesrob:
I don't think you're doing anything wrong.  I've done some tests and it turns out that samtools sort isn't very good at keeping track of how much memory it's using under certain conditions.

The quick solution for now would be to allow for this when setting the memory limit.  Judging by your tests, it's using about 50% more than you asked for, so if you reduce the limit by the same factor it should end up using about the right amount.  The only disadvantage of doing this is the sort might be a bit slower than it would otherwise be, but at least it won't go into swap and take forever.

Analysis:

The problem is with how bam_sort_core_ext counts the memory it has used when re-using bam structures.  Basically, it adds up the total space used by the data it's read, but doesn't account for that used by previously allocated structures that haven't been recycled yet.  This can be easily shown by adding a couple of printfs and running a test:

```
k = 1743091 max_k = 2097152 mem = 805306456 max_mem = 805306368
k = 1530684 max_k = 2097152 mem = 805306592 max_mem = 805306368
k = 291277 max_k = 2097152 mem = 155341736 max_mem = 805306368
```

On the second chunk, 212407 structs are still allocated but were not reused, so aren't accounted for in mem.

jmarshall:
Good fixes here probably require htslib changes to coalesce mallocs.

PR #593, opened 2016-06-21
memory optimization for better sorting performance

nathanweeks:
This proposed change improves sorting performance by storing BAM records in a contiguous memory region that is allocated/deallocated with a single malloc() and freed with a single free(), eliminating overhead associated with many small memory allocations/deallocations. This resulted in a 20-25% reduction in wall time when sorting ~127 million BAM records entirely in memory on two different benchmark systems (both with 64GB memory).

As a prerequisite, this requires the addition of a flexible array member to the bam1_t struct described in https://github.com/samtools/htslib/pull/393

The underlying buffer is both allocated and freed within bam_sort_core_ext(), and the existing bam1_t "data" member is made to point to the flexible array member for backwards compatibility with subroutines invoked from bam_sort_core_ext(). For backwards compatibility with other code, (e.g., bam_read1() in htslib), the bam1_t "data" member is made to point to the flexible array member. So that a realloc() on the bam1_t data member doesn't occur in the htslib bam_read1(), the "m_data" member is set to INT_MAX.

This change may make memory usage a bit more predictable (potentially resolving https://github.com/samtools/samtools/issues/494). It shouldn't cause an increase in the actual memory used by the process over the existing approach, since the extra physical memory won't be used until the corresponding memory pages are touched. To make the memory limit more precise, a guesstimate of the maximum size of the buf[] array could be subtracted from the size of the allocated bam_array. This also largely obviates the need for a previous patch I submitted (https://github.com/samtools/samtools/pull/507), which avoided costly memory deallocation (e.g., hundreds of millions of calls to free() for the data set I had benchmarked with) by simply exiting without freeing the memory, letting the OS reclaim it at the end of execution.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1741,7 +1741,7 @@ int bam_sort_core_ext(int is_by_qname, const char *fn, const char *prefix,
     size_t max_k, k, max_mem;
     bam_hdr_t *header = NULL;
     samFile *fp;
-    bam1_t b = {0}, **buf;
+    bam1_t b, **buf;
     uint8_t *bam_mem, *bam_mem_offset;
 
     if (n_threads < 2) n_threads = 1;
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1741,7 +1741,7 @@ int bam_sort_core_ext(int is_by_qname, const char *fn, const char *prefix,
     size_t max_k, k, max_mem;
     bam_hdr_t *header = NULL;
     samFile *fp;
-    bam1_t b, **buf;
+    bam1_t b = {.m_data = 0}, **buf;
     uint8_t *bam_mem, *bam_mem_offset;
 
     if (n_threads < 2) n_threads = 1;
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1741,12 +1741,11 @@ int bam_sort_core_ext(int is_by_qname, const char *fn, const char *prefix,
     size_t max_k, k, max_mem;
     bam_hdr_t *header = NULL;
     samFile *fp;
-    bam1_t *b, **buf;
-    uint8_t *bam_array;
+    bam1_t b = {0}, **buf;
+    uint8_t *bam_mem, *bam_mem_offset;
 
     if (n_threads < 2) n_threads = 1;
     g_is_by_qname = is_by_qname;
-    max_k = 0x10000; // initialized to initial size of buf array
     max_mem = _max_mem * n_threads;
     buf = NULL;
     fp = sam_open_format(fn, "r", in_fmt);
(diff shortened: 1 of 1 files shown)


daviesrob:
I've considered doing something like this.  The main problem with this version is the way you try to work out if the next record won't fit in the buffer:
```
if (((char *)(b + 1) + b->l_data + BGZF_MAX_BLOCK_SIZE > (char *)(bam_array + max_mem - 1))) {
```

This appears to assume that the next record will not be `BGZF_MAX_BLOCK_SIZE` bytes bigger than the last one.  While that may work for Illumina, it may well fail for variable-length reads like PacBio and ONT.

My conclusion is that the only safe way to do this is to copy the data into the buffer after having obtained it with `sam_read1`, fixing up the internal `->data` pointer on the way.  That way you know it's actually going to fit.  If it doesn't, you leave the record where it was, and call `sort_blocks` with the locations of the copied records plus one that didn't move.

Doing the above should go a long way to fixing issue #494.  I don't know how fast it would be, though.  It depends how long the extra copies would take compared to all the extra mallocs and frees.  I haven't got around to trying it yet, but I'd be interested to see the results if you can make it work.

nathanweeks:
@daviesrob: thanks for pointing that out; I had misinterpreted the htslib code and thought that BGZF_MAX_BLOCK_SIZE was a hard limit for alignment records. I've updated the pull request to implement your suggestion.

I've also implemented the technique suggested by @jkbonfield (https://github.com/samtools/htslib/pull/393#issuecomment-227353107) to obviate the need for the adition of a flexible array member in bam1_t (although in the long run it should be more efficient to convert the "data" pointer into a flexible array member to reduce sizeof(bam1_t) by 8 bytes, and allow the memory for a BAM record to be managed with fewer malloc()'s and free()'s---however, this would require code changes in a number of places).

I benchmarked the changes on a single compute node of NERSC's Cori system (2 x 16-core Intel "Haswell" processors @ 2.3 GHz, 128 GB DDR4 2133 Mhz memory) using the SRR1611184.bam file (~13 GB, 127 million records) from the elPrep demo (http://www.exascience.com/public-files/elprep-demo/SRR1611184.bam) and averaging the times of 3 trials. "samtools sort -@ 32 SRR1611184.bam" was ~11% faster in the memory-optimized ("memopt") version of samtools vs. the original samtools 1.3 (~414 seconds vs. ~467 seconds).  The total memory usage (resident set size) was modestly reduced in the memopt version (~26.5 GB vs. ~28.9 GB).

If "-m 2G -@ 32" is specified so that the sort runs entirely in-memory, then the memopt samtools is ~19% faster than samtools 1.3 (~379 seconds vs. ~470 seconds), with significantly less total memory usage (~37.2 GB vs. ~45.1 GB).

Since this minor source-code mod fix seems to effectively kill two birds with one stone, I'd strongly recommend that some form of it it be considered for inclusion until perhaps a more general memory-allocation solution affecting more of is decided on. As BAM sorting is a common operation (and I'd assume one of the more time-intensive samtools operations), I think it make sense to prioritize optimization for this use case.

I'll mention that long reads could be handled without the extra memcpy() for all data if sam_read1() were split into two functions, one that reads just enough of the record to determine its length (an approximation for SAM could be based on the line length), while the second reads the remainder of the record. I had a go at implementing this for SAM and BAM (I'm not as familiar with CRAM, but I assume it's possible):

https://github.com/nathanweeks/htslib/blob/feature/bam-mem-opt-no-memcpy/sam.c
https://github.com/nathanweeks/samtools/blob/feature/bam-mem-opt-no-memcpy/bam_sort.c

However, performance was no better than this memopt version---perhaps memory bandwidth isn't a bottleneck at this point in the code?---so it's probably not worth the additions to the HTSlib API for this application.

nathanweeks:
The warning looks like a bug in gcc < 4.9:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119

In any case, you're correct: the initialization is unnecessary. I've removed it. 

Regarding uintptr_t vs. size_t: as it's currently coded, bam_mem_offset is converted to uintptr_t for both comparison, as well as conversion for arithmetic and bitwise operations before being converted back to a pointer. CERT recommends using uintptr_t for pointer-to-integer or integer-to-pointer conversions:
https://www.securecoding.cert.org/confluence/display/c/INT36-C.+Converting+a+pointer+to+integer+or+integer+to+pointer

This is probably since C apparently doesn't guarantee that size_t can hold a pointer:
http://stackoverflow.com/questions/1464174/size-t-vs-intptr-t

e.g., while I don't think anyone is going to run SAMTools on IBM i (formerly known as OS/400) anytime soon, apparently the pointers are 128-bit, while size_t is 32-bit: 

http://stackoverflow.com/a/1572189

Nevertheless, size_t should work in the vast majority of computing environments.

nathanweeks:
My bad---at least the bam1_t _m_data_ member needs to be set to 0 the first time sam_read1() is invoked so that memory for the _data_ member is allocated. I've done this using a designated initializer, which gcc 4.2.1 (AFAIK the oldest still likely to be seen in the wild) compiles without warning. This also has the side effect of initializing the other structure members to 0 anyway (same as `b = {0}`)

jmarshall:
Just sidestep this entire question by having that single record `b` as a `bam1_t *` allocated in the usual way via `bam_init1`/`bam_destroy1`.

daviesrob:
With bam_mem_offset, my suggestion was actually to declare it as
```
size_t bam_mem_offset;
```
and then set `buf[k]` using
```
buf[k] = (bam1_t *)(bam_mem + bam_mem_offset);
```
so `bam_mem_offset` is used as an index into the `bam_mem` array instead of a pointer.

This allows the rest of the arithmetic on `bam_mem_offset` to be simplified, and in particular removes the need to cast from pointer to integer and back again.

nathanweeks:
@jmarshall: While that ostensibly adds the overhead of a pointer deference every time a member is accessed, in practice it shouldn't matter, and since more conventional with respect to the rest of the samtools codebase, I agree that it should be done that way.

@daviesrob: Thanks for the suggestion; that does make the code look a bit clearer.

I supposed it's about getting to the point where these commits could be squashed...

daviesrob:
Rebased, squashed, bug-fixed and merged as d39ed264b1.

Thanks very much for this, sorry it took so long to get around to adding it.

PR #594, opened 2016-06-27
Clarify bedcov subcommand usage

### LINKED ISSUE #588: Description of "samtools bedcov" is misleading
See [here](https://www.biostars.org/p/195497/) for an example of what people assume it to be doing.

In short, bedcov is described as computing the "read depth per BED region". What it's actually returning is the "sum of per-base read depths per BED region". I haven't followed the discussion leading to the creation of the bedcov tool, so I can't say whether what it's doing is actually what was originally intended...but in any case it'd be nice if the description were tweaked.

dkj:
Resolves #588 

Clear and accurate @dpryan79 ?

### DIFF
--- a/samtools.1
+++ b/samtools.1
@@ -698,7 +698,9 @@ samtools bedcov
 .RI [ options ]
 .IR region.bed " " in1.sam | in1.bam | in1.cram "[...]"
 
-Reports read depth per genomic region, as specified in the supplied BED file.
+Reports the total read base count (i.e. the sum of per base read depths) for each
+genomic region specified in the supplied BED file. Counts for each alignment file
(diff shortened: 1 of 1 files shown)


dpryan79:
:+1: Yup, looks good to me!

jmarshall:
Thanks both.  Reformatted for _samtools.1_ legibility in 80 columns and applied as 26e1ea5b4a58e7bff0fac45c821e7fd2df45261d.

brentp:
It would be clearer if the description under when just `samtools` is run without argument was something like:
```
bedcov    total reads per BED region
```
instead of currently:
```
bedcov         read depth per BED region
```

PR #595, opened 2016-06-28
Threading pool

jkbonfield:
See samtools/htslib#397 for discussions.


bam_sort.c
```diff
@@ -1815,6 +1848,7 @@ int bam_sort_core_ext(int is_by_qname, const char *fn, const char *prefix,
         }
         if (bam_merge_core2(is_by_qname, fnout, modeout, NULL, n_files, fns,
                             MERGE_COMBINE_RG|MERGE_COMBINE_PG|MERGE_FIRST_CO,
+                            //MERGE_FIRST_CO | MERGE_NO_TRANS,
```
mp15: Probably want to remove this line as it GNDN

bam_sort.c
```diff
@@ -1166,13 +1167,35 @@ int bam_merge_core2(int by_qname, const char *out, const char *mode,
...
+    }
+
     translation_tbl = (trans_tbl_t*)calloc(n, sizeof(trans_tbl_t));
     if (!translation_tbl) goto mem_fail;
     RG = (char**)calloc(n, sizeof(char*));
     if (!RG) goto mem_fail;
 
     // prepare RG tag from file names
-    if (flag & MERGE_RG) {
+    if ((flag & MERGE_RG)) {
```
mp15: Was this change to prevent a warning? It doesn't look necessary given you didn't do the same at 1171?

.travis.yml
```diff
@@ -15,6 +15,18 @@ matrix:
     - os: osx
       compiler: clang
 
+# For linux systems
+addons:
```
mp15: I think this bit should be promoted to it's own patch and landed asap.

bam_sort.c
jkbonfield: No idea to be honest; it doesn't ring any bells, but it's been a while.   It could be removed and then see if warnings do indeed appear.

bam_sort.c
jkbonfield: If we remove that we also want to remove the whole MERGE_NO_TRANS code elsewhere.  I recall working on this in merge so that it was faster and didn't spend a lot of pointless time translating IDs between files when infact all the IDs are identical because we just took the same header and spat it out N times.  This is simply and optimisation to remove that necessity.

Alas what I can't remember is why I didn't enable it.  I'm no expert in merge, so maybe it was just fear of the unknown or maybe it was a work in progress and something was amiss. Hmm.  Can you see anything wrong with the logic of what it's attempting to do?

Is this simply an issue of convolution between samtools merge and samtools sort?

jkbonfield:
This is ancient and has cruft which I'm wary of pulling in immediately before a release.

I'll look through it and see what's necessary.  It may even be preferable to just create a new PR given the state this is in.

Technically it may not even be necessary in many cases now we have --input-fmt-option added to pretty much every tool, permitting e.g. ```--input-fmt-option nthreads=8```, but -@ is simpler.

jkbonfield:
Closing in favour of PR #635

PR #621, opened 2016-11-04
Fix usage description of bamshuf/collate

kdmurray91:
Hi,

The short usage description of `samtools collate` conflicted with the longer description of options and with the `getopt` flag string. The attached fixes this.

Cheers,
K


jmarshall:
D'oh, thanks.

(Next time, you might like to base your branch on **develop** rather than **master**, but no biggie.)

PR #631, opened 2017-01-30
Added delete flag to merge, to unlink input bams as they are read

mpinese:
We recently needed this functionality in our work, as we were hitting disk limits during samtools merge.

It's an unsafe / 'expert mode' flag, as the files are not recoverable if the merge fails partway through, but I thought maybe others might find it useful.


daviesrob:
This is both incredibly scary (as it could lead to data loss) and may not work very well anyway.  It will only usefully close files early if they contain data from completely different regions of the output.  In the general case, where the range of positions or names in each input is very similar to the range in the outputs, all of the input files will be needed until the merge is almost finished.

PR #644, opened 2017-02-24
Fixed a typo: changed "suffices" to "suffixes".

steveb-cirina:
This is a very minor change.


jmarshall:
Thanks; probably this was an example of [this interesting phenomenon](http://english.stackexchange.com/questions/277945/is-the-plural-of-prefix-really-prefixes-rather-than-prefices).

steveb-cirina:
@jmarshall  Thanks for the link, that is an interesting Stack Exchange site.

PR #651, opened 2017-03-09
Stop samtools configure from always trying to configure htslib

daviesrob:
Because having samtools configure accidentally wipe out htslib's one is annoying.

Add an option --enable-configure-htslib to allow configuration of htslib
to be turned on and off.  The default is to only configure htslib when
it is in a subdirectory of samtools (as happens with the distribution
tarballs), and only if --with-htslib=search is used.

### DIFF
--- a/m4/ax_with_htslib.m4
+++ b/m4/ax_with_htslib.m4
@@ -25,8 +25,10 @@
 #   With --with-htslib=system, checks for a working default installation.
 #
 #   If a source tree is found or specified, it is added to AC_CONFIG_SUBDIRS
-#   (which unfortunately may cause a "you should use literals" warning when
-#   autoconf is run).
+#   if either --enable-configure-htslib is set, or where htslib is included
+#   in a subdirectory (for packages that want to supply an embedded htslib).
+#   Unfortunately this may cause a "you should use literals" warning when
+#   autoconf is run.
 #
 #   The following output variables are set by this macro:
 #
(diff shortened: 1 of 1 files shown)


jkbonfield:
There's something odd about this.

If I have htslib as a subdirectory of samtools in completely fresh checkouts I get all sorts of errors. Eg:

```
@ deskpro107386[tmp/samtools]; make clean
config.mk:33: ../htslib/htslib.mk: No such file or directory
config.mk:34: ../htslib/htslib_static.mk: No such file or directory
make: *** No rule to make target `../htslib/htslib_static.mk'.  Stop.
```

Configure and make do work, but not all targets do.  Maybe I'm confused about our build system.  Is the sub-directory mode only supported for tarballs and not git clones?  Should I be testing a release-tarball instead? (If so, how do I create one to test this?)

jkbonfield:
Ah ignore that - it fails unless we cd into the htslib subdirectory first and do an ```autoreconf -i``` in there first.  The error message above is misleading as it's reporting ../htslib, but actually it'll find the htslib subdir.

jkbonfield:
Merged, but there are still strange oddities around using subdirs with occasional failures that I can't pin down after doing things like ```make distclean``` and trying to restart the process.  This isn't related to this PR though (hence merged).  It looks to be a long-standing issue, which most likely won't affect the standard use of our all-in-one tarballs either.

Issue #653, opened 2017-03-13
Changes required to build on illumos (née OpenSolaris)


jblachly:
[illumos](https://www.illumos.org/) is a high performance server oriented operating system descended from Solaris. Neither `htslib` (see https://github.com/samtools/htslib/issues/487 ) nor `samtools` currently build out of the box, but can with a few minor changes. In addition item 1 below should allow building on Oracle Solaris.

1. Add check for requirement to link with libsocket (needed for both illumos and Solaris)
2. Silence warnings about implicit definition of alloca() when not using a GNU libc (this is optional)

PR incoming; these changes mirror precisely those made in my `htslib` PR. This and the matching `htslib` PR are my first time really modifying GNU autoconf scripts, so feedback is welcomed.

PR #654, opened 2017-03-13
Changes required to build on illumos

jblachly:
PR In two commits (only libsocket part is strictly required to build).

Will fix issue #653

### DIFF
--- a/bam2bcf_indel.c
+++ b/bam2bcf_indel.c
@@ -28,6 +28,24 @@ DEALINGS IN THE SOFTWARE.  */
 #include <assert.h>
 #include <ctype.h>
 #include <string.h>
+
+#if HAVE_ALLOCA_H
+# include <alloca.h>
+#elif defined __GNUC__
+# define alloca __builtin_alloca
+#elif defined _AIX
+# define alloca __alloca
+#elif defined _MSC_VER
+# include <malloc.h>
+# define alloca _alloca
+#else
+# include <stddef.h>
+# ifdef  __cplusplus
+extern "C"
+# endif
+void *alloca (size_t);
+#endif
+
 #include "htslib/hts.h"
 #include "htslib/sam.h"
 #include "bam2bcf.h"
--- a/configure.ac
+++ b/configure.ac
@@ -111,5 +111,8 @@ dnl connect() etc. fns are in libc on linux, but libsocket on illumos/Solaris
 libsocket=unneeded
 AC_SEARCH_LIBS(connect, socket, [libsocket=needed], [])
 
+dnl alloca.h should be included on platforms that supply it
+AC_FUNC_ALLOCA
+
 AC_CONFIG_FILES([config.mk])
 AC_OUTPUT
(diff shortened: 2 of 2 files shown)


jkbonfield:
libsocket support is most welcome, thanks.

The alloca thing I'd personally just kill the code that uses it as it's very isolated in usage.  Looking at bam2bcf_indel.c I see several ways of allocating memory in the same function:

```
        for (i = 1, n_types = 1; i < m; ++i)
            if (aux[i] != aux[i-1]) ++n_types;
...
    score1 = calloc(N * n_types, sizeof(int));
    score2 = calloc(N * n_types, sizeof(int));
...
        sc   = alloca(n_types * sizeof(int));
        sumq = alloca(n_types * sizeof(int));
...
    free(score1); free(score2);
```

So sometimes we calloc and free while other times we allocate off the stack.  Probably we should be consistent, and for the sake of portability just replace those alloca calls with malloc and possibly realloc if in a loop.

jkbonfield:
I just noticed this nugget in bam2bcf_indel.c - the solution for alloca is even more obvious now!

```
        if (n_types >= 64) {
            free(aux);
            // TODO revisit how/whether to control printing this warning
            if (hts_verbose >= 2)
                fprintf(stderr, "[%s] excessive INDEL alleles at position %d. Skip the position.\n", __func__, pos + 1);
            return -1;
        }
```

Given there is a maximum of 64, we should probably ```#define MAX_N_TYPES 64``` and then use constant sized arrays throughout, also permitting us to throw away all those callocs (albeit replace with memsets or local assignments).

jkbonfield:
Alloca changes are obsoleted by the C99 branch.

The libsocket bit is needed, and I'll look into merging this once the C99 code is merged.  I also needed a few other tweaks when run on OpenIndiana.

- It needed configuring with --without-ncursesw as by default the ax_with_ncurses.m4 finds an ncursesw library and then due to no headers for it it fails to go on to check for the ncurses library (which does work).
- It needs the same -lpthread instead of -pthread option during linking that we did to htslib (now also part of the samtools C99 branch).
- misc/ace2sam needs linking against htslib as simple inclusion of kstring.h causes symbols to be required from htslib (ksplit_core) even if we don't call the functions that call that function.  This is a linker difference, but I'm surprised we haven't hit this before.
- I also needed some tweaks to the mpileup tests as the default shell (korn shell) is utterly broken and doing bizarre things during read calls (maybe a LOCALE issue).  Switching it to bash and switching the awk used to nawk makes the test harness work.

jkbonfield:
Please try https://github.com/jkbonfield/samtools/tree/C99+illumos and let us know if it works for you too.  It isn't a PR here yet because it depends on another PR not yet merged (C99) and it also needs some tidying up of authors (it's me, not @daviesrob, but I used a shared account on his VM without realising the implications to git!).

jmarshall:
Thanks for the PR, which has now been replaced by PR #1165. The `alloca` stuff is no longer needed as alloca has been exterminated (yay!), and the configure AC_SEARCH_LIBS for `-lsocket` stuff is no longer needed as since #649 we get that via `HTSLIB_static_LIBS` if necessary.

Per @jkbonfield's comments: the --without-ncursesw thing is still useful, which would appear to be a bug in _ax_with_curses.m4_; the `-lpthread` stuff has long since landed from other PRs; the ksh weirdness has been worked around differently in #1151; and updated versions of the ace2sam and awk fixes are in the new PR.

PR #656, opened 2017-03-31
Improvements for C99/XOPEN compliance.

jkbonfield:
Tested on gcc6.

Replaced alloca with malloc.
Avoid arithmetic on void*.
Ensure unsigned types for bit fields.
Use of strings.h for strcasecmp.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@
 CC       = gcc
 AR       = ar
 CPPFLAGS =
+#CFLAGS   = -g -Wall -O2 -pedantic -std=c99 -D_XOPEN_SOURCE=600
 CFLAGS   = -g -Wall -O2
 LDFLAGS  =
 LIBS     =
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ libbam.a:$(LOBJS)
 	$(AR) -csru $@ $(LOBJS)
 
 samtools: $(AOBJS) libbam.a libst.a $(HTSLIB)
-	$(CC) -pthread $(ALL_LDFLAGS) -o $@ $(AOBJS) libbam.a libst.a $(HTSLIB_LIB) $(CURSES_LIB) -lm $(ALL_LIBS)
+	$(CC) $(ALL_LDFLAGS) -o $@ $(AOBJS) libbam.a libst.a $(HTSLIB_LIB) $(CURSES_LIB) -lm $(ALL_LIBS) -lpthread
 
 # For building samtools and its test suite only: NOT to be installed.
 libst.a: $(LIBST_OBJS)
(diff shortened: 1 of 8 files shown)


bam2bcf_indel.c
```diff
@@ -439,10 +439,12 @@ int bcf_call_gap_prep(int n, int *n_plp, bam_pileup1_t **plp, int pos, bcf_calla
...
     { // compute indelQ
-        int *sc, tmp, *sumq;
-        sc   = alloca(n_types * sizeof(int));
-        sumq = alloca(n_types * sizeof(int));
-        memset(sumq, 0, sizeof(int) * n_types);
+        int sc_a[16], sumq_a[16];
+        int tmp, *sc = sc_a, *sumq = sumq_a;
+        if (n_types > 1) {
+            sc   = malloc(n_types * sizeof(int));
+            sumq = calloc(n_types,  sizeof(int));
```
mcshane: See samtools/bcftools#589

jkbonfield: Samtools doesn't build with -Wc++-compat, so maybe it's not worth changing unless purely for consistencies sake. (This is time limited anyway?)

mcshane: Yeah. Was just mentioning it here for consistency.

Makefile
```diff
@@ -24,6 +24,7 @@
 CC       = gcc
 AR       = ar
 CPPFLAGS =
+#CFLAGS   = -g -Wall -O2 -pedantic -std=c99 -D_XOPEN_SOURCE=600
```
jmarshall: Why checking in mysterious commented-out stuff?

bam2bcf_indel.c
```diff
@@ -488,7 +490,7 @@ int bcf_call_gap_prep(int n, int *n_plp, bam_pileup1_t **plp, int pos, bcf_calla
                 if (seqQ > 255) seqQ = 255;
                 p->aux = (sc[0]&0x3f)<<16 | seqQ<<8 | indelQ; // use 22 bits in total
                 sumq[sc[0]&0x3f] += indelQ < seqQ? indelQ : seqQ;
-//              fprintf(stderr, "pos=%d read=%d:%d name=%s call=%d indelQ=%d seqQ=%d\n", pos, s, i, bam1_qname(p->b), types[sc[0]&0x3f], indelQ, seqQ);
+                //              fprintf(stderr, "pos=%d read=%d:%d name=%s call=%d indelQ=%d seqQ=%d\n", pos, s, i, bam1_qname(p->b), types[sc[0]&0x3f], indelQ, seqQ);
```
jmarshall: Now hyper-indented…

bam2bcf_indel.c
jmarshall: `if (n_types > 16)` perhaps?

Makefile
jkbonfield: It's more of a self document as to how to enable strict C99 mode.  I think it adds value as a comment personally.

bam2bcf_indel.c
jkbonfield: Both noted and agreed.  Thanks

jkbonfield:
Updated from review comments and force pushed.  Thanks both.

jkbonfield:
Note this doesn't solve everything needed to build on Illumina / OpenIndiana (Solaris), but I'm not specifically trying to get an OS specific set of patches here.

It is however a prerequisite, along with a -lpthread vs -pthread change analogous to the htslib one (which I may append on to this too) and various tweaks to the test harness (awk & shell versions mainly).

jmarshall:
As noted in my subsequently-ignored comment on samtools/htslib#255, wouldn't using `AX_PTHREAD` lead to better results than just changing to raw and not necessarily correct `-lpthread` everywhere?

jkbonfield:
I didn't ignore your comment.  I simply tested what worked on all the platforms we have access to and -lpthread works everywhere so far while -pthread does not.

Yes we could use AX_PTHREAD, but my solution worked when not using autoconf, unless you're now advocating for making autoconf the default and mandatory way of building (my preference for years!).

jmarshall:
What the comment said is that raw `-lpthread` does not work well on macOS.

jkbonfield:
Specifically that it didn't work well on MacOS X 12 years ago.  As far as I can tell it works fine now (I tested it).

However we could still go for AX_PTHREADS if we can find real evidence that there is a need for it.  I experimented with it and found it broken currently, but perhaps I was just misusing it.  Either way it's much more work to get implemented so I went with the easy fix until I find good reason for the complex fix.

jkbonfield:
I'll correct that last statement.  AX_PTHREAD depends on config.sub and config.guess, which are only installed by autoreconf -i if it detects you have AM_INIT_AUTOMAKE in your configure.ac.  You can run automake --add-missing manually (eg in a bootstrap script) to get them, but the recommended solution appears to be just to bundle copies of them with your source tree if you don't wish to be using automake.

Messy and I'm going off this whole idea.

That said, it did actually decide to not use -lpthread for the Mac so it clearly spotted the lack of need.

jmarshall:
It is true that there no longer appear to be different flavours of _libSystem.dylib_ (such as _libSystem_debug.dylib_), so that particular problem may have gone away.

Ah, yes — `AX_PTHREAD` wanting _config.sub_/_config.guess_ is why I was formerly toying with just writing out `case $(uname -s) in Darwin) PTHREAD= ;; *) PTHREAD=-pthread ;; esac` or so explicitly in _configure.ac_.  Whatevs then…

PR #659, opened 2017-04-06
Added option to produce fastq files from BC or other tags

jenniferliddle:
Required for the 10x 
This also handles tags containing separators.


mcshane:
Introduces some clang compiler warnings. See https://travis-ci.org/samtools/samtools/jobs/219228463

Also, need to add documentation to `samtools.1`

jenniferliddle:
Good points.
Compiler warnings fixed, and man page updated.

sam_view.c
```diff
@@ -746,47 +840,16 @@ static bool bam1_to_fq(const bam1_t *b, kstring_t *linebuf, const bam2fq_state_t
...
-                    for (i = 0; i < qlen; ++i) {
-                        kputc(33 + qual[i], linebuf);
-                    }
-                }
-            }
+        if (qual && *qual) {
+            kputs(qual, linebuf);
         } else {
-            for (i = 0; i < qlen; ++i) {
+            for (i = 0; i < strlen(seq); ++i) {
```
jkbonfield: Although modern compilers now optimise this through C extensions like gcc ```__attribute_pure__``` I am unsure of how widespread this and equivalent optimisations are.   Most likely it's standard now, but it still makes me nervous.  Without it, this loop turns from O(N) into O(N^2) complexity.  I would suggest doing strlen once and referring to that within the loop.

Edit: even with gcc 6 and clang 3.8 this optimisation is only made with -O so in a traditional debugging environment the code would become O(N^2).

sam_view.c
```diff
@@ -708,36 +799,39 @@ static readpart which_readpart(const bam1_t *b)
...
     if (state->has12) {
+        readpart readpart = which_readpart(rec);
         if (readpart == READ_1) kputs("/1", linebuf);
         else if (readpart == READ_2) kputs("/2", linebuf);
     }
     if (state->copy_tags) {
         for (i = 0; copied_tags[i]; ++i) {
             uint8_t *s;
-            if ((s = bam_aux_get(b, copied_tags[i])) != 0) {
+            if ((s = bam_aux_get(rec, copied_tags[i])) != 0) {
```
jkbonfield: There is an implicit assumption here that all copied_tags[] array entries are for :Z: tags.  While this is true in the standard, it may not be true in actual data files.  This could lead to calling kputs on a non-string and running off the end of the bam record causing a crash or leaking the contents of the remainder of a malloced block of data.

Please add a check for ```*s == 'Z'```.

sam_view.c
```diff
@@ -795,50 +858,215 @@ static bool bam1_to_fq(const bam1_t *b, kstring_t *linebuf, const bam2fq_state_t
     return true;
 }
 
+/*
+ * Create FASTQ lines from the barcode tag using the index-format 
+ */
+static bool tags2fq(bam1_t *rec, const bam2fq_state_t *state, const bam2fq_opts_t* opts)
+{
+    uint8_t *p;
+    char *ifmt_orig = strdup(opts->index_format);
```
jkbonfield: Is this constant strdup necessary?  You don't appear to ever be modifying the data pointed to by ifmt.  It also leads to memory leaks (see valgrind --leak-check=full).  (Possibly at the ```if (!tag) return;``` statement.)

sam_view.c
```diff
@@ -795,50 +858,215 @@ static bool bam1_to_fq(const bam1_t *b, kstring_t *linebuf, const bam2fq_state_t
...
+
+        if (action=='i' && *sub_tag && state->fpi[file_number]) {
+            make_fq_line(rec, sub_tag, sub_qual, &linebuf, state);
+            fputs(linebuf.s, state->fpi[file_number++]);
+        }
+        free(sub_qual); free(sub_tag);
+
+    }
+
+    free(ifmt_orig);
```
jkbonfield: Memory leak of ```linebuf.s``` ?

sam_view.c
```diff
@@ -679,24 +702,92 @@ typedef struct bam2fq_opts {
...
+        j++;
+    }
+    return str;
+}
+
+/* return the read, reverse complemented if necessary */
+static char *get_read(const bam1_t *rec)
+{
+    int len = rec->core.l_qseq + 1;
+    char *read = calloc(1, kroundup32(len));
```
jenniferliddle: I was following Matt's suggestion at https://matt.sh/howto-c
It does, as you say, give you a nul termination. Also, I *like* calloc()

Looking at the code carefully, I agree that kroundup32() is not strictly necessary here, but I have in the past found problems with a malloc'd string when using strcpy() when it was not rounded up and optimisation was on. It took me some time to track that one down, so I'm understandable cautious now.

Checking the return value: well spotted.

sam_view.c
```diff
@@ -708,36 +799,39 @@ static readpart which_readpart(const bam1_t *b)
...
-    uint8_t *qual = bam_get_qual(b);
-    const uint8_t *oq = NULL;
-    if (state->use_oq) {
-        oq = bam_aux_get(b, "OQ");
-        if (oq) oq++; // skip tag type
+    int n = 0;
+    while (**s) {
+        if (**s == '*') { n=-1; (*s)++; break; }
+        if ( !isdigit(**s)) break;
+        if (isdigit(**s)) { n = n*10 + ((**s)-'0'); }
```
jenniferliddle: I agree it's unnecessary. Changed.

sam_view.c
jenniferliddle: Well spotted.

sam_view.c
jenniferliddle: Fixed, if only to fix your nerves :-)

sam_view.c
jenniferliddle: Hmmm. Historical reasons. I think that getLength(&ifmt) *did* change the data at one time.

sam_view.c
jenniferliddle: Oh yes...

sam_view.c
daviesrob: I disagree with Matt.  calloc() can only ever be as fast as malloc(), and whenever it reuses an existing region from the pool it's a lot slower.

kroundup32() can be wasteful when the size is a multiple of the page size.  The bytes needed for book-keeping the heap leak into the next page, resulting in nearly 4k unused for each allocation.  It would be much better to malloc(rec->core.l_qseq + 1) and set read[rec->core.l_qseq] = '\0';

jenniferliddle: removed kroundup32() as unnecessary.

jmarshall:
I don't understand why you're allocating memory by hand at all.

1. We have an abstraction for strings: `kstring_t`.  Is there a reason for not using it here?  (In particular, by having a `kstring_t *tmp` as a parameter passed in, so that the buffer is reused rather than re-mallocing a million times.)

2. There is no inherent need for memory allocation or `reverse()` here at all.  Is there a reason for not doing this the way the existing code does, directly from the `bam1_t` into `linebuf`?  Abstract it into local static `kput_seq` and `kput_qual` functions if you will.

sam_view.c
```diff
@@ -34,13 +34,21 @@ DEALINGS IN THE SOFTWARE.  */
...
 #include "htslib/khash.h"
 #include "htslib/thread_pool.h"
 #include "samtools.h"
 #include "sam_opts.h"
+
+#define DEFAULT_BARCODE_TAG "BC"
+#define DEFAULT_QUALITY_TAG "QT"
+
+#define xstr(s) str(s)
+#define str(s) #s
```
jmarshall: These things are string literals already, so there's no need for the xstr trick.

jmarshall:
I have just been reading the proposed code, not running or benchmarking it; so my comments are about code quality and comprehensibility.  IMHO comprehensibility has been reduced; whether that is justified or not is another question.

I don't think [ranting like this](https://github.com/samtools/samtools/blob/6d612fea9cb324b15525e172a63270c32420494a/sam_view.c#L728-L729) is particularly appropriate for the mainline branch, and I'm astonished that commentary and code such as

    /* reverse a string in place
       (stolen from http://stackoverflow.com/questions/8534274/is-the-strrev-function-not-available-in-linux)
    */

was proposed for inclusion and flabbergasted that it was merged.  Are we complying with the licensing conditions of this stolen code?  (Hint: [no](https://stackexchange.com/legal#attribution).)

jkbonfield:
Please submit a PR to fix this.

PR #663, opened 2017-04-11
C99+illumos

jkbonfield:
Mostly testing tweaks to cope with Korn-shell and awk vs nawk.


mp15:
This looks pretty non-invasive, did we ever consider merging this?

jmarshall:
Replaced by PR #1165, which includes the ace2sam fix and the awk changes. AC_SEARCH_LIBS for `-lsocket` is no longer needed as since #649 we get that from `HTSLIB_static_LIBS` if necessary. The ksh weirdness was worked around differently in #1151.

PR #668, opened 2017-04-20
Output CRAM as temporary sort format iff there are more than 64k cigar ops

### LINKED ISSUE #667: Samtools sort CIGAR operations limitation, despite not using BAM
samtools 1.4 sort gives me an error when trying to sort a SAM or CRAM file with more than 64K CIGAR operations.

Example command:

/data/projects/phillippy/software/samtools-1.4/samtools sort combined_with_header.sam -o combined_with_header_and_sorted.sam

Example output:

[E::bam_write1] too many CIGAR operations (66411 >= 64K for QNAME "CHM1.gi|929855600|gb|LJII01000029.1|")
samtools sort: failed to create temporary file "combined_with_header_and_sorted.sam.tmp.0000.bam": Value too large for defined data type

I have exactly the same problem when trying to sort a CRAM file.

jkbonfield:
Fixes #667.

In theory we could use CRAM for all temporary output files, but
currently it is too high in memory usage.  See the comments included
in this commit for a discussion on how this could be avoided, but for
now we are making the smallest change that fixes the issue as there
are larger issues yet to resolve in sort memory usage.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1219,6 +1219,16 @@ int bam_merge_core2(int by_qname, const char *out, const char *mode,
         if ((translation_tbl+i)->lost_coord_sort && !by_qname) {
             fprintf(stderr, "[bam_merge_core] Order of targets in file %s caused coordinate sort to be lost\n", fn[i]);
         }
+
+        // Potential future improvement is to share headers between CRAM files for
+        // samtools sort (where all headers are identical.
+        // Eg:
+        //
+        // if (i > 1) {
+        //     sam_hdr_free(cram_fd_get_header(fp[i]->fp.cram));
+        //     cram_fd_set_header(fp[i]->fp.cram, cram_fd_get_header(fp[0]->fp.cram));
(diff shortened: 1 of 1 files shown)


bam_sort.c
```diff
@@ -1648,18 +1658,30 @@ static void *worker(void *data)
...
-//        {"no_ref",      CRAM_OPT_NO_REF,  {1},     NULL}
-//    };
-//    opt[0].next = &opt[1];
-//    if (write_buffer(name, "wc1", w->buf_len, w->buf, w->h, 0, opt) < 0)
-//        w->error = errno;
+
+    uint32_t max_ncigar = 0;
+    int i;
+    for (i = 0; i < w->buf_len; i++) {
+        int nc = w->buf[i]->core.n_cigar;
```
jmarshall: uint32_t

jkbonfield: Well read.  Is there a point to this comment?

jkbonfield: Ah no matter, you mean the last part of the 4 lines quoted, not the first part (uint32_t max_ncigar).   A little less terseness would help!

Issue #666, opened 2017-04-15
option to reverse select specified regions


jyh1:
Sometimes I need to filter out alignments which are mapped to a specific chromosome from a bam file. It seems that samtools has no built-in support for this kind of thing and I found some workarounds  [there](https://www.biostars.org/p/128967/). 
Could there be an option to reverse select the specified regions or something like **-U** but works at the index level?
#551

PR #669, opened 2017-04-23
option -v to filter out reads in given regions.

jyh1:
I wanted to be able to remove reads appearing in the specified indexes. For
    instance, we could exclude reads mapped to chrM with command `samtools view
    -v a.bam chrM`, which is equivalent with `samtools view a.bam chr1 chr2 ...`.
    For consitency, `samtools view -v a.bam` will print nothing.

    A very ad-hoc solution. The right way is to support this in the sam.c of htslib?
#666

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -237,6 +237,7 @@ int main_samview(int argc, char *argv[])
 {
     int c, is_header = 0, is_header_only = 0, ret = 0, compress_level = -1, is_count = 0;
     int is_long_help = 0;
+    int is_reverse_selection = 0;
     int64_t count = 0;
     samFile *in = 0, *out = 0, *un_out=0;
     FILE *fp_out = NULL;
@@ -245,6 +246,8 @@ int main_samview(int argc, char *argv[])
     char *fn_in = 0, *fn_out = 0, *fn_list = 0, *q, *fn_un_out = 0;
     sam_global_args ga = SAM_GLOBAL_ARGS_INIT;
     htsThreadPool p = {NULL, 0};
+    int index_count = 0;
+    char **index_selection = NULL;
 
     samview_settings_t settings = {
         .rghash = NULL,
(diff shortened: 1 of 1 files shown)


mblue9:
This would be SO useful imo! I would love to have this option right now for filtering out mitochondrial reads.

colindaven:
Isn't this similar in scope to https://github.com/samtools/samtools/pull/423 ?

daviesrob:
@colindaven It's not quite the same.  #423 inverts the filter test but leaves the region list unchanged - so flags and BED files (which act as a filter) get inverted.  This attempts to invert the region list but leaves the filter unchanged.  The interaction between the two may get a bit tricky, so it needs some thought.

jyh1:
@daviesrob Exactly. In essence, the region list is just a kind of filter and I think it only exists for efficiency reason so we could take advantage that the reads are indexed. I believe that's where all the confusion comes from. I think another option could be allowing the user to inefficiently select the regions at the filter level, as oftentimes it is what most of us end up doing by adding an awk following the pipe.

jkbonfield:
How is the inversion happening with regards to overlaps?  If we have a region Z:1000-2000 on a seq Z of length 3000, is this equivalent to reads that overlap Z:1-999 and Z:2001-3000 or is it equivalent to reads that do not overlap Z:1000-2000?  I can see both potentially being useful.

jyh1:
@jkbonfield Right, both will be useful in some circumstances. In this pull request it is equivalent to reads that do not overlap Z:1000-2000. I couldn't figure out a nice way to support both of them at the same time without over complicating the options. IMO it would be very useful to simply support filtering out an entire seq as it is the most common case (like removing chrM reads) and currently it is a pain in the ass to do this simple job,  you have to either convert to sam, pipe to an awk, convert back or do some bash scripting exercise like [here](https://www.biostars.org/p/128967/). The users could always resort to `bedtools intersect` if they really need precisely mask out certain regions.

Issue #671, opened 2017-04-25
New -T, --copy-tags option for samtools fastq


pd3:
@jenniferliddle, in addition to the existing option

     -t  ... copy RG, BC and QT tags to the FASTQ header line

and the new extension `--barcode-tag` added by https://github.com/samtools/samtools/commit/6d612fea9cb324b15525e172a63270c32420494a, it would be useful to be able to copy arbitrary tags. I'd like to propose a new option for that, say

    -T, --copy-tags .. list of comma-separated tags to copy
 
My use case is copying the BX tag in 10x Genomics data.

jenniferliddle:
My first step was to add a unit test for the existing -t option. I discovered that if the -s option is given, then -t *only* copies the tags to the -s file, not to the -1 or -2 files.

Is this the expected behaviour?

Do we want the new -T option to follow this behaviour?

pd3:
Not sure what is the expected behavior for the `-t` option, but I'd expect it to do it for all files and the same for `-T`.  Thank you for looking into this.

jkbonfield:
The copy tags functionality arrived with 84d070296492e86e450c4f2abda8e9521ccd6da2, but at this point there wasn't any -1 and -2 option.  @mcshane can you please clarify that the intention was to write to all files?

I suspect this omission arrived with the -1 and -2 code subsequently added inf0cb72c7 or subsequent  20b5b13d or b73d36e9 updates (ping @mp15).  I haven't checked though.  We should check out those various commits to see when this difference arrived so we can see if it was accidental or by design.  (I strongly suspect it's an accidental difference.)

mcshane:
The tags should be copied to all output files.

jenniferliddle:
Thanks people. I'm always hesitant to make any changes that affect current functionality, but since we seem to have a consensus on this, I shall make the change.

PR #674, opened 2017-05-04
Added -T option for fastq 

jenniferliddle:
Implements  #671


sam_view.c
```diff
@@ -817,6 +825,37 @@ static int getLength(char **s)
...
+{
+    uint8_t *s = bam_aux_get(rec, tag);
+    if (s) {
+        char aux_type = *s;
+        switch (aux_type) {
+            case 'C':
+            case 'I':
+            case 'S': aux_type = 'I'; break;
+            case 'c':
+            case 's': aux_type = 'i'; break;
```
jkbonfield: Initially I thought there was a bug as that you have C/I/S above and only c/s below it, but I see the I->I is just a nop.  I'll nuke the ```case 'I'``` line on merge unless you have an objection.

sam_view.c
```diff
@@ -817,6 +825,37 @@ static int getLength(char **s)
...
+        kputc(aux_type=='I'? 'i': aux_type, linebuf);
+        kputc(':', linebuf);
+        switch (aux_type) {
+            case 'H':
+            case 'Z': kputs(bam_aux2Z(s), linebuf); break;
+            case 'i': kputw(bam_aux2i(s), linebuf); break;
+            case 'I': kputuw(bam_aux2i(s), linebuf); break;
+            case 'A': kputc(bam_aux2A(s), linebuf); break;
+            case 'f': kputd(bam_aux2f(s), linebuf); break;
+            default:  kputs("*** Unknown Type ***", linebuf); break;
```
jkbonfield: This isn't an appropriate way of returning an error.  It ought to be non-void and return -1 or similar.

It also doesn't handle the ```B``` aux type (which in itself is a monstrosity to handle).  If you wish to not handle this particular type then it should be in a comment to state this and also documented.  Note this isn't just a case of not being able to output it if asked for, but any record containing this type produces an error irrespective of whether we wanted to emit this tag.  Even so, it's a big improvement on the previous version which simply core dumped!

Given the complexity of the B tag type I'm wondering if what we really need is to break htslibs ```sam_format1()``` function in two, adding a ```sam_serialise_aux()``` function there, and to then call that function from here to avoid code duplication.

This could make the code more efficient too.  Instead of having a list of tags to copy and for each tag type scanning through the bam aux data to find that tag, we'd stream through the tag list (as done in ```sam_format1```) and either serialise the tag or skip it depending on whether the tag is present in the list of tags to copy (a quick hash lookup).  **I'd be happy with revamping this though to be a subsequent PR**(if ever required0 so for now if you want to get on just make 'B' tag an appropriate warning message about not being supported and raise the issue of B being unsupported as a wishlist item (it's unlikely anyone would actually care for it!).

sam_view.c
```diff
@@ -1116,6 +1158,22 @@ static bool init_state(const bam2fq_opts_t* opts, bam2fq_state_t** state_out)
     state->filetype = opts->filetype;
     state->def_qual = opts->def_qual;
 
+    state->taglist = kl_init(ktaglist);
```
jkbonfield: This memory is leaked, although it's a minor issue as the program is exiting.

Please consider adding a ```kl_destroy``` call at the end to tidy up.

sam_view.c
jenniferliddle: No objection.

sam_view.c
jenniferliddle: > This isn't an appropriate way of returning an error. It ought to be non-void and return -1 or similar.

Ah. I wasn't trying to return an error, and I didn't want to treat this as a fatal error, but just to mark it somehow as 'unhandled' and carry on. Not handling type 'B' was a deliberate decision, but in that case I agree that 'unknown type' is a bit misleading. Perhaps 'unhandled type' would be better.
An alternate approach would be to silently ignore any tag with an unhandled type, or put a warning out to stderr.

sam_view.c
jenniferliddle: Will do.

sam_view.c
jkbonfield: Ah yes I see what it's trying to do, however at the moment if you run:

```
./samtools fastq -t  ../htslib/test/auxf#values.sam
```

you get this output:

```
@Fred	RG:Z:ID
CTGAGCTAGC
+
**********
@Jim	BC:B:*** Unknown Type ***
TTTTTTTTTT
+
""""""""""
```

I see now it's pure fluke my exhaustive htslib tag testing file happens to be using BC (amongst many others) which is an official tag and so it's being output by default.

I'd still prefer a stderr message or similar (emitted once rather than spammed), or even a hard failure.   If we've explicitly asked for tag XX and we can't emit it then that is probably a failure condition.  Inlining the complain to the fastq output is just going to byte us later on.

jkbonfield:
Squashed and merged as d7c2329486ca5c670dc2cb6cc8ec5714501ba050

PR #675, opened 2017-05-04
ability to sort/merge by tag 

pmarks:
Add a new -t TAG argument to sort and merge, allowing the file to be sorted by tag. For highly multiplexed BAM files (e.g. single-cell RNA-seq, or linked-reads), this is useful to group all the data for one 'barcode' together in the BAM file.

This emerged as a useful feature in discussion w/ @gavinha and @imielinski among others.

Any suggestions on how to improve the patch or add unit tests would be helpful.


walaj:
I second this PR. Very useful functionality, and it works in my tests (tried separately sorting by a ``Z`` tag and ``i`` tag).

walaj:
One small hangup is that the SAM format defines the following sort orders (``SO``): unknown (default), unsorted, queryname and coordinate.

The tag-sorted BAMs of course don't follow any of these. The format would have to be updated to specify tag-sorted info in the BAM header.

In the meantime, this PR should be amended to also change the sort order to ``unknown`` in the header. Right now, it stays as ``coordinate`` even after the tag sorting, which will confuse some tools once it finds alignments out of genomic order.

imielinski:
this is great!


Will give it a spin ..


Marcin

________________________________

pmarks:
@jkbonfield @jmarshall @jenniferliddle  - any comments on this PR?  Is there any additional process I need to follow to get this considered?

bam_sort.c
```diff
@@ -1611,13 +1633,71 @@ static int change_SO(bam_hdr_t *h, const char *so)
...
+        return (((uint64_t)a->core.tid<<32|(a->core.pos+1)<<1|bam_is_rev(a)) < ((uint64_t)b->core.tid<<32|(b->core.pos+1)<<1|bam_is_rev(b)));
+    }
+}
+
+// Sort record by tag, using pos or read name as a secondary key if tags are identical. Reads not carrying the tag sort first. 
+// Tags are first sorted by the type character (in case the types differ), or by the appropriate comparator for that type if they agree.
+static inline int bam1_lt_by_tag(const bam1_p a, const bam1_p b) 
+{
+    uint8_t* aux_a = bam_aux_get(a, g_sort_tag);
+    uint8_t* aux_b = bam_aux_get(b, g_sort_tag);
```
daviesrob: bam_aux_get() does a linear scan through the read aux data, so putting two of them here is likely to slow the sort down a lot.  It would be better to replace the definition of bam1_p with something like this:
```
typedef struct bam1_p {
    const bam1_t *b;
    const uint8_t *tag;  // Or maybe a ptrdiff_t giving the offset of the tag in the aux data
} bam1_p;
```

The tag pointers can be populated when the bam records are read in, prior to the sort.  This will give a 2*log(n) reduction in the number of calls to bam_aux_get().

You would need to do the same thing during the merge stage, but note that merge can change the aux data in `bam_translate()`.  You would need to set the tag pointer **after** this has happened.

bam_sort.c
```diff
@@ -1611,13 +1633,71 @@ static int change_SO(bam_hdr_t *h, const char *so)
...
+
+    if (aux_a == 0 && aux_b > 0) {
```
daviesrob: Should be:
```
if (aux_a == NULL && aux_b != NULL) {
```

The comparison `aux_b > 0` is undefined behaviour, see [CERT ARR36-C](https://www.securecoding.cert.org/confluence/display/c/ARR36-C.+Do+not+subtract+or+compare+two+pointers+that+do+not+refer+to+the+same+array)

daviesrob: Should be:
```
if (aux_a == NULL && aux_b != NULL) {
```

The comparison `aux_b > 0` is undefined behaviour, see [CERT ARR36-C](https://www.securecoding.cert.org/confluence/display/c/ARR36-C.+Do+not+subtract+or+compare+two+pointers+that+do+not+refer+to+the+same+array)

bam_sort.c
```diff
@@ -1611,13 +1633,71 @@ static int change_SO(bam_hdr_t *h, const char *so)
...
+        return 1;
+    } else if (aux_a > 0 && aux_b == 0) {
+        return 0;
+    } else if (aux_a == 0 && aux_b == 0) {
+        return bam1_lt_core(a,b);
+    } 
+
+    if (*aux_a < *aux_b) {
```
daviesrob: This doesn't quite work, as for integer tags `*aux_a` and `*aux_b` can have different types (say c and i), but still be comparable.  You might want to cover the integer case before this (and do we want to allow comparison of floats and integers?)

bam_sort.c
```diff
@@ -1611,13 +1633,71 @@ static int change_SO(bam_hdr_t *h, const char *so)
...
+        return 1;
+    } else if (*aux_a > *aux_b) {
+        return 0;
+    } else {
+        if (*aux_a == 'c' || *aux_a == 'C' || *aux_a == 's' || *aux_a == 'S' || *aux_a == 'i' || *aux_a == 'I') {
+            int32_t va = bam_aux2i(aux_a);
```
daviesrob: Should be `int64_t`.

daviesrob: Should be `int64_t`.

bam_sort.c
```diff
@@ -1756,6 +1837,12 @@ int bam_sort_core_ext(int is_by_qname, const char *fn, const char *prefix,
 
     if (n_threads < 2) n_threads = 1;
     g_is_by_qname = is_by_qname;
+    if (sort_by_tag) {
+        g_is_by_tag = 1;
+        g_sort_tag[0] = sort_by_tag[0];
+        g_sort_tag[1] = sort_by_tag[1];
```
daviesrob: `strncpy()` again.

bam_sort.c
pmarks: Our use case is currently for string tags that encode a barcode sequence, but I can imagine changing to an integer barcode for efficiency. Not sure about floats, but seems worth handling all in one shot.

bam_sort.c
pmarks: did a quick profile and indeed it's spending a ~60% of it's time in bam_aux_get(), so I will implement your suggestion of a wrapper struct that is pre-populated with the appropriate tag pointer.

pmarks:
@daviesrob: all comments addressed in f1885e0 - ready for another review. Thanks!

pmarks:
fixed a bug in merge, and added a merge test. (also resolved a conflict with a change to a test BAM from another commit to develop). ready for review.

pmarks:
@daviesrob ping

daviesrob:
Sorry, I've been a bit busy with some other stuff for a while.  But I haven't forgotten this, and hope to get it reviewed and in by the end of this week.

daviesrob:
This looks OK now, and the extra overhead for the tag sort doesn't look too bad.

I've added a few small updates:

- The tests were comparing to compressed bam files.  This can be a problem if a different version of zlib (for example [cloudflare's optimized one](https://github.com/cloudflare/zlib)) is used as the compressed data can be different even though the uncompressed data is the same.  I've changed them to use sam instead.  (As an aside, I see a few other expected bam files have sneaked in elsewhere.  They will also need to be changed.)

- I've allowed float tags to be compared to integer ones, as it's not too difficult.

- I've added documentation to the manual page, including a description of the sorting rules.

The revision history for this pull request is a bit messy, so I suspect I'll squash it on merging.  In general I think it's best if you don't try to merge changes from develop into pull requests as it keeps things cleaner.  We can easily sort out any issues at the point when we merge the whole thing.  If you really do want to be up to date, a better alternative is to rebase the whole thing and then force push it on top of the existing PR.

bam_sort.c
```diff
@@ -1664,7 +1664,7 @@ static inline int bam1_lt_core(const bam1_p a, const bam1_p b)
 uint8_t normalize_type(const uint8_t* aux) {
     if (*aux == 'c' || *aux == 'C' || *aux == 's' || *aux == 'S' || *aux == 'i' || *aux == 'I') {
         return 'c';
-    } else if (*aux == 'f' || *aux == 'F') {
+    } else if (*aux == 'f' || *aux == 'd') {
```
pmarks: Should this be (*aux == 'f' || *aux == 'F' || *aux == 'd') ?

pmarks:
@daviesrob great, thanks!  Noticed one thing in the aux type normalization - is 'F' a valid type that should be lumped in with 'f'?  Happy to squish on merge -- I'll be more careful w/ the history.  Anything else you need from me?

bam_sort.c
daviesrob: I don't think so.  [The sam specification](http://samtools.github.io/hts-specs/SAMv1.pdf) lists `AcCsSiIfZHB` as valid type specifiers on page 12.  `d` is a historic samtools/htslib extension for "double" - I don't think it's ever written, but support for reading it is there.

daviesrob:
We're pretty much there.

There is one other issue I noticed.  I'm not sure if it's right to relax the lost_coord_sort test in `bam_merge_core2()`.  When sorting by tag, if the secondary sort is by position it will still get messed up if the references are in a different order in the source files.

daviesrob:
Spotted a problem with merge by tag then name.  Should be sorted now.

daviesrob:
Merged as 3c7e138eccd60ac4bbceb1523eaabb0435fe0050.

Thanks for the pull request, but in future please don't merge in develop.  It took quite a lot of git-fu to get it into a reasonable state...

PR #682, opened 2017-05-15
Make dist and distcheck targets

jkbonfield:
This is partly to replace the c-maint repository and additionally to add the ability to validate the tar balls is constructed correctly by unpacking, building and testing.

It is excessively messy getting this to work with the joint samtools/htslib tarball due to the catch-22 nature of having config.mk specify the HTSDIR and being included and used within Makefile targets in a circular fashion.  However it does at least appear to work!

Differences to the c-maint one are the addiction of the acinclude.m4 (I am unsure if this is correct or not, but it's here while I mull it over) and the lack of changing time stamps within the tar.  It *does* guarantee that configure post-dates the rest of the tar ball however.


configure.ac
```diff
@@ -23,7 +23,7 @@
 # DEALINGS IN THE SOFTWARE.
 
 dnl Process this file with autoconf to produce a configure script
-AC_INIT([Samtools], m4_esyscmd_s([make print-version]),
+AC_INIT([Samtools], m4_esyscmd_s([./version 2>/dev/null]),
```
mcshane: `./version.sh`?

jkbonfield: Ah yes.  Curious it works though!  Thanks for the spot.

jkbonfield:
I haven't done bcftools yet but was waiting on feedback for the samtools equivalent first, and for the bcftools autoconf branch to be merged.

Any feedback on this so far?

Issue #485, opened 2015-11-11
Argument for index file


AlistairNWard:
Would it be possible to include a command line argument for the BAM index? Currently, assuming the index has FILE.bam.bai extension fails when tools, e.g. Picard have been used to generate an index of the form FILE.bai. If there is a command line argument, we can assume FILE.bam.bai in its absence, or use the value specified otherwise.

This is a significant overhead at the moment, so would really appreciate this simple fix.

mcshane:
This will be possible again in the 1.3 release (8f60492d1a81b362cf7630ee3ab8669fd21420e0) via:
```
samtools index /path/to/bam /path/to/index
```

AlistairNWard:
This doesn't resolve the issue. This lets me create any index but not use the index. I need to run a command:

samtools view /path/to/bam region

This is only possible for an indexed BAM and the assumption is that the index is FILE.bam.bai. If the index is FILE.bai, I cannot view this region. I need to be able to use the argument:

samtools view -x FILE.bai FILE.bam region

so that the index file can still be used.

mcshane:
samtools will always look for look for both `FILE.bam.bai` and `FILE.bai`.
```
$ samtools view test/dat/test_input_1_a.bam ref1:7-7
[main_samview] random alignment retrieval only works for indexed BAM or CRAM files.
```
```
$ samtools index test/dat/test_input_1_a.bam test/dat/test_input_1_a.bai
$ samtools view test/dat/test_input_1_a.bam ref1:7-7
r001	163	ref1	7	30	8M4I4M1D3M	=	37	39	TTAGATAAAGAGGATACTG	*	XX:B:S,12561,2,20,112	YY:i:100	RG:Z:fish	PG:Z:colt
```

AlistairNWard:
Thanks for the help Shane, I'm actually thinking of a different scenario. A use case we have is streaming a BAM file through samtools via https, where the bam file name is obfuscated. Samtools then assumes an index of the same name with .bai or .bam.bai appended. But given the obfuscation, this file may not exist. So, we would want to supply both the BAM and index files on the command line. Does this make sense?

adamspargo:
Can you not obfuscate the name before you index the BAM, so that the index 
has the same name? Or maybe I'm missing something.

On Thu, 12 Nov 2015, Alistair Ward wrote:

> 
> Thanks for the help Shane, I'm actually thinking of a different scenario. A use case we have is streaming a BAM file
[...]

PR #686, opened 2017-06-07
Add option for samtools view to specify the index file explicitly

ramyala:
This is useful when index files aren't in the same folder.

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -63,6 +63,7 @@ typedef struct samview_settings {
     int flag_alloff;
     int min_qlen;
     int remove_B;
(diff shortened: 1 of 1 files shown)


jmarshall:
It would be good for functionality like this to be designed such that it can be extended to commands with multiple SAM/BAM/CRAM input files.  See discussion in #485.

smrgit:
is there any hope for this being merged/implemented?  it seems like this has been discussed for quite a long time... @jmarshall ?

valeriuo:
Implemented as https://github.com/samtools/samtools/pull/978

PR #691, opened 2017-06-15
Fastq illumina

jenniferliddle:
Added -i option to add illumina style tags to the fastq header lines
Added option to produce .gz or .bgzf format fastq files
Added -c flag to set compression level for above compressed files.


mcshane:
Maybe `.bgz` instead of `.bgzf` for the file extension? I've seen the former in the wild, but not the latter...

jenniferliddle:
Let's handle both .bgz and .bgzf just to be sure.

PR #700, opened 2017-06-28
Limited the number of LN tag reads to one, inside a SQ line.

valeriuo:
Added a flag in order to detect the LN tag match and prevent secondary readings of this tag on the same SQ line of a SAM header.

### DIFF
--- a/padding.c
+++ b/padding.c
@@ -382,6 +382,7 @@ bam_hdr_t * fix_header(bam_hdr_t *old, faidx_t *fai)
 {
     int i = 0, unpadded_len = 0;
     bam_hdr_t *header = 0 ;
+    unsigned short ln_found;
 
     header = bam_hdr_dup(old);
     for (i = 0; i < old->n_targets; ++i) {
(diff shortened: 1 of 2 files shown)


padding.c
```diff
@@ -382,6 +382,7 @@ bam_hdr_t * fix_header(bam_hdr_t *old, faidx_t *fai)
 {
     int i = 0, unpadded_len = 0;
     bam_hdr_t *header = 0 ;
+    unsigned short ln_found = 0;
```
daviesrob: Move to underneath line 408 so this gets reset on each header line

padding.c
```diff
@@ -421,10 +422,11 @@ bam_hdr_t * fix_header(bam_hdr_t *old, faidx_t *fai)
 
             /* Parse the @SQ lines */
             while (cp != end) {
-                if (end-cp >= 2 && strncmp(cp, "LN", 2) == 0) {
+                if (!ln_found && end-cp >= 2 && strncmp(cp, "LN", 2) == 0) {
```
daviesrob: This will ignore any extra LN tags, so they will be copied into the new file by lines 447 to 452.  It might be better to skip them and not copy them over.

padding.c
```diff
@@ -421,10 +422,11 @@ bam_hdr_t * fix_header(bam_hdr_t *old, faidx_t *fai)
...
                     // Rewrite the length
                     char len_buf[100];
                     int tid;
+                    ln_found = 1;
                     for (tid = 0; tid < header->n_targets; tid++) {
                         // may want to hash this, but new header API incoming.
                         if (strncmp(name, header->target_name[tid], name_end - name) == 0) {
```
daviesrob: When printing the replacement string, we still need to ensure that it's shorter than the original.  The best way might be to copy cp into another pointer, and move the loop at line 438 up higher so we know where the tab is before trying  to make the replacement.  Then check that the return value of sprintf() is less than or equal to the difference between the two pointers.

PR #711, opened 2017-07-27
Fix #708 by enforcing a check between index file and format

### LINKED ISSUE #708: samtools fasta core dumps using --i1 without --index-format
samtools fastq core dumps if you use the --i1 option but don't specify the -index-format option.

I suggest adding a default to --index-format defaults of "i*"

valeriuo:
I added a supplementary check to notify the user that the index format is missing, in case an index file is provided.

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -1208,6 +1208,13 @@ static bool parse_opts(int argc, char *argv[], bam2fq_opts_t** opts_out)
         return false;
     }
 
+    if (nIndex==0 && opts->index_file[0]) {
+        fprintf(stderr, "index_format not specified, but index file given\n");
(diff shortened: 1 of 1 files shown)


valeriuo:
There are a couple of aspects here:

1. The validation mechanism of the command was already put in place and the addition of the extra check fits smoothly with the rest of the code. This check basically asks the user to provide a valid format. 

2. Setting a default format could open the door for unwanted behavior. For example, if people mistakenly type in the format as n\***j**10, the software will replace it with the default i*. And what should the default format be if two index files are provided, i* or i\*i\*?

jenniferliddle:
Both good points. I agree, make the user do the work of telling us what they want instead of guessing.

PR #715, opened 2017-08-06
Add optional QNAME column to mpileup text output

jmarshall:
We recently had a need to answer questions like “find all the reads with a T at genomic position NNN”. Thinking `mpileup` text output had an option to show read names, a quick 'n' dirty script parsing mpileup output sprang to mind — but it turns out `mpileup` doesn't do that.

There may be a better approach to answering such questions, but this is a simple one…


daviesrob:
Thanks for this, will merge shortly.

PR #716, opened 2017-08-08
Added markdup to mark duplicates.

whitwham:
samtools markdup takes a coordinate sorted file that has had samtools fixmates -m run on it.  Optionally duplicate alignments can be removed from the output.


bam_mate.c
```diff
@@ -218,8 +218,39 @@ static int sync_mate(bam1_t* a, bam1_t* b)
...
 
+
+static uint32_t calc_mate_score(bam1_t *b)
+{
+    uint32_t score = 0;
+    uint8_t  *qual = bam_get_qual(b);
+    int i;
+
+    for (i = 0; i < b->core.l_qseq; i++) {
+        if (qual[i] >= 15) score += qual[i];
```
tseemann: What's the significance of this hard-coded Q=15 here?

samtools.1
```diff
@@ -88,6 +88,8 @@ samtools addreplacerg -r 'ID:fish' -r 'LB:1334' -r 'SM:alpha' -o output.bam inpu
 samtools collate aln.sorted.bam aln.name_collated.bam
 .PP
 samtools depad input.bam
+.PP
+samtools markdup in.algnsorted.bam out.bam
```
tseemann: What does `algnsorted` mean?  Name or Coordinate?
See fixmate docs: 
`Usage: samtools fixmate <in.nameSrt.bam> <out.nameSrt.bam>`

bam_mate.c
whitwham: It is the quality cut-off value that is used by Picard and Biobambam for working out the mate score for duplicate marking.  Picard gets it from the samtools jdk.  I do not know why that value was chosen but it seems to be as close to a standard as we have.  There is a small comment about it in markdup where the values are used.  Proper documentation is still being written.

samtools.1
whitwham: From the man page: 

` Mark duplicate alignments from a coordinate sorted file that
  has been run through fixmate with the -m option.
`

bam_mate.c
tseemann: I would suggest putting it as a constant somewhere.

PR #724, opened 2017-09-18
Improve merge stage in sort

daviesrob:
Give sort its own merge function instead of calling into samtools merge.  This allows a lot of complexity about header merging to be stripped out, and means the output header no longer changes when sort has to merge files together.

The new function can also read bam records from memory as well as from files, so there is no longer any need to save the last set of part-sorted outputs.  This allows sort to use multiple threads when all of the data fits in memory.

The comparison function for merge is also made stable for the name and by_tag options (is already was for position).  This ensures that sort gives exactly the same output for cases where it needs to merge as those where it does not (i.e. single thread and everything fits in memory).  One of the by_tag merge test output files has to change as a result of this - it included lots of records which tied according to the ordering criteria.  Both outputs are valid according to the command-line parameters given, but the new one should give more reproducible results.  For example:

```
samtools sort -o /tmp/merge.p.1.sam test/dat/test_input_1_b.sam
samtools sort -o /tmp/merge.p.2.sam test/dat/test_input_1_d.sam
samtools merge -f -s 1 -c -p /tmp/merge.p.m.sam /tmp/merge.p.1.sam /tmp/merge.p.2.sam
samtools sort -t PG -o /tmp/merge.p.m.pg.1.sam /tmp/merge.p.m.sam
```
gives the same result as:
```
samtools sort -t PG -o /tmp/merge.p.1.pg.sam test/dat/test_input_1_b.sam
samtools sort -t PG -o /tmp/merge.p.2.pg.sam test/dat/test_input_1_d.sam
samtools merge -f -s 1 -c -p -t PG /tmp/merge.p.m.pg.2.sam /tmp/merge.p.1.pg.sam /tmp/merge.p.2.pg.sam
```
whereas before the ordering of `/tmp/merge.p.m.pg.1.sam` and `/tmp/merge.p.m.pg.2.sam` would be different.


bam_sort.c
```diff
@@ -126,26 +126,30 @@ typedef struct {
...
         int t;
         if (a.b.b == NULL || b.b.b == NULL) return a.b.b == NULL? 1 : 0;
-        t = bam1_lt_by_tag(b.b,a.b);
-        return t;
+        t = bam1_cmp_by_tag(a.b, b.b);
+        if (t != 0) return t > 0;
     } else if (g_is_by_qname) {
-        int t;
+        int t, fa, fb;
         if (a.b.b == NULL || b.b.b == NULL) return a.b.b == NULL? 1 : 0;
```
valeriuo: This line of code is duplicate for both cases. So we could put it above the if/else block. Also, I would rewrite it as:
```
if (a.b.b == NULL)
    return 1;
if (b.b.b == NULL)
    return 0;
```
In this way, there are at most two comparisons, while previously there were at least two.

bam_sort.c
```diff
@@ -126,26 +126,30 @@ typedef struct {
...
-
-static inline int bam1_lt_by_tag(const bam1_p a, const bam1_p b);
+static inline int bam1_cmp_by_tag(const bam1_p a, const bam1_p b);
 
 // Function to compare reads in the heap and determine which one is < the other
 static inline int heap_lt(const heap1_t a, const heap1_t b)
 {
     if (g_is_by_tag) {
         int t;
         if (a.b.b == NULL || b.b.b == NULL) return a.b.b == NULL? 1 : 0;
```
valeriuo: Could we have a bit more verbose variable names? I can see that neither `heap_t` nor `bam1_p` are exported, so maybe we can make clearer what `b` stands for in each case.
Also, `bam1_p` is defined in **bam_rmdup.c** as 
`typedef bam1_t *bam1_p;`

PR #731, opened 2017-10-25
Optionally mark supplementary alignments of duplicates as duplicates.

whitwham:
This should allow samtools markdup to conform more closely to [this](https://github.com/CCDG/Pipeline-Standardization/blob/master/PipelineStandard.md#duplicate-marking) proposal.  Added LZ4 to allow fast reasonably sized temporary files used in markdup.


tmp_file.h
```diff
@@ -0,0 +1,121 @@
+/*
+    tmp_file.h - write to and read from a temporary binary file
+    for fast storage plus added compression.
+    
+    Andrew Whitwham, August 2017
```
mp15: This file needs a specific copyright statement and to indicate who the copyright holder is in addition to the author.

tmp_file.c
```diff
@@ -0,0 +1,445 @@
+/*
+    tmp_file.c - write to and read from a temporary binary file
+    for fast storage plus added compression.
+
```
mp15: This file needs a specific copyright statement and to indicate who the copyright holder is in addition to the author.

bam_markdup.c
```diff
@@ -825,10 +988,27 @@ int bam_markdup(int argc, char **argv) {
         hts_set_opt(in,  HTS_OPT_THREAD_POOL, &p);
         hts_set_opt(out, HTS_OPT_THREAD_POOL, &p);
     }
-
+
```
mp15: Please remove the whitespace.

bam_markdup.c
```diff
@@ -763,7 +917,10 @@ static int markdup_usage(void) {
     fprintf(stderr, "Option: \n");
     fprintf(stderr, "  -r           Remove duplicate reads\n");
     fprintf(stderr, "  -l           Max read length (default 300 bases)\n");
+    fprintf(stderr, "  -S           Mark supplemenary alignments of duplicates as duplicates (slower).\n");
     fprintf(stderr, "  -s           Report stats.\n");
+    fprintf(stderr, "  -t           Mark primary duplicates with the name of the original in a \'do\' tag."
+                                  " Mainly for information and debugging.\n");
```
mp15: Please also document -T option you added

daviesrob:
When opening a temp file, it might be better to make `tmp_file_open_write()` take a prefix, and get it to generate a file name.  That means you can open the file with `open(tmp->name, O_EXCL|O_CREAT, 0600)`.  This means you can keep trying until you manage to open a file descriptor with a unique name (or give up after too many tries).

On UNIX-like systems you might want to unlink the file after opening so that it gets cleaned up automatically if the program crashes.  On Windows, it looks like the approximate equivalent is [`CreateFile`](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363858(v=vs.85).aspx) with the `FILE_FLAG_DELETE_ON_CLOSE` flag.

y9c:
It seems that this is not compatible with UMI option?

whitwham:
> It seems that this is not compatible with UMI option?

If you have a problem please open a new issue with full details.  Commenting on a nearly eight year old pull request is not helpful.

PR #732, opened 2017-10-25
Make use of the multiple region iterator from HTSlib

valeriuo:
1. Filter the regions specified in the BED file by the regions provided via command line arguments.
2. Create a final sorted list of regions of interest and pass it to HTSlib, together with the file index, for creating a single iterator for the entire list. The iterator has a sorted list of file offsets that covers all the domain of interest, thus removing the need for seeking back and forth inside the alignment file.
3. Use the multiple region iterator in a single loop and avoid creating an iterator for each region and separate filtering against the BED file.


jkbonfield:
Squashed and merged.  Thanks.

PR #740, opened 2017-11-13
Mac-compatible completion

dmckean:
I've altered `samtools_tab_completion` to be compatible with bash 3 (which does not have the builtin `compopt` ) and macOS's (BSD's?) version of `xargs`, which seems to pass strings that look like options to `printf` as options.


daviesrob:
Thanks.  I checked on both Linux and a Mac, and this fixed the problem.  Now merged.

Issue #723, opened 2017-09-16
How to extract embedded reference from CRAM file?


tseemann:
If I have a CRAM file with `embed_ref=1` how do I get the reference out?

I can't find any `samtools` gateway to this, which I assume is probably in the `htslib` API ?

A use case for this is to for calling SNPs to a reference, where the reference is in the CRAM file you supply to the variant caller, removing the need to provide an external reference file. Or just for storing genome assemblies with the reads.

jkbonfield:
Sadly there isn't a way to do this, but I like the idea.

Note it may not be complete.  The reference stored is anywhere that data aligns against, but if we have a lack of coverage in the sequencing we'll get a lack of coverage in the stored reference.  That's probably fine for SNP calling as you don't need reference outside of sequenced regions.  We'd just need to pad it out with Ns.

tseemann:
Sorry - are you saying the embedded reference in a CRAM file is *not* the complete reference?

So using an external reference is *different* to using an embedded reference?

Say it ain't so!

jkbonfield:
It's the same reference, but not necessarily complete.

Imagine encoding a 10Mb region of chrX using an embedded reference.  Do we really want 3Gb of reference added to the CRAM file?  It's not helpful in any way!  So instead we have the reference for the 10Mb region we encode.

In practice, each random access CRAM slice contains its own little bit of reference.  This makes it fast and easy to fetch the relevant reference.  Each slice states it covers <ref>:<from>-<to> which normally would trigger a query to fetch the reference, by file, url, whatever, but here it finds it embedded so uses the locally embedded copy instead. 

Potentially (very likely) some slices will overlap slightly, due to reads overlapping, while some may have clear space between them due to lack of coverage.  That second case means there is no reference embedded at that point - just as with the example of a 10Mb chrX cram file.  That's what I mean about being incomplete.  It doesn't mean that the encoded records are in any way different to using an external reference.  The same data will be stored, we're just using an internally cached copy of the reference instead of the external one.

(Although note in theory we could generate the consensus and supply that as the reference to store.  It saves space as we're no longer encoding the same SNPs over and over again, but we have to offset that with the cost of storing NM/MD instead of computing them on-the-fly.)

tseemann:
Ok, this means it is unsuitable for my needs.  With bacteria, the reference is smaller than the data given the reference.   

[...]

Issue #720, opened 2017-09-16
CRAM reference search order doesn't include 'embed_ref`


tseemann:
http://www.htslib.org/doc/samtools.html says:
```
The search order to obtain a reference is:

Use any local file specified by the command line options (eg -T).
Look for MD5 via REF_CACHE environment variable.
Look for MD5 in each element of the REF_PATH environment variable.
Look for a local file listed in the UR: header tag.
```
Is it safe to assume that if my CRAM was created with `embed_ref=1` then it will use that first?

jkbonfield:
Yes that is a good point.  It should only search for the reference if it's not already embedded within the file itself.

There are some questions I  have to investigate regarding embed_ref though before documenting it.  The format only permits one reference per slice, but in some situations it is beneficial to encode data for multiple references in the same slice, to avoid creating too many slices (name sorted data is the obvious one, but this also crops up with highly fragmented assemblies).  I cannot recall how I dealt with those corner cases.

tseemann:
Is this the only documentation on "slices" ?  I don't understand what they do/are.

```
multi_seq_per_slice=0|1
CRAM output only; defaults to 0 (off). By default CRAM generates one container per reference sequence, except in the case of many small references (such as a fragmented assembly).

seqs_per_slice=INT
CRAM output only; defaults to 10000.

slices_per_container=INT
CRAM output only; defaults to 1. The effect of having multiple slices per container is to share the compression header block between multiple slices. This is unlikely to have any significant impact unless the number of sequences per slice is reduced. (Together these two options control the granularity of random access.)
```

jkbonfield:
See pic5 at the top of page 6 of the CRAM spec: https://github.com/samtools/hts-specs/blob/master/CRAMv3.pdf

I tihnk the defaults are probably fine for you unless you intend on doing a lot of fine grained random access, in which case seqs_per_slice could be reduced (eg to 1000).  I think I concluded it's 5-10% larger with 10fold reduction in slice size.  (Conversely larger slice sizes don't get much beyond 2% smaller.)

However feel free to experiment!  I know you'll report any bugs :)

PR #748, opened 2017-12-04
Add fastaref subcommand.

ThomasHickman:
This PR adds a fastaref subcommand, which creates a fasta file containing references for a given file, obtained by looking at the SQ line. Note: this depends on https://github.com/samtools/htslib/pull/589 getting merged, so the travis build will fail until that PR gets merged.


ThomasHickman:
To fit the format of all the other samtools command, the usage is now:
```
samtools fastaref [options] <in.bam>

Options:
  -o FILE  Output file name [stdout]
  -k LIST  Output the specified keys into the fasta file (separated by commas)  [LN,AH,AN,AS,M5,SP]
  -l INT   Maximum length of outputted lines [60]
```

bam_fastaref.c
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * Given a pointer to the start of a sam field, populates samField with a
+ * the malloced sam field.
+ * 
+ * Updates samFieldPosition to be the position after the sam field.
+ * 
+ * @returns 1 if successful
+*/
+int get_sam_field(char** samFieldPosition, char** samField){
```
valeriuo: Why don't you return `samField` as a result?

bam_fastaref.c
```diff
@@ -0,0 +1,347 @@
...
+    str->l = str->m = 0; str->s = NULL;
+}
+
+/**
+ * Given text of an SQ line beginning at sequenceText, populates values for the SN and M5
+ * fields and returns selected portions of the header according to options
+ * 
+ * This updates sequenceText after reading it.
+ */
+int getInfoForSQLine(char** sequenceText, char** SN, char** M5, kstring_t* restHeader, FastarefOptions *options){
```
valeriuo: If you just want to parse the `@SQ` line, you should look into the functions `sam_hdr_parse` (`htslib/sam.c`), `sam_hdr_parse_` (`htslib/cram/sam_header.c`) and `bam_header_to_cram` (`htslib/cram/cram_samtools.c`). 
Unfortunately, we don't have an unified and homogeneous header API (it's high on our TO-DO list), but there is plenty of functionality scattered around, which you can use to implement your feature. There is no need to rewrite it.

ThomasHickman: @valeriuo There doesn't seem to be any functions in `htslib/cram/sam_header.c` to get all headers which a specific tag (it looks like `sam_hdr_find` only returns the first header item). Would the expected thing to do be to expose the whole of the `SAM_hdr` data structure to `bam_fastaref.c`, or add a function to `sam_header.c` to expose this functionality?

valeriuo: @ThomasHickman, check `sam_hdr_find_key` function (`sam_header.c`), with an example in `cram_io.c:1869`. Whatever is in `sam_header.h` is already exposed by HTSlib and you can use it in samtools.

jkbonfield: Sadly that function isn't exposed.  Only the htslib/*.h files are considered as external API.  The cram/*.h are internal and shouldn't be used outside htslib.  (Although in one case I did, out of sheer frustration!)

This is doubly annoying because we *do* have a header API, namely in the cram subdirectory, but not one blessed by htslib as OK to use.  This header API was derived from io_lib and arrived when I ported over CRAM, simply because I needed one for my own work.  We do really need to write down a list of requirements and then evaluate what we have to see if the current code fits it or whether we want to start afresh / amend it.

ThomasHickman:
I've created the function `parseSQLine`, which generates a khash containing the fields of the SQ line and changed it according to proposed changes

bam_fastaref.c
```diff
@@ -0,0 +1,350 @@
...
+    int keysStrLength = strlen(keysStr);
+
+    if((keysStrLength + 1) / 3 * 3 != keysStrLength + 1){ // keysStrLength + 1 is not divisible by 3
+        fprintf(stderr, "%s is not a valid set of keys\n", keysStr);
+        return NULL;
+    }
+
+    *numKeys = (keysStrLength + 1) / 3;
+    char** outputKeys = malloc(*numKeys * sizeof(char*));
+
```
valeriuo: Check the result of `malloc`.

bam_fastaref.c
```diff
@@ -0,0 +1,350 @@
...
+#include "htslib/sam.h"
+#include "htslib/ref.h"
+#include "htslib/kstring.h"
+#include "htslib/khash.h"
+#include "htslib/hfile.h"
+#include "htslib/cram.h"
+#include "sam_header.h"
+
+KHASH_INIT(s2s, char*, char*, 1, kh_str_hash_func, kh_str_hash_equal)
```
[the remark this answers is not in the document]
ThomasHickman: I used this as `KHASH_MAP_INIT_STR` specifies `const char *` as the type of the keys. The keys which I specify are dynamically allocated, so need to be freed and therefore need to be declared as a `char *`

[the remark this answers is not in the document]
valeriuo: Agree, but we usually go around this limitations with `strdup`, for setting the key, and explicit cast to `(char *)` when freeing the key memory space. Take a look at `bedidx.c:207` and `bedidx:252`. The idea is that keys in a map should be constant.

bam_fastaref.c
```diff
@@ -39,96 +39,68 @@ DEALINGS IN THE SOFTWARE.  */
...
-
-    while(1){
-        fieldEnd = strpbrk(fieldStart, "\t\n\0");
-
-        int valueLength = fieldEnd - fieldStart - 3;
-        if(valueLength < 1){
+kh_s2s_t* parseSQLine(char* sqLine){
+    char* parsePosition = sqLine + 4; // 4 = len("@SQ\t")
+    char* field;
+    kh_s2s_t* khash = kh_init(s2s);
```
valeriuo: I couldn't find the code that frees the memory allocated to `khash`.

bam_fastaref.c
```diff
@@ -39,96 +39,68 @@ DEALINGS IN THE SOFTWARE.  */
...
+
+    while((field = strsep(&parsePosition, "\t")) != NULL){
```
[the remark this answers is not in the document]
ThomasHickman: @jrandall suggested that I used a c function here, but I don't think either me or him realised that this was C99 compliant. I'll revert back to the previous implementation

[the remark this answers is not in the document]
ThomasHickman: Added an implementation of `strsep` instead, as it keeps code readability, and doesn't change what I've written too much

bam_fastaref.c
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    header = sam_hdr_read(inFile);
+    if (header == NULL) {
+        print_error("fastaref", "failed to read header for \"%s\"", options->samFileName);
+        returnCode = 1;
+        goto cleanup;
+    }
+
+    char* readPosition = header->text;
```
valeriuo: Your parsing logic operates directly on the `text` string of the header structure. Not only does `_strsep` change the original string, but it could make the initial text pointer **NULL**, if it doesn't find a delimiter. You would then leak memory. 
I suggest making a copy of the header text (with `strdup`) and work on the copy. You should also save the initial pointer of the copy and use it at the end, when freeing the memory allocated to the copy. 
Thus, you decouple your logic from the header layer.

valeriuo:
The code looks OK and I was able to build it. However, we'll wait for https://github.com/samtools/htslib/pull/589 to be merged in order to do some testing, before merging this PR in develop.

jmarshall:
The man page section added just says

> samtools fastaref [options] <in.bam>
>
> Generates a fasta reference file, by reading SQ lines.

It's really unclear what this new command actually does. Presumably it uses the CRAM reference-finding machinery to output the reference bases in FASTA form, looking them up based on `M5` tags in the `@SQ` headers? What if your BAM file doesn't have `M5` tags?

It would be great if the man page described all this.

The options list has a `-k`/`-o` typo. And what you're calling _keys_ everything else in the user interface calls _(header) tags_.

ThomasHickman:
> It's really unclear what this new command actually does. Presumably it uses the CRAM reference-finding machinery to output the reference bases in FASTA form, looking them up based on M5 tags in the @SQ headers? What if your BAM file doesn't have M5 tags?

This is an accurate description of what this is trying to do. If the reference finding code fails to find a reference (maybe due to their being no `SQ` lines), this command fails with an error printed to stdout. I could have a look at changing the documentation to be a bit clearer along with fixing https://github.com/samtools/htslib/pull/589 when I get time to have a look at this.

tseemann:
@ThomasHickman will this extract an *embedded* reference inside a CRAM file?

ThomasHickman:
@tseemann That sounds like an interesting idea, but this doesn't support extracting an embedded reference from a cram at the moment - this just uses the M5 field in the SQ line to query the `REF_PATH`, `REF_CACHE` or query `https://www.ebi.ac.uk/ena/cram/md5/`

tseemann:
The reason I ask is that it isn't actually possible to do right now.

See: 
1. https://github.com/samtools/samtools/issues/723
2. https://github.com/samtools/samtools/issues/720

PR #752, opened 2017-12-18
[Proposal/don't merge] using GNU guile as a plugin system for 'samtools view'.

lindenb:
Background:

> Wikipedia: GNU-Guile ( https://gnu.org/software/guile/ ) is the preferred extension system for the GNU Project which features an implementation of the Scheme programming language.  (...) For extending programs, Guile offers "libguile" which allows the language to be embedded in other programs, and integrated closely through the C API; similarly, new types and subroutines defined through the C API can be made available as extensions to Guile itself (...)  

for example , Guile is used as a plugin for GNU make: https://www.gnu.org/software/make/manual/html_node/Guile-Integration.html

The current exprimental PR adds a `--with-guile` to `configure.ac` and provides a few bindings to map the hts-lib C functions to scheme/guile (e.g: https://github.com/lindenb/samtools/blob/pl_guile/htsguile.c#L17 )

now , `samtools view ` has a new option `-g <scheme-code>` where the user can insert a custom scheme code for filtering the reads.

Here is a working example where the user provides a function  `(read-filter R)` that will only keep the reads having a length greater than 10 and containing the sequence `GCCTTGGCTCTTGTTCC`.

```
./samtools view   -g \
   '(define read-filter (lambda (R) (and (> (hts-read-length R ) 10 ) (integer? (string-contains  (hts-read-seq R) "GCCTTGGCTCTTGTTCC")))))' input.bam

```

Is the samtools community interested by such plugin ?

 if yes, I'll continue to work on this, if not, fair enough,  I'll move this into a standalone C tools :-)

best,

Pierre


kyleabeauchamp:
This looks very cool.  However, I would say for my needs, I would be more eager to have improved interoperability (and features) with pysam, rather than supporting an embedded scripting language.

lindenb:
moving to https://github.com/lindenb/htsguile

lh3:
IMHO, Lua is a much nicer language for the embedding purpose.

Issue #753, opened 2017-12-29
faidx - argument list is too long.


zeeev:
Hi,

It looks like I'm hitting the argument limit for command line options (~2K). Is it possible to add a file of sequence names/positions? I'd be happy to implement this change, but wanted to check before doing a PR.

Thanks,

--Zev

billzt:
It is very easy to use xargs:
```
xargs --arg-file=[region.file] samtools faidx [refer.fa]

zeeev:
@billzt sure there are many alternative solutions.

PR #754, opened 2017-12-30
added fn to faidx

zeeev:
Hi,

This pull request corresponds to issue #753.  I've added the ability to pass a file to samtools faidx. This modification does not change the UI. below is an example where samtools is passed both a plain text string and a file containing regions to pull.

`samtools faidx contigs.fasta test.txt $(cat test2.txt)`

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -28,9 +28,55 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdlib.h>
 #include <stdio.h>
 #include <unistd.h>
+#include <string.h>
 
 #include <htslib/faidx.h>
 #include "samtools.h"
+
+int print_fasta(faidx_t *fai, char * region){
+  int seq_len;
+  int ex = EXIT_SUCCESS;
+  char *seq = fai_fetch(fai, region, &seq_len);
+  if ( seq_len < 0 ) {
+      fprintf(stderr, "Failed to fetch sequence in %s\n", region);
+      ex = EXIT_FAILURE;
+      return ex;
+  }
+  printf(">%s\n", region);
+  size_t i, seq_sz = seq_len;
+  for (i=0; i<seq_sz; i+=60)
+  {
+      size_t len = i + 60 < seq_sz ? 60 : seq_sz - i;
+      if (fwrite(seq + i, 1, len, stdout) < len ||
+          putchar('\n') == EOF) {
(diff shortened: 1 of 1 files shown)


faidx.c
```diff
@@ -73,26 +119,26 @@ int faidx_main(int argc, char *argv[])
...
+     * needs to be the same, i.e. space delimited. This allows for a large
+     * number of regions. A file or a region can be passed on the command line.
+     * This maintains the UI and solves the "Argument too long" issue.
+     */
+      if( access( argv[optind], F_OK ) != -1 ) {
+        FILE * fn = fopen(argv[optind], "r");
+        char * regions  = read_string_from_fp(fn, 2000);
+        fclose(fn);
+
+        char* token = strtok(regions, " ");
```
lindenb: Hi  Zev,
`strtok` is harmful, you shouldn't use it because you have no guarantee that `print_fasta` will not call `strtok` and change your internal strtok state.

zeeev: Hi Pierre, 
I can see how the static pointer could cause trouble down the line. Would you support the use of strtok_r? Thanks for noticing the potential problem.

daviesrob:
I think it would be better to use an option to specify the list of regions.  Testing to see if the region argument is a file could easily catch unwary users out.  For example, `samtools faidx GRCh38_15.fa chr1 > chr1` would result in an empty file instead of the expected dump of chromosome 1.  Also, calling `access()` on each region may be quite slow for users of distributed filesystems (notably Lustre, but likely others as well).

As an alternative to a new option, you could prefix the file names with a symbol.  The tricky part is choosing which symbol to use, as pretty much any non-whitespace character can occur in a fasta file entry name.  An equals sign (`=`) might be the best choice -- it's less likely to be the first character of a region name due to not being compatible with SAM and the shell won't try to expand it (which would be the case with `*`).

For reading the file, it would be preferable to use `hopen()` (from `htslib/hfile.h`) and related functions so that you can read from URLs as well as ordinary files.  For reading the file, you could use `kgetline()` (from `htslib/kstring.h`) coupled with `hgets()` instead of the `read_string_from_fp()` function.  You can use `kstrtok()` to tokenize the lines.

It looks like you're only reading one line from the file.  This seems a bit unusual -- I would normally make this sort of file with one region per line, so I'd expect the whole file to be read.

whitwham:
Alternative resolution with pull request #840.

Issue #758, opened 2018-01-09
samtools fastq with -O option


bintriz:
When I run samtools fastq with -O option for a bam file which has OQ tag, it gives segmentation fault.
When I run samtools fastq with -O option for a bam file which doesn't have OQ tag, it gives quality  score with " as follows.

@H5FYCALXX:5:1208:1733515:0
AAAGAGATCTGATAAGTCCCAGGACTTCAGAAGAGCTGTGAGACCTTGGCCAAGTCACTTCCTCCTTCAGGAACATTGCAGTGGGCCTAAGTGCCTCCTCTCGGGACTGGTATGGGGACGGTCATGCAATCTGGACAACATTCACCTTTA
+
""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
@H5FYCALXX:5:1208:1735085:0
AAAGAGATCTGATAAGTCCCAGGACTTCAGAAGAGCTGTGAGACCTTGGCCAAGTCACTTCCTCCTTCAGGAACATTGCAGTGGGCCTAAGTGCCTCCTCTCGGGACTGGTATGGGGACGGTCATGCAATCTGGACAACATTCACCTTTA
+
""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
@H5FYCALXX:2:1118:2302059:0
GATAAGTCCCAGGACTTCAGAAGAGCTGTGCGACCTTGGCCAAGTCACTTCCTCCTTCAGGAACATTGCAGTGGGCCTAAGTGCCTCCTCTCGGGACTGGTATGGGGACGGTCATGCAATCTGGACAACATTCACCTTTAAAAGTTTATT

My exception is printing quality score as it is if OQ is not available.

jenniferliddle:
Oh dear. It looks as if the -O option is missing from our standard unit test suite.
I'll take a look...

jenniferliddle:
Thanks for the bug report.
I've reproduced the problem, and fixed it in PR #759

PR #759, opened 2018-01-09
Fixed bug in -O option to samtools fastq

jenniferliddle:
This fixes the bug reported in issue #758


sam_view.c
```diff
@@ -1133,15 +1133,12 @@ static bool bam1_to_fq(const bam1_t *b, kstring_t *linebuf, const bam2fq_state_t
...
-        if (oq) {
-            oq++;
-            qual = strdup(bam_aux2Z(oq));
-            if (!qual) goto fail;
-            if (b->core.flag & BAM_FREVERSE) { // read is reverse complemented
-                reverse(qual);
-            }
+    if (state->use_oq) oq = bam_aux_get(b, "OQ");
+    if (oq) {
+        qual = strdup(bam_aux2Z(oq));
```
valeriuo: We should check the return value of `bam_aux2Z`. If the type differs from **Z** or **H**, `strdup` will still segfault.

jkbonfield:
Merged as c984bd7a0e1c3e5a0a56de096f13749a09bf5292

PR #766, opened 2018-01-31
Add quiet mode to calmd

colindaven:
In SLURM, the voluminous fillmd comments (1 per read) are put into Stderr. These are more debug information and not informative for end users.

I added a -Q option for quiet mode.

Tested on a local server, the compiled version a) can be toggled on and off successfully with the -Q switch and b) generates a BAM file identical to the normal samtools (1.4.1 release via conda).

I was not sure about the following lines, but the rest seemed quite straightforward.

```
void bam_fillmd1(bam1_t *b, char *ref, int flag, int quiet_mode)
{
    bam_fillmd1_core(b, ref, INT_MAX, flag, 0, quiet_mode);
}
```

### DIFF
--- a/bam_md.c
+++ b/bam_md.c
@@ -46,7 +46,7 @@ DEALINGS IN THE SOFTWARE.  */
 
 int bam_aux_drop_other(bam1_t *b, uint8_t *s);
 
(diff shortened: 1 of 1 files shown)


valeriuo:
Merged in a222d100ae1062d04811ba6aafac93a3d271bc5e

PR #770, opened 2018-02-01
Make addreplacerg choose output format by file extension

### LINKED ISSUE #767: Addreplacerg ignores output filename and produces SAM
Despite providing a .bam output filename, addreplacerg seemingly produces a SAM output file. Possibly an omission from #461?

daviesrob:
Add missing call to sam_open_mode()

Fixes #767

### DIFF
--- a/bam_addrprg.c
+++ b/bam_addrprg.c
@@ -358,7 +358,9 @@ static void orphan_only_func(const state_t* state, bam1_t* file_read)
 }
 
 static bool init(const parsed_opts_t* opts, state_t** state_out) {
+    char output_mode[8] = "w";
     state_t* retval = (state_t*) calloc(1, sizeof(state_t));
+
     if (retval == NULL) {
         fprintf(stderr, "[init] Out of memory allocating state struct.\n");
         return false;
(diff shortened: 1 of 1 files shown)


bam_addrprg.c
```diff
@@ -374,7 +376,9 @@ static bool init(const parsed_opts_t* opts, state_t** state_out) {
     retval->input_header = sam_hdr_read(retval->input_file);
 
     retval->output_header = bam_hdr_dup(retval->input_header);
-    retval->output_file = sam_open_format(opts->output_name == NULL?"-":opts->output_name, "w", &opts->ga.out);
+    if (opts->output_name) // File format auto-detection
+        sam_open_mode(output_mode + 1, opts->output_name, NULL);
```
valeriuo: We should check the return value of `sam_open_mode` and print an error message, if there are issues with detecting the format.

daviesrob: This is only guessing the format based on the file suffix.  I don't think it should be an error if it fails.  Even a warning would be quite annoying, especially if the file being opened is "-".

PR #779, opened 2018-02-20
Adds the option of effectively removing the depth limit.

### LINKED ISSUE #764: Running samtools with no -d
Is it possible to run samtools with no -d filter?

We want to collect un-truncated depth stats?

valeriuo:
Fixes #764


tseemann:
I like this, thanks for the patch @valeriuo !

PR #789, opened 2018-03-14
Stats only for included regions

### LINKED ISSUE #640: Samtools stats does not limit coverage calculation to the target specified with -t
Hi,

When running `samtools stats` and specifying the target regions with the `-t` option, the reported coverage histogram is not strictly limited to the target regions.  Instead, it reports coverage for the whole length of all reads which overlap at all with the target regions, including portions of the reads which extend outside the target.

Is this by design? Would it be possible to limit the coverage calculation to only the parts of the reads within the target?

valeriuo:
Adds a CLI option (-W, --within) to select only the reads fully included in the target regions (supplied via the target file). Works only in conjunction with a target file (-t). The default is to select the overlapping regions (previous behavior).
Fixes #640


ChenYuelong:
Hi~

Valeriuo,

I`m sorry to bother you that I have not describe my question about   [#640](https://github.com/samtools/samtools/issues/640) clearly. 

I made a example to explain it.

If I give a target.bed like
>chr17	48745061	48745063

and I use `samtools depth` to stat the depth was 
>chr17	48745062	60
>
>chr17	48745063	59

but in the result of `samtools stat`,I may get `COV` result like

>COV	[1-1]	1	5

>COV	[3-3]	3	14

>COV	[4-4]	4	2

>COV	[5-5]	5	5

>COV	[6-6]	6	2

>COV	[7-7]	7	2

>COV	[8-8]	8	2

>COV	[9-9]	9	1

>COV	[10-10]	10	4

>...

----
but in my opion,the result should be like 

>COV [59-59] 59 1

>COV [60-60] 60 1


-----
For your commit,the result may be another format like

>COV [1-1] 1 0

>...

> COV [1000-1000] 1000 0

because no read was completely in the bed region only 2bp length.

**Sorry to text answer you because my pool skill of C  and English**

PR #794, opened 2018-03-15
Fix travis to use owner/htslib <branch> or upstream/htslib <develop>.

jkbonfield:
This also removes the HTSREPO custom travis  environment variable, but I don't think this is necessary now,

Should someone wish to actually fork samtools/htslib for their own development, not just for making PRs, then they'll simply need to change the hardcoded samtools/htslib.git part.

The philosophy is if I make an edit to samtools and push it to jkbonfield/samtools:FOO without a corresponding jkbonfield/htslib:FOO then it should test my PR against the copy of htslib where it will land once the PR is merged - ie samtools/htslib:develop.  If on the other hand I make an edit to htslib and a corresponding edit to samtools, pushing to jkbonfield/htslib:FOO and jkbonfield/samtools:FOO then I need my samtools PR testing in the context of my corresponding htslib PR, otherwise it will fail.

I updated the travis logic to match the appveyor code in this regard, instead of always using jkbonfield/htslib:develop if jkbonfield/htslib:FOO doesn't exist.


jmarshall:
`dirname $TRAVIS_REPO_SLUG` is a nice trick, but you're discarding a couple of advantages of the separate _.travis/clone_ script:

* you can look at the travis log and see the exact git clone command that is being run

* `--depth=1` speeds up travis builds (though possibly making it 10 or 50 or so would help with the occasional race conditions when you merge a PR before travis has run its build)

jkbonfield:
I meant to copy over the --depth=1 option but forgot, although the speed difference is pretty tiny compared to the overall build.

You can usually tell which remote it used already, albeit not so obvious.  Eg:

```
fatal: Remote branch travis_fix not found in upstream origin
```

That's from the initial clone, so it used the second one.  Agreed it's not quite so verbose.  Possibly adding something like `(cd $HTSDIR; git remote -v)` would be a suitable workaround to explicitly log which repo and branch was fetched?

jmarshall:
It's positively obtuse :smile:

That separate script exists precisely because I couldn't find a good way to log this vital information satisfactorily from directly within _.travis.yml_. Why not just change its HTSREPO argument to $TRAVIS_REPO_SLUG and do the cleverness inside the shell script? Is it so terrible having a _.travis_ directory? (At the time, it seemed likely that other _.travis_ scripts would follow, e.g. for building binary packages.)

jkbonfield:
Fixed by #798.

PR #797, opened 2018-03-16
Stats computed only for bases within a target region

### LINKED ISSUE #640: Samtools stats does not limit coverage calculation to the target specified with -t

valeriuo:
Fixes #640


jkbonfield:
I'm still looking through the code changes, but on testing it I see it works for -t regions.bed but not for regions specified on the command line.  The two should be consistent with each other.

stats.c
```diff
@@ -222,6 +227,10 @@ static void error(const char *format, ...);
 int is_in_regions(bam1_t *bam_line, stats_t *stats);
 void realloc_buffers(stats_t *stats, int seq_len);
 
+static int regions_lt(const void *r1, const void *r2) {
+    int from_diff = ((pos_t *)r1)->from - ((pos_t *)r2)->from;
```
jkbonfield: pos_t->from is uint32_t.  Therefore subtracting any two values would be potentially 33bits long.  Maybe this should be `int64_t diff` to accommodate that, and any realistic future change where we make chromosomes more than 32-bit long.

stats.c
```diff
@@ -1676,26 +1790,57 @@ int main_stats(int argc, char *argv[])
...
-        int i;
-        for (i=optind; i<argc; i++)
-        {
-            hts_itr_t* iter = bam_itr_querys(bam_idx, info->sam_header, argv[i]);
-            while (sam_itr_next(info->sam, iter, bam_line) >= 0) {
-                if (info->split_tag) {
-                    curr_stats = get_curr_split_stats(bam_line, split_hash, info, targets);
-                    collect_stats(bam_line, curr_stats);
+        if(targets)
+            error("Both a target file and target regions have been specified. Please choose only one method.\n");
```
jkbonfield: Complex though it is, specifying both regions and targets is quite common place in some other commands.

Historically regions have been used to do index skipping with targets files doing filtering while streaming, so if there are a lot of regions to work over for chr20 the user may use `20` as the region (to stream the whole chromosome) with the targets file to filter the output.

tseemann:
Personally I would prefer `MIN()` and `MAX()` be written as (inline) `C` functions with typed parameters, rather than the macro versions with their potential side-effects. I would assume the compiler would do quite well still when optimizing it. Alternatively you could use safe GNU C macros like this:
``
#define min(a,b) \
  ({ typeof (a) _a = (a); \
  typeof (b) _b = (b); \        //Min defined
  _a < _b ? _a : _b; })
```

daviesrob:
Unfortunately both `typeof()` and statement expressions are gcc extensions.  As we aim to target C99, we can't allow them.

The only real problem with the MAX and MIN macros is that they evaluate the inputs twice.  I do see that MIN is used with `bam_endpos()`, which iterates through the CIGAR data.  We should put the value into a variable as evaluating it twice (or in fact three times) is a bit wasteful.

jkbonfield:
Squashed and merged as 9ce8c64493f7ea3fa69bc5c1ac980b1a8e3dcf1f

PR #818, opened 2018-04-10
Fast paired collate

### LINKED ISSUE #781: fast paired collate
Could a tool be created/modified to produce collated pairs, read 1 and read 2 primary only, in a slightly less I/O intensive fashion?

Admittedly I don't know the internals of `samtools collate` but an option to indicate reads should be output on detection of 2nd read (still spilling to disk when distant) that could be run as follows would be a great addition:

```
samtools view -uF 2304 in.bam | \
  samtools collate -2 --output-fmt SAM -O - collate_tmp | \
    progNeedingSamPairs
```

It would be sensible to output r1/r2 ordered with this option

whitwham:
Adds the -f option to samtools collate to allow fast paired collate.  The option only outputs primary alignments with secondary and supplementary ones being lost.

Closes #781


bamshuf.c
```diff
@@ -153,19 +283,146 @@ static int bamshuf(const char *fn, int n_files, const char *pre, int clevel,
...
+                        goto fast_fail;
+                    }
+                } else { // we have a match
+                    // write out the reads in R1 R2 order
+                    bam1_t *r1, *r2;
+
+                    if (b->core.flag & BAM_FREAD1) {
+                        r1 = b;
+                        r2 = kh_value(stored, itr).bi->b;
+                    } else {
```
daviesrob: Note that both BAM_FREAD1 and BAM_FREAD2 can be set at the same time.

Test needs to be `b->core.flag & (BAM_FREAD1 | BAM_FREAD2) == BAM_FREAD1` and similar for read2.

whitwham: Maybe, but the code is there to order the writing of a pair in R1 R2 order.  If one or both have BAM_FREAD1 and BAM_FREAD2 set what does that do to the order?  Does it have any meaning any more?

jkbonfield: Temlate-internal reads that are neither the first nor the last get READ1+READ2 set.  It's not a common thing to see though.

If there's filtering done elsewhere and we know only read 1 *or* read 2 can get to this code then this change will be irrelevant.

PR #825, opened 2018-04-23
Remove redundant ks_mergesort() call in bam_sort.c

sashatoday:
There is already sorted block after condition:
```
// Sort last records
if (k > 0) {
        ...
        num_in_mem = sort_blocks(n_files, k, buf, prefix, header, n_threads, in_mem);
        ...
}
```
Thus, there is no need to sort single block again in:
```
if (n_files == 0 && num_in_mem < 2) { // a single block
        ks_mergesort(sort, k, buf, 0);
        ...
}
```


daviesrob:
Thanks.  This should have been removed as part of 4280d8fa - looks like I missed it.

PR #827, opened 2018-04-23
samtools view help message output

### LINKED ISSUE #824: Samtools sort: truncated file Aborting
[W::sam_read1] Parse error at line 2
samtools sort: truncated file. Aborting

I am getting above error while sorting the bam file.

valeriuo:
Prints the help message when no argument is provided.
Prints an error message when unable to parse the sampling argument.
Redirects the help message to `stderr` when no input file is detected.
Fixes #824.

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -279,6 +279,8 @@ int main_samview(int argc, char *argv[])
     /* parse command-line options */
     strcpy(out_mode, "w");
     strcpy(out_un_mode, "w");
+    if (argc == 1 && isatty(STDIN_FILENO))
+        return usage(stdout, EXIT_SUCCESS, is_long_help);
     while ((c = getopt_long(argc, argv,
                             "SbBcCt:h1Ho:O:q:f:F:G:ul:r:?T:R:L:s:@:m:x:U:M",
                             lopts, NULL)) >= 0) {
@@ -290,7 +292,18 @@ int main_samview(int argc, char *argv[])
                 srand(settings.subsam_seed);
                 settings.subsam_seed = rand();
             }
-            settings.subsam_frac = strtod(q, &q);
+ 
+            if (q && *q == '.') {
+                settings.subsam_frac = strtod(q, &q);
+                if (*q) ret = 1;
+            } else {
+                ret = 1;
+            }
+
+            if (ret == 1) {
+                print_error("view", "Incorrect sampling argument \"%s\"", optarg);
+                goto view_end;
+            }
             break;
         case 'm': settings.min_qlen = atoi(optarg); break;
         case 'c': is_count = 1; break;
@@ -367,7 +380,10 @@ int main_samview(int argc, char *argv[])
         strcat(out_mode, tmp);
         strcat(out_un_mode, tmp);
     }
-    if (argc == optind && isatty(STDIN_FILENO)) return usage(stdout, EXIT_SUCCESS, is_long_help); // potential memory leak...
+    if (argc == optind && isatty(STDIN_FILENO)) {
+        print_error("view", "No input provided or missing option argument.");
+        return usage(stderr, EXIT_FAILURE, is_long_help); // potential memory leak...
+    }
 
     fn_in = (optind < argc)? argv[optind] : "-";
     // generate the fn_list if necessary
--- a/sam_view.c
+++ b/sam_view.c
@@ -244,7 +244,6 @@ static void check_sam_close(const char *subcmd, samFile *fp, const char *fname, 
 int main_samview(int argc, char *argv[])
 {
     int c, is_header = 0, is_header_only = 0, ret = 0, compress_level = -1, is_count = 0;
-    int is_long_help = 0;
     int64_t count = 0;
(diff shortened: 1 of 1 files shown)


sam_view.c
```diff
@@ -291,6 +293,8 @@ int main_samview(int argc, char *argv[])
                 settings.subsam_seed = rand();
             }
             settings.subsam_frac = strtod(q, &q);
+            if (settings.subsam_frac < 1e-10)
+                print_error_errno("view", "Could not parse sampling argument \"%s\"", optarg);
```
jmarshall: The best way to see if `strtod` succeeded is to look at `q` and see if it consumed any characters.

sam_view.c
```diff
@@ -367,7 +371,7 @@ int main_samview(int argc, char *argv[])
         strcat(out_mode, tmp);
         strcat(out_un_mode, tmp);
     }
-    if (argc == optind && isatty(STDIN_FILENO)) return usage(stdout, EXIT_SUCCESS, is_long_help); // potential memory leak...
+    if (argc == optind && isatty(STDIN_FILENO)) return usage(stderr, EXIT_FAILURE, is_long_help); // potential memory leak...
```
jmarshall: Now that this is an error message, it makes no sense for it to depend on whether stdin is a tty or not.

Previously this line was there so that you could say `stuff | samtools view -o foo.bam` and it would copy stdin to _foo.bam_ but if you said `samtools view` you would get a usage display. If you really wanted to copy from stuff you typed on the terminal, you could say `samtools view -o foo.bam -`.

Now that the plain `samtools view` usage is handled separately, either this line should disappear or the `if` condition should be as is and the error message should say what's really going on:

> Refusing to sit here reading from standard input and looking like I'm in an infinite loop. If you really want to type my standard input on the terminal, add `-` as a filename

jmarshall: Good point about needing `isatty` in the new bit at the top: the cases that are being differentiated include `samtools view` to show the usage and `stuff-outputting-BAM | samtools view` to show SAM records.

It's unsurprising that samtools's argument handling contains error handling bugs: it is old code written by people who did not consider detecting user error to be a priority. For similar reasons, Unix filter conventions and `-` handling are uneven.

Samtools `view` has better argument handling than some of the other samtools commands and is especially heavily used in the middle of pipelines (compared to other subcommands). So IMHO avoiding breaking existing argument handling behaviours **should** hold up this PR :smile:

jkbonfield: Oh I agree that breaking things should hold it up!  We shouldn't go backwards.

I just meant that not fixing other existing bugs isn't necessarily a problem.  So for example `cat foo.bam | samtools view -r` doesn't fail.  It was true before, and it's still true now.  Ideally we'd fix all these in one step, but if split over subsequent PRs I could live with that too.

sam_view.c
```diff
@@ -290,7 +292,18 @@ int main_samview(int argc, char *argv[])
...
+ 
+            if (q && *q == '.') {
+                settings.subsam_frac = strtod(q, &q);
+                if (*q) ret = 1;
+            } else {
+                ret = 1;
+            }
+
+            if (ret == 1) {
+                print_error_errno("view", "Incorrect sampling argument \"%s\"", optarg);
```
jmarshall: errno will almost never mean anything here, so just use `print_error`. Also the `fprintf` added below would be better as `print_error` — prettifying these error messages is what the `print_error` family is for.

daviesrob:
I've had a go at fixing `-?` and unrecognised option handling.

Now we get:

`samtools view`  => short help to stdout; exit success
`samtools view -?` => long help to stdout; exit success
`samtools view -a` => short help to stderr; exit failure
`samtools view --foo` => short help to stderr; exit failure

Printing up an unrecognised long option turns out to be hard.  I went for the `optind - 1` option based on discussion [in this stack exchange answer](https://stackoverflow.com/questions/2723888/where-does-getopt-long-store-an-unrecognized-option).  It seems to be the most portable solution.

sam_view.c
```diff
@@ -345,13 +351,29 @@ int main_samview(int argc, char *argv[])
         //case 'x': out_format = "x"; break;
         //case 'X': out_format = "X"; break;
                  */
-        case '?': is_long_help = 1; break;
+        case '?':
+            if (optopt == '?') {  // '-?' appeared on command line
```
jkbonfield: This all feels excessively complicated and obfuscated for what is, when all said and done, a rather pointless feature that is probably better done by changing `-?` to `--help`.

If we really want to keep it as -? then maybe something far simpler (as suggested by the same stackoverflow answer) works:

```
-        case '?': is_long_help = 1; break;
+        case '?':
+            if (optind>0 && !strcmp(argv[optind-1],"-?"))
+                is_long_help = 1;
+            else
+                return usage(stderr, EXIT_FAILURE, 0);
+            break;
```

Yes it doesn't cope with bizarreness like `samtools view -h?`, but really I don't see this being something we should go out of our way to support.  My preference is still just drop the use of `-?` completely as it's just asking for confusion.

PR #829, opened 2018-04-24
disable underscore interpretation in title for gnuplot

zmiimz:
I use underscore in names of BAM stats  files  and taking those as title in plot-bamstats creates interpretation problem in gnuplot  (i.e. first symbol after underscore is interpreted as subscript). The  keyword "noenhanced"  in "gnuplot title" instruction changes this.

### DIFF
--- a/misc/plot-bamstats
+++ b/misc/plot-bamstats
@@ -682,7 +682,7 @@ sub plot_qualities
                 set yrange [0:$yrange]
                 unset ylabel
                 set xlabel "Cycle (rev reads)"
-                set label "$$args{title}" at screen 0.5,0.95 center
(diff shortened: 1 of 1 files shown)


daviesrob:
Thanks for this, although I'm having trouble reproducing the original problem as `plot-bamstats` doesn't appear to set enhanced mode on the terminal itself.  How are you getting into enhanced mode?

zmiimz:
See here https://sourceforge.net/p/gnuplot/bugs/1857/

daviesrob:
Ah, I see.  My gnuplot is too antiquated to have this bug.  In that case, I'll merge this in.

PR #832, opened 2018-04-26
Make samtools idxstats work on SAM and CRAM files.

### LINKED ISSUE #303: samtools idxstats produces incorrect values for cram
(copying here from samtools-help) I’ve come across one issue working with cram files in the samtools 1.x release, which is that idxstats will run on cram files, but doesn’t actually provide read counts.

This can be reproduced on master (b7c4b4d05154887ec8587165c6da505aad024f52) :

```
# from samtools directory:
$ samtools idxstats test/mpileup/ce#5b.bam
CHROMOSOME_I    15072423        1       0
CHROMOSOME_II   15279345        1       0
CHROMOSOME_III  13783700        1       0
CHROMOSOME_IV   17493793        1       0
CHROMOSOME_V    20924149        3       0
*       0       0       0
$ samtools idxstats test/mpileup/ce#5b.cram
CHROMOSOME_I    15072423        0       0
CHROMOSOME_II   15279345        0       0
CHROMOSOME_III  13783700        0       0
CHROMOSOME_IV   17493793        0       0
CHROMOSOME_V    20924149        0       0
*       0       0       139886584375224
```

jkbonfield:
This isn't fast, but it fixes #303 and gives consistency across all
formats.  Cram is around 5x faster than unindexed BAM which in turn is
maybe 2x faster than unindexed SAM, but even CRAM is slow compared to
the almost instant speed of indexed BAM.

A possible future extension would be fixing the CRAM index format by
an additional two extra columns, but this is a spec related problem
and we'd still need to support older files.

Note in looking at this I saw we had several *bugs* before, rather than simply omissions.  When given a cram it wasn't only producing misleading data with zeros, but it also read off the end of the memory structures when getting the unmapped sums.   Secondly the man page explicitly stated that it worked on both sam and cram, which was obviously not true.

PS.  I know - nasty use of goto, but it's self explanatory at least.


jkbonfield:
Ah, I was trying to work out how to determine when the index works, because the ideal scenario would be to load the index (bam, cram, etc) and then see if it has sufficient data to work with.  I tried hts_idx_get_meta and it didn't help, but I hadn't spotted hts_idx_get_stat.  That'll make the code path a bit less contorted. :-)

jmarshall:
For BAM, I think it's safe to assume that tools either don't write meta-bins at all or do write them for all chromosomes in the file.

Agree you shouldn't worry too much about ancient BAM indices on 100% unmapped data… in particular, because `hts_idx_load_core()` doesn't distinguish between “index doesn't have a n_no_coor field” and n_no_coor==0. Hmmm… Or maybe it's worth fixing that by setting that missing field in `hts_idx_load_core()` to MAX_UINT64 or so rather than 0, so that `hts_idx_get_n_no_coor()` could use that value to signal this field's absence precisely too…

jkbonfield:
I see `hts_idx_get_stat` fails on references that have no coverage too, so it's not possible to use it to distinguish between empty references and indices that don't have stats available.

I think therefore the current code is probably sufficient and the best we can manage without htslib changes too.

PR #835, opened 2018-04-27
Faster sorting algorithm by coordinates (radixsort)

sashatoday:
I made a try to improve the current sorting approach with implementation of the radix sort algorithm.

It turned out to be 3-6 times faster compared to the ks_mergesort() depending on data and hardware configuration (I used 2 test stands: a laptop with 4 cores and a cluster node with 16 cores).

I'm looking forward to comments/suggestions on the code.

It would be great if you could evaluate the effectiveness of the proposed solution with your data sets and environment.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1938,14 +1938,14 @@ static int write_buffer(const char *fn, const char *mode, size_t l, bam1_tag *bu
 }
 
 #define NUMBASE 256
-#define MAXDIGIT 64
 #define STEP 8
 
 void ks_radixsort(size_t n, bam1_tag *buf, uint64_t *pos_arr)
 {
     int i, curr = 0;
     bam1_tag *buf_ar2[2], *bam_a, *bam_b;
-    uint64_t *pos_ar2[2], *pos_a, *pos_b, shift = 0;
+    uint64_t *pos_ar2[2], *pos_a, *pos_b;
+    uint64_t max_pos = 0, max_digit = 0, shift = 0;
 
     buf_ar2[0] = buf;
     buf_ar2[1] = (bam1_tag *)malloc(sizeof(bam1_tag) * n);
@@ -1960,13 +1960,21 @@ void ks_radixsort(size_t n, bam1_tag *buf, uint64_t *pos_arr)
         goto err;
     }
 
-    while (shift < MAXDIGIT){
+    pos_a = pos_ar2[0];
+    for (i = 0; i < n; ++i)
+        if (pos_a[i] > max_pos) max_pos = pos_a[i];
+    while (max_pos) {
+        ++max_digit;
+        max_pos = max_pos >> 1;
+    }
+
+    while (shift < max_digit){
         int remainders[NUMBASE] = { 0 };
         pos_a = pos_ar2[curr]; pos_b = pos_ar2[1-curr];
         bam_a = buf_ar2[curr]; bam_b = buf_ar2[1-curr];
-        for (i = 0; i < n; i++)
+        for (i = 0; i < n; ++i)
             remainders[(pos_a[i] >> shift) % NUMBASE]++;
-        for (i = 1; i < NUMBASE; i++)
+        for (i = 1; i < NUMBASE; ++i)
             remainders[i] += remainders[i - 1];
         for (i = n - 1; i >= 0; i--) {
             int j = --remainders[(pos_a[i] >> shift) % NUMBASE];
@@ -1978,7 +1986,7 @@ void ks_radixsort(size_t n, bam1_tag *buf, uint64_t *pos_arr)
     if (curr == 1) {
 		bam1_tag *end = buf + n;
 		bam_a = buf_ar2[0]; bam_b = buf_ar2[1];
-		for (; bam_a < end; ++bam_b) *bam_a++ = *bam_b;
+		for (; bam_a < end;) *bam_a++ = *bam_b++;
 	}
 
 err:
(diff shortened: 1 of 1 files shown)


jkbonfield:
I can confirm the function itself is significantly faster (I saw 5 fold speed up), although the overall sort is mostly dominated by other things (temporary files, compressing / decompressing, etc) so user visible speed changes will be anywhere from nil to 30% or so.  (Switching to libdeflate will probably help too.)

Still, a gain is a gain so I'll review the code with a plan to merge. Thank you.

kyleabeauchamp:
Very cool!

jkbonfield:
I've been tweaking it a bit and have some observations.

1. Using NUMBASE 16 and STEP 4 appears to be around 30% faster.  This is likely either better cache usage or more likely simply that 16 is a small enough loop size for compilers to unroll and optimise.

2. The only thing that stops us from bailing out the radix sort early is that tid -1 (unmapped reads) causes our pos to be 0xffffffff00000000, so it needs to go the whole cycle through.  Hacking the code that produces this field to use tid == header->n_targets in place of -1 then permits the radixsort while loop to break part way through.  This reduces time further, so it's now over double the speed than before.

3. The malloc calls need checking for failure.

I'm happy to send you these changes for review, or if you're enjoying the exercise yourself then I'll let you make them!  Your call, and thanks for the PR.

jkbonfield:
Ugh yes it's very machine specific.  'm guessing this is some cache layout thing.

All the same binary and data (assuming a constant random() implementation).  Binaries built using gcc 8.1 with -O3.

```
Intel(R) Core(TM) i5-4570 CPU @ 3.20GHz
   16  100000000  8.87 s
  256  100000000  6.47 s
65536  100000000 16.70 s

AMD Opteron(tm) Processor 6378
   16  100000000 20.58 s
  256  100000000 14.61 s
65536  100000000 31.76 s

Intel(R) Xeon(R) CPU E5-2660 0 @ 2.20GHz
   16  100000000 14.11 s
  256  100000000 23.36 s
65536  100000000 21.83 s

Intel Core Processor (Broadwell) - VM
   16  100000000 11.50 s
  256  100000000 31.97 s
65536  100000000 20.68 s
```

Realistically though it doesn't matter hugely as all are significantly faster than the old code.  I'm just curious to see why it's behaving this way.

jkbonfield:
I rebased your branch and tweaked it a little.  Please let me know if you agree with these changes:

https://github.com/jkbonfield/samtools/tree/radixsort-jkb

- Replaced pos_arr with storing position inside the bam1_tag structure itself.  This means we only need one thing to swap during sorting and a lot of code vanishes.  (Although we get many minor changes elsewhere as we don't want to use anonymous unions.)

- White space fixes (tabs vs spaces).

- Better error checking; added a return value to radixsort.  Also made it static.

valeriuo:
Since it's now static, we should also get back to calling the function `radixsort`.

sashatoday:
Perfect solution! I agree with these changes.

I used an array for faster access to the pos. Also I was trying to vectorize the sort, but there was a slight increase in speed, which is not really significant.
Anyway I'm completely agree with this modification. Using a union looks clearer in the code than a separate array.

Also I would propose to replace in your branch:
```c
if (g_is_by_tag) {
    buf[k].u.tag = bam_aux_get(buf[k].bam_record, g_sort_tag);
} else if (g_is_by_qname) {
    buf[k].u.tag = NULL;
 } else {
    buf[k].u.tag = NULL;
}
```
with this:
```c
if (g_is_by_tag) {
    buf[k].u.tag = bam_aux_get(buf[k].bam_record, g_sort_tag);
 } else {
    buf[k].u.tag = NULL;
}
```
as it was before PR.

What about NUMBASE, I don't even know which one would be better in general case.
I leave it up to you.

jkbonfield:
Good point about that conditional.

I think in some cases the pos array did get vectorised - I tried various compilers and options and it wasn't universal but it did get vectorised in some cases.  However the speed gain wasn't great really as it's mainly memory bandwidth limited and the number of instructions per cycle is low.

That's also the big difference between NUMBASE 16 and 256.  *When* 16 works it does so by hugely increasing the number of instructions per cycle, as 16 needs more sort steps than 256, but it varied across platforms,  Maybe it's possible to align things just so and get the NUMBASE 16 working well everywhere, but it sounds fragile so I think 256 is OK.  To be honest it's over optimising it anyway - this is no longer remotely the bottleneck!

Thanks for your work on this.

sashatoday:
I was glad to make useful changes. Thank you for your assistance.

jkbonfield:
I tried renaming it to radixsort again, thinking it would be OK, but the problem is with MacOS X stdlib.h which includes a function named radixsort and exposed without guards.  The man page claims it's part of 4.4BSD, but I haven't seen this available on other platforms. 

@sashatoday do you mind if I just squash my changes into your commit (with you as the author) during merge?

jmarshall:
Confirmed that FreeBSD and OpenBSD also explode when presented with

    static void radixsort(size_t n, bam1_tag *buf, unit64_t *pos_arr)

as it conflicts with the `radixsort()` declared in their _stdlib.h_ (which of course _bam_sort.c_ includes). 4.4BSD release notes say

> The  `qsort` function has  been  rewritten  for  additional  performance.   In addition,   three  new  types  of  sorting  functions,  `heapsort`, `mergesort`, and `radixsort` have been  added  to  the  system.

so apparently a BSDism that didn't make it into POSIX.

I suppose you could consider using the system `radixsort` where available, but more than likely more trouble than it's worth.

jkbonfield:
The system defined one doesn't look appropriate for our usage anyway, given it's nominally for variable length strings instead of numbers.

sashatoday:
No, I don't mind. Thanks!

> @sashatoday do you mind if I just squash my changes into your commit (with you as the author) during merge?

kyleabeauchamp:
Thanks both!  The sort of enhancement will be helpful to a lot of folks.

PR #840, opened 2018-05-01
Added option to read regions from a text file.

### LINKED ISSUE #753: faidx - argument list is too long.

whitwham:
This commit also fixes a bug in writing output to file.  The tests have been updated accordingly. Closes #753 and is an alternative solution to the one provided in #754.


faidx.c
```diff
@@ -42,18 +42,76 @@ DEALINGS IN THE SOFTWARE.
...
     fprintf(fp, "Usage: samtools faidx <file.fa|file.fa.gz> [<reg> [...]]\n");
     fprintf(fp, "Option: \n"
-                " -o, --output   FILE Write FASTA to file.\n"
-                " -n, --length   INT  Length of FASTA sequence line. [60]\n"
-                " -c, --continue      Continue after trying to retrieve missing region.\n"
-                " -h, --help          This message.\n");
+                " -o, --output      FILE Write FASTA to file.\n"
+                " -n, --length      INT  Length of FASTA sequence line. [60]\n"
+                " -c, --continue         Continue after trying to retrieve missing region.\n"
+                " -r, --region_file FILE File of regions.  One per line.\n"
```
valeriuo: It should be stated that a line in the region file has the format **reg:from-to** (`chr1:12209228-12209246`) and it is **1-based**, so to be clear that is not a BED file or any other TAB delimited format.

jmarshall: Is there a reason it couldn't additionally accept the regions as a BED file? This would be useful and in line with other samtools commands (though starting to duplicate the already-existing `bedtools getfasta`).

While you're editing the usage… wouldn't putting FILE/INT/etc beside their options be more readable?

    " -o, --output FILE       Write FASTA to file\n"
    …etc…
    " -r, --region_file FILE  File of regions.  One per line\n"

jmarshall: Oh, also usual would be `--region-file` with hyphen rather than underscore.

whitwham: Regions have their own section in the man page and neither the current faidx usage or man entry defines them.  I will add a little bit more to the usage but not a full definition.  BED files are a different issue and if anyone wants to raise one then I will look at it.  Yes it should be --region-file.

PR #844, opened 2018-05-11
Fix --version when building from a Git repository

jmarshall:
Samtools's version infrastructure has been broken since January:

```
$ make
make: Nothing to be done for `all'.
$ git pull
Already up to date.
$ git describe --dirty --always
1.8-20-g4ff8062                           <--
$ make
make: Nothing to be done for `all'.
$ ./samtools --version
samtools 1.7-5-g2bd9d6b-dirty             <--
Using htslib 1.8-13-g19a66ce-dirty
```

This is caused by 3ecdcc4760ae76bbdedaa40ae0d4c0bff26a5b20, which changed _bam_sort.c_ and _bamshuf.c_ to use HTSlib's `sam_hdr_change_HD()`, but also included some unrelated code that appears to be local test code and/or incomplete.

Reverting those changes — which are not mentioned in the commit's log message — fixes the versioning and makes `samtools --version` report the right thing when built from a busy samtools Git checkout.

### DIFF
--- a/bamtk.c
+++ b/bamtk.c
@@ -31,6 +31,7 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include "htslib/hts.h"
 #include "samtools.h"
+#include "version.h"
 
 int bam_taf2baf(int argc, char *argv[]);
 int bam_mpileup(int argc, char *argv[]);
@@ -63,6 +64,10 @@ int main_addreplacerg(int argc, char *argv[]);
 int faidx_main(int argc, char *argv[]);
 int dict_main(int argc, char *argv[]);
 
+const char *samtools_version()
+{
+    return SAMTOOLS_VERSION;
+}
 
 static void usage(FILE *fp)
 {
--- a/sam_utils.c
+++ b/sam_utils.c
@@ -28,10 +28,8 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdarg.h>
 #include <string.h>
 #include <errno.h>
-#include <stdlib.h>
 
 #include "samtools.h"
-#include "version.h"
 
 static void vprint_error_core(const char *subcommand, const char *format, va_list args, const char *extra)
 {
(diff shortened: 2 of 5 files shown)


jkbonfield:
This looks like migration of code from bamtk.c to sam_utils.c, but it's the same code.  I cannot explain why it would give the wrong information in one file but not the other.  What am I missing here?

Also, when I rebuilt from a clean checkout it does indeed work fine and gives the correct version.  Is this something to do with git pull between releases and something not getting rebuilt?  Maybe due to pull not fetching tags by default?

jmarshall:
Don't do it from a clean checkout, do it from a busy working checkout. What you're missing is the incorrect dependencies, whereby after a `git pull` or similar, `version.sh` is rerun and _version.h_ is rebuilt, but the results do not make it into the `samtools` executable.

As the unrelated additions are variously wrong (`if (verbose > 0)` in _test_filter_header_rg.c_) or still unused and arguably misguided (`samtools_version_short()`), it seems best to revert the lot and individual changes can come back again in a corrected form if they are thought desirable.

jkbonfield:
OK, but I think the migration out of bamtk.c into a common file is beneficial as it can then be used elsewhere.  Eg it's minimal, but the test harness now gets access to version information instead of only the samtools binary itself.

It seems like the correct solution would be to fix the dependencies in the Makefile.

jmarshall:
As the unrelated additions are variously wrong (if (verbose > 0) in test_filter_header_rg.c) or still unused and arguably misguided (samtools_version_short()), it seems best to revert the lot and individual changes can come back again in a corrected form if they are thought desirable.

> the migration out of bamtk.c into a common file is beneficial

That migration is also incomplete. The dependencies would need to be fixed, and items such as _misc/wgsim.c_ would want to use it instead of `SAMTOOLS_VERSION` directly. So it seems like the correct solution would be to revert it and bring it back in corrected form (with a dependency update and other things using it) in a commit that mentions that the version infrastructure is being updated.

(Why does the test harness need up-to-date version information? The additional changes also show that that just makes writing the expected output harder…)

jmarshall:
Thanks Rob!

PR #854, opened 2018-05-24
make test on Windows (MINGW)

valeriuo:
Fixes Windows test run, as some files generated on Windows have CR/LF line endings.

### DIFF
--- a/test/test.pl
+++ b/test/test.pl
@@ -158,6 +158,7 @@ sub parse_params
     {
         $SIG{TERM} = $SIG{INT} = sub { clean_files($opts); };
     }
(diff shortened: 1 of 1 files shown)


jmarshall:
Fortunately this is a Perl script rather than a shell script. So you could only add `--strip-trailing-cr` (or `-b`) when `$^O` indicates that it's on some flavour of Windows. There's already quite a lot of similar checking of `$^O` in this script…

(Might be best to add a suitable `$$opts{diff}` or similar in `parse_params()`, setting it to `"diff"` or `"diff --strip-trailing-cr"` as appropriate.)

PR #859, opened 2018-05-31
Mpileup depth limit settings

### LINKED ISSUE #851: Mpileup --max-depth
Dear! 
I know that mpileup has a parameter for avoiding excessive memory usage. It takes this value as 8000 by default but I would like to reduce this value up to 1000. 

I was trying to do it with:
samtools mpileup -d 1000 ...
but in the output there were still positions with coverage much higher than 1000. 

I was wondering what happened as I was expecting samtools mpileup kind of sampling reads per position with a maximum of 1000, but it is not what the tool is doing. 

Did I missunderstand the procedure of this parameter? If not, how can I reduce the memory usage for samtools mpileup?

Thanks a lot

### LINKED ISSUE #858: mpileup without max-depth limitation
Is it possible to enable running samtools mpileup without any max-depth?
We need such a feature to calculate exact mismatch rates.

Thanks!

valeriuo:
Add the possibility of increasing the depth limit to the maximum integer value
by passing the **-d** option with the value 0.
Reduce the depth lower limit to the default (250).
Fixes #858 
Fixes #851

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -115,7 +115,8 @@ static inline void pileup_seq(FILE *fp, const bam_pileup1_t *p, int pos, int ref
 #define MPLP_SMART_OVERLAPS (1<<12)
 #define MPLP_PRINT_QNAME (1<<13)
 
-#define MPLP_MAX_DEPTH 250
+#define MPLP_MAX_DEPTH 8000
+#define MPLP_MAX_INDEL_DEPTH 250
 
 void *bed_read(const char *fn);
 void bed_destroy(void *_h);
@@ -422,6 +423,7 @@ static int mpileup(mplp_conf_t *conf, int n, char **fn)
     if (conf->flag & MPLP_BCF)
     {
         const char *mode;
+
         if ( conf->flag & MPLP_VCF )
             mode = (conf->flag&MPLP_NO_COMP)? "wu" : "wz";   // uncompressed VCF or compressed VCF
         else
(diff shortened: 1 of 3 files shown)


bam_plcmd.c
```diff
@@ -554,12 +556,16 @@ static int mpileup(mplp_conf_t *conf, int n, char **fn)
...
-    if (max_depth * sm->n < 8000) {
-        max_depth = 8000 / sm->n;
-        fprintf(stderr, "<%s> Set max per-file depth to %d\n", __func__, max_depth);
+    if ( !conf->max_depth ) {
+        max_depth = INT_MAX / sm->n;
+    } else {
+        max_depth = conf->max_depth;
+        if ( max_depth * sm->n > 1<<20 )
+            fprintf(stderr, "(%s) Max depth is above 1M. Potential memory hog!\n", __func__);
+        if ( max_depth < MPLP_MIN_DEPTH ) {
```
jkbonfield: This isn't quite the same as before.

If we have 1000 samples and asked for max depth of 10 then previously that's what we'd have got as 10 * 1000 >= 8000. If we'd have set it to 2 then it'd have bumped it to 8 for us.  This is a bit strange as setting it to 8000 say (the default) actually means 8000 * 1000 total aggregate depth.  So sometimes the code is per sample and sometimes it's total.  I think this is likely a mistake.

Now we're doing the more logical thing which is a per sample limit that applies equally regardless of whether we have 1 or 1000 samples; it does what the user expects.  However if they have 10,000 samples they may well wish to set max_depth (per sample) to just 50 say.  I think what we have is correct, the new minimum value is perhaps too high.

I can't really judge what is appropriate though.  @pd3 care to comment?

pd3: In bcftools the original code was changed to this https://github.com/samtools/bcftools/blob/develop/mpileup.c#L607-L610, so I am assuming this was inspired by it.

Note that in bcftools this is only a warning. The value 250 is the number of reads that contribute to the calculation of genotype likelihoods, but the code there does not enforce the limit in case the user wants to override it.

PR #877, opened 2018-06-21
Improve error reporting in bed_read()

### LINKED ISSUE #863: samtools depth does not handle a bed file
(run `samtools --version`)

Yes - using latest full release, Samtools 1.8. I first noticed the bug on Samtools 1.6, but upgrading to 1.8 did not resolve the issue

Linux 3.10.0-327.el7.x86_64 on a high-performance computing cluster

I can not get samtools depth to recognize a bed file for where to count depth positions. The command I'm using is

`samtools depth some.bam -b some.bed`

I've tried putting the bed file in quotes (` "some.bed" `), and various ways of formatting the bed file (added/removed header, added/removed comment lines). But every time, I get the error:

```
[bed_read] Parse error reading some.bed at line 1
samtools depth: Could not read file "some.bed": No such file or directory
```

It seems like this error would be on my side, but I can't figure out what's wrong. I'm able to supply regions like `-r "Chr1:456-789"`. I'm just having issues with using a bed file.

daviesrob:
Prints a more informative message when region end is less than start.

Sets errno to zero on parse errors so that callers using print_error_errno() to report failure don't print a misleading reason for being unable to read the file.

Fixes #863 (or at least makes it easier to see what the problem is).


bedidx.c
```diff
@@ -286,8 +286,17 @@ void *bed_read(const char *fn)
...
-            fprintf(stderr, "[bed_read] Parse error reading %s at line %u\n",
-                    fn, line);
+            if (num < 1) {
+                fprintf(stderr,
+                        "[bed_read] Parse error reading \"%s\" at line %u\n",
+                        fn, line);
+            } else {
+                fprintf(stderr,
+                        "[bed_read] Parse error reading \"%s\" at line %u : "
+                        "end (%u) may not be less that start (%u)\n",
```
jmarshall: _than_. And on second thoughts I'm not sure _may_ was a great choice…

bedidx.c
```diff
@@ -324,12 +333,16 @@ void *bed_read(const char *fn)
     //bed_unify(h);
     return h;
  fail:
+    save_errno = errno;
     fprintf(stderr, "[bed_read] Error reading %s : %s\n", fn, strerror(errno));
```
jmarshall: All existing callers of `bed_read()` follow its failure with a call to `print_error_errno()`. So this `fprintf` is redundant — so `fail_no_msg` can be folded into `fail`. With less need for errno juggling :smile:

jmarshall:
Thanks. Still a typo in “not be less **than** start”.

daviesrob:
I must be having a bad typing day.  We're getting to squash territory here, I think.

PR #878, opened 2018-06-25
Add a reverse complement option to faidx.

### LINKED ISSUE #869: Add a new parameter for faidx to reverse-complement the extracted FASTA sequences
Currently `samtools faidx` can extract FASTA sequences in a specific region, but can not reverse-complement them on the fly. How about add a new parameter  (e.g. -r) to reverse-complement them? If so, it would be much more easier to extract coding sequences if they were on the negative strand on genomic DNA

whitwham:
Allows faidx and fqidx to reverse complement extracted sequence.
Closes #869.

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -67,7 +67,6 @@ static unsigned char comp_base[256] = {
 240, 241, 242, 243, 244, 245, 246, 247, 248, 249, 250, 251, 252, 253, 254, 255,
 };
 
-
 static void reverse_complement(char *str, int len) {
     char c;
     int i = 0, j = len - 1;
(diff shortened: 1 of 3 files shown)


daviesrob:
Annotating the names is useful, but we might want to make it optional.  The [bedtools getfasta](https://bedtools.readthedocs.io/en/latest/content/tools/getfasta.html) command uses `(+)` and `(-)`.  [seqkit seq](https://bioinf.shenwei.me/seqkit/usage/#seq) leaves it unchanged.

Selecting the strand in the region list might have to be left until later, given the nightmare that is the region parser...

daviesrob:
I've added an option to append strand information to the sequence name.  Might have over-done it a bit, but it's quite flexible...

jmarshall:
I see you've merged this over-engineered :smile: approach now, but just to be clear: writing out “>1:50000-50100” without annotation while revcomping the sequence means that samtools is misleading the user about what the sequence is. (In the short term, the user may remember using the **-i** option, but the sensible policy for long-term comprehensibility is for output files to be self-describing.) 

I don't think leaving the default as “no indication of revcomping” is a good choice.

daviesrob:
Agreed, it would be better to change the default.

I'm planning to go with making the `/rc` option the default.  It keeps forward strand reads unchanged, and if we ever allow reverse complementing individual regions then it will avoid a name clash if someone wants to output the same region on both strands.

https://github.com/daviesrob/samtools/tree/mark-strand implements this.  Look OK to you?

jmarshall:
:+1: Thanks Rob.

PR #884, opened 2018-06-27
Marked BCF/VCF mpileup as deprecated.

jkbonfield:
The tests that use these features still exist, but this is deliberate.

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -919,44 +919,26 @@ static void print_usage(FILE *fp, const mplp_conf_t *mplp)
 "\n"
 "Output options:\n"
 "  -o, --output FILE       write output to FILE [standard output]\n"
-"  -g, --BCF               generate genotype likelihoods in BCF format\n"
-"  -v, --VCF               generate genotype likelihoods in VCF format\n"
-"\n"
-"Output options for mpileup format (without -g/-v):\n"
 "  -O, --output-BP         output base positions on reads\n"
 "  -s, --output-MQ         output mapping quality\n"
 "      --output-QNAME      output read names\n"
 "  -a                      output all positions (including zero depth)\n"
 "  -a -a (or -aa)          output absolutely all positions, including unused ref. sequences\n"
 "\n"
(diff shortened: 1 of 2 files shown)


bam_plcmd.c
```diff
@@ -919,46 +919,24 @@ static void print_usage(FILE *fp, const mplp_conf_t *mplp)
...
-"\n"
-"Notes: Assuming diploid individuals.\n");
 
     free(tmp_require);
     free(tmp_filter);
 }
 
+static void deprecated(char opt) {
+    fprintf(stderr, "[warning] samtools mpileup option `%c` is functional, "
+            "but deprecated. Please switch to using bcftools in future.\n", opt);
```
daviesrob: Possibly we should say "bcftools mpileup" instead of just "bcftools"?

samtools.1
```diff
@@ -1351,38 +1357,10 @@ Disable read-pair overlap detection.
 .B Output Options:
 .TP 10
 .BI "-o, --output " FILE
-Write pileup or VCF/BCF output to
+Write pileup output to
 .IR FILE ,
 rather than the default of standard output.
-
-(The same short option is used for both
```
daviesrob: I fear we still need to mention this horror somehow, if only to say that you need to use `-o ./123` instead of `-o 123` due to historic overloading of the `-o` option.

samtools.1
```diff
@@ -1225,10 +1225,16 @@ samtools mpileup
...
 .RI [ ... ]]
 
-Generate VCF, BCF or pileup for one or multiple BAM files. Alignment records
+Generate pileup for one or multiple BAM files. Alignment records
 are grouped by sample (SM) identifiers in @RG header lines. If sample
 identifiers are absent, each input file is regarded as one sample.
 
+Mpileup can still produce VCF and BCF output, but this feature is
+deprecated and will be removed in a future release.  Please use
+Bcftools for this feature.  (Documentation on the deprecated options
```
daviesrob: "bcftools mpileup" again.

PR #886, opened 2018-07-05
Add missing Makefile dependencies

jmarshall:
Update dependencies, similar to samtools/htslib#727.

In a couple of cases, removes spurious inclusions from the .c file rather than adding them to the Makefile. The _bam_lpileup.c_ one came from a MinGW-support-PR (see 447831d10a0c1a5b32f48cbf497211ff8fc0a5ee) and its removal will soon be tested in that environment in particular via appveyor.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT/Expat License
 
-Copyright (C) 2008-2014 Genome Research Ltd.
+Copyright (C) 2008-2018 Genome Research Ltd.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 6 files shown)


daviesrob:
Thanks.

On a related note, I wonder if we should be using `#include "config.h"` instead of `#include <config.h>` to avoid another `C_INCLUDE_PATH` related problem ... ?

jmarshall:
The [autoconf manual](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Configuration-Headers.html) specifically recommends `<config.h>`. Admittedly that's for the benefit of separate build-dir builds, which need a little work (and probably need the don't-use-configure mode of operation to disappear first).

Distinctive filenames (_htslib-version.h_ vs _samtools-version.h_) might have been a good way to fix that `C_INCLUDE_PATH` snafu. Similarly _htslib-config.h_ vs _samtools-config.h_ perhaps…

daviesrob:
Hmm, yes.  I guess I'll leave it as `<config.h>` for now then and revisit later.

PR #887, opened 2018-07-05
Improve fasta/q documentation and fix flushing bug

### LINKED ISSUE #885: samtools fasta (samtools fastq) does not extract paired reads
(run `samtools --version`)

samtools 1.7
Using htslib 1.7

- Linux 4.4.0-128-generic
- x86_64
- gcc (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609

I'm using the bioconda package.

Using the test data provided in the samtools sources

```
samtools fasta samtools_fastx-in1.bam -0 /tmp/pairs
```

should write (according to the help) all paired reads to  /tmp/pairs. But it does not. Also name sorting the file seems not to help. 

See also https://github.com/samtools/samtools/issues/634

Possible solutions: 
- improve documentation and specify what `-0` actually does 
- fix the bug

daviesrob:
Try to make what the fasta and fastq output options do clearer in both the online help and the manual page.

Fix a bug where stdout was not flushed when the -0 output option was used (found when testing out the examples).

Fixes #885 (samtools fasta (samtools fastq) does not extract paired reads)

### DIFF
--- a/samtools.1
+++ b/samtools.1
@@ -793,7 +793,7 @@ samtools bedcov
 .IR region.bed " " in1.sam | in1.bam | in1.cram "[...]"
 
 Reports the total read base count (i.e. the sum of per base read depths)
-for each genomic region specified in the supplied BED file. The regions 
+for each genomic region specified in the supplied BED file. The regions
 are output as they appear in the BED file and are 0-based.
 Counts for each alignment file supplied are reported in separate columns.
 
@@ -1841,7 +1841,7 @@ samtools cat [-b list] [-h header.sam] [-o out.bam] <in1.bam> <in2.bam> [ ... ]
 Concatenate BAMs or CRAMs. Although this works on either BAM or CRAM,
 all input files must be the same format as each other. The sequence
 dictionary of each input file must be identical, although this command
-does not check this. This command uses a similar trick to 
+does not check this. This command uses a similar trick to
 .B reheader
 which enables fast BAM concatenation.
 
--- a/samtools.1
+++ b/samtools.1
@@ -1516,15 +1516,69 @@ samtools fasta
 .I in.bam
 
 Converts a BAM or CRAM into either FASTQ or FASTA format depending on the
-command invoked. The FASTQ files will be automatically compressed if the 
-filenames have a .gz or .bgzf extention.
+command invoked. The files will be automatically compressed if the
+file names have a .gz or .bgzf extension.
+
+The input to this program must be collated by name.
+Use
+.B samtools collate
+or
+.B samtools sort -n
+to ensure this.
+
+For each different QNAME, the input records are categorised according to
+the state of the READ1 and READ2 flag bits.
+The three categories used are:
+
+1 : Only READ1 is set.
+
+2 : Only READ2 is set.
+
+0 : Either both READ1 and READ2 are set; or neither is set.
+
+The exact meaning of these categories depends on the sequencing technology
+used.
+It is expected that ordinary single and paired-end sequencing reads will be
+in categories 1 and 2 (in the case of paired-end reads, one read of the pair
+will be in category 1, the other in category 2).
+Category 0 is essentially a \*(lqcatch-all\*(rq for reads that do not
+fit into a simple paired-end sequencing model.
+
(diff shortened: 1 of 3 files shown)


samtools.1
```diff
@@ -1517,14 +1517,40 @@ samtools fasta
...
+The three categories used are:
+
+1 : Only READ1 is set.
+
+2 : Only READ2 is set.
+
+0 : Either both READ1 and READ2 are set; or neither is set.
+
+For each category only one sequence will be written for a given QNAME.
+Records which include quality values will be preferred over ones that do not,
```
bernt-matthias: This is a bit unclear. What happens if both have quality values? How exactly are the quality values compared?

samtools.1
```diff
@@ -1517,14 +1517,40 @@ samtools fasta
 
 Converts a BAM or CRAM into either FASTQ or FASTA format depending on the
 command invoked. The FASTQ files will be automatically compressed if the 
-filenames have a .gz or .bgzf extention.
+file names have a .gz or .bgzf extension.
```
bernt-matthias: I'm missing here the distinction of the cases if -s is set or not.

samtools.1
```diff
@@ -1615,6 +1650,53 @@ mean 'read until the separator or end of tag', for example:
 .B n*i*
 ignore the left part of the tag until the separator, then use the second part
 .RE
+
```
bernt-matthias: I really like the examples :)

jmarshall:
Nice catch on the `hstdout` never gets closed if you use `-0` thing.

Even with your patch, if you use all of `-0`, `-1`, `-2` it still never gets closed so is leaked (and still isn't flushed, but maybe with all those options there's sure to be nothing in it). Maybe clearer would be just the simple thing (with all its error handling back in of course…):

```
for (i = 0; i < 3; ++i)
    if (state->fpr[i] != state->hstdout)
        bgzf_close(state->fpr[i]);

bgzf_close(state->hstdout);
```

daviesrob:
@jmarshall `state->hstdout` is only used by the `-0`, `-1` and `-2` outputs, but the suggested version is a bit simpler.

I've also made it only open `state->hstdout` if it's actually needed, and as a bonus added a check that it actually worked...

samtools.1
```diff
@@ -1536,15 +1536,43 @@ The three categories used are:
 
 0 : Either both READ1 and READ2 are set; or neither is set.
 
+The exact meaning of these categories depends on the sequencing technology
+used.
```
bernt-matthias: I guess it depends also on the mapper (in particular the cases that are caught with -0)

bernt-matthias:
Looks good to me (just one minor comment to the lates doc related commit). Great improvement. Thanks.

samtools.1
daviesrob: This text actually came from the [SAM](http://samtools.github.io/hts-specs/SAMv1.pdf) specification.  These bits are supposed to show the relative locations of the reads on the piece of DNA that was sequenced.  They're set by the sequencing protocol in use, and are an input to the mapper rather than an output (or should be for mappers that understand SAM/BAM/CRAM files as input).

bernt-matthias: OK. Thanks for the info. Didn't know that. Since I tell most mappers which files contain read1 and a which contains read2 I thought the mapper generates the info.

yhoogstrate:
Hi all (pinging to you because commit has been merged @daviesrob, @bernt-matthias, @jmarshall, @jkbonfield),

Thanks for this useful discussion and provided example.
As written in the commited updated docs, for paired-end output collate/name-sorting is needed prior to running `samtools fastq` to avoid that the read order gets out of sync. (If I understood it correctly)
I suspect there aren't that many use cases where collate is not needed, so would it make sense to change the paired-end example from:

```
samtools fastq -1 paired1.fq -2 paired2.fq -0 /dev/null -s /dev/null -n -F 0x900 in.bam
```

into

```
samtools collate -O in.bam | samtools fastq -1 paired1.fq -2 paired2.fq -0 /dev/null -s /dev/null -n -F 0x900
```

Kind regards,

Youri

PR #892, opened 2018-07-11
Memory leaks & buffer 1-beyond-end reads in corner cases.

jkbonfield:
This enables the test harness to pass when we use `-fsanitize=address` (which also does basic memory leak checking).  In turn, this permits us to enable it in travis, thus spotting more problems in pull requests.

I also changed it to use `./configure` too as without it it was building htslib without sanitise and samtools with it.  (This is probably a bug in our Makefile.)  We should consider an equivalent htslib travis mod too.

Note the sudo: required bit is because clang uses a second monitoring process and this cannot work when run from a container.  We only enable sudo for the one build though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,11 @@ language: c
 
 matrix:
   include:
+    # An optimised build with address, leak and undefined behavior checking
     - os: linux
       compiler: clang
+      sudo: required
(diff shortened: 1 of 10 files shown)


daviesrob:
Merged in cc47dd0 (to fix the buffer overrun for the 1.9 release), 284feed and 7a5f779.

Issue #104, opened 2013-12-06
Sub man pages for subcommands


jmarshall:
It might be nice if typing `man samtools view` gave you a man page just for the view subcommand.  Due to a new feature in man-db 2.5.6,

> By default, man will now try to interpret pairs of manual page names given on the command line as equivalent to a single manual page name containing a hyphen (e.g. 'man foo bar' => foo-bar(1)). This supports the common pattern of programs that implement a number of subcommands, allowing them to provide manual pages for each that can be accessed using similar syntax as would be used to invoke the subcommands themselves. Suggested by H. Peter Anvin, Federico Lucifredi, and others on the git mailing list.

We can arrange this by installing files named _samtools-view.1_ and so on.  We might possibly even be able to arrange it so that the _samtools.1_ file has introductory paragraphs and then just `.so` includes all the subcommand man pages.

jmarshall:
I have a partial proof-of-concept of this, and early in a release cycle would be the time to do it.  It will also facilitate making the [man page on the web](http://www.htslib.org/doc/samtools.html) more navigable.

PR #894, opened 2018-07-13
Split manpages

jkbonfield:
It's a bit of a work in progress and we need some discussion on how much to leave in the `samtools.1` man page.

However each program specific component from the original man page is now it's own manual named e.g. `samtools-merge.1`.  This means with a modern man command we can do e.g. `man samtools merge` (or samtools-merge otherwise).  It greatly helps when searching for options specific to a single sub command.

How to update the htslib.org man pages is left as an exercise to the PR merger. ;-)


jmarshall:
There was a feature request about this, #104.

The proof-of-concept at the time used `.so` to include the other pages in _samtools.1_ (and troff conditionals to avoid the sub-pages' preamble, headings, etc), so that the full _samtools.1_ man page would be similar to the pre-split one. That may or may not be worthwhile…

jkbonfield:
Thanks, I'd forgotten to check if we had this as an issue already.

I did think about using .so (it's only used in one place for fasta/fastq), but decided that the main samtools.1 page is best off not being an amalgamation of all separate commands, but something more useful.

jmarshall:
At the time, the man page had less detailed blurbs about somewhat fewer subcommands and no global options or environment variables to describe. So maintaining the amalgamation _samtools.1_ was semi-worthwhile back then, but agreed that there's more useful content for it now.

jkbonfield:
Things to-do still:

- Document `samtools stats`.  At the very least it needs a list of the file components it outputs and what they mean.  We can cull this from the comments in the output file.

- Consider reordering samtools.1 man page, either alphabetical (preferred) or into related groups.

- Fix the synopsis section of samtools.1 so it contains all the subcommands rather than an (arbitrary?) subset of them.

- Figure out the authors of the man page license section (unseen by users).  A lot of work and maybe I'll just gloss over this one.

- Check man page options to see if they match the actual options being used.  (I already found and fixed one example of this.)

doc/samtools-addreplacerg.1
```diff
@@ -0,0 +1,98 @@
...
+.\" IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+.\" FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+.\" THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+.\" LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+.\" FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+.\" DEALINGS IN THE SOFTWARE.
+.
+.\" For code blocks and examples (cf groff's Ultrix-specific man macros)
+.
+.\" For code blocks and examples (cf groff's Ultrix-specific man macros)
```
jmarshall: These two lines (blank and _For code blocks and examples_) have gotten duplicated in all the split files.

jkbonfield: Thanks.  I'll fix this.  They were all derived from the same template. :-)

jmarshall:
Listing authorship on the individual pages emphasises those who happen to have invented a new subcommand and does not credit some like Rob who have contributed significantly but more generally.

I would suggest dropping the AUTHOR sections from the individual pages and rewriting the umbrella _samtools.1_ page's AUTHOR section to credit everyone who has contributed significantly to samtools (and maybe htslib too, though it's probably the same list of people).

jkbonfield:
For sure the author section of the general man page needs expanding to encompass everyone.  I haven't changed that yet.  However I disagree that the authors for the individual sections aren't warranted.

I think at this point given htslib has no documentation at all, we would include htslib modifications in the samtools.1 man page as far as attribution goes.  If we ever get around to documenting the library then we can revisit this, but I wasn't attempting to solve every ailment in the first PR.

jkbonfield:
Just to add one more thing, I view this as one of many incremental improvements.  We shouldn't hold it up because it's not *perfect*.  What it does now is by and large one thing - splitting up the components.  Further fixes will be in further commits and PRs.

A case in point is the mess that is "samtools stats" with almost nil documentation.  We could start working on this, but it's best to get this PR done first and then start working on the samtools-stats.1 man page directly, as otherwise we just give ourselves more work to do in conflict resolutions.

daviesrob:
I've rebased this to current develop, including updates for the last release and changes to the man page that have appeared since then.  Sorry about the loss of history, but squashing it all before rebasing made the process much easier.

It still needs an update to the AUTHORs, which has so far defeated me.  The list of contributors is quite long - maybe we should list them all in a file somewhere instead (like, for example, [gcc](https://gcc.gnu.org/onlinedocs/gcc/Contributors.html) does)?

PR #898, opened 2018-07-18
Clarify decode_md meaning

EvanTheB:
I am confused about the wording of decode_md. I believe the default is to generated NM, MD iff the values are not in the file. decode_md=0 therefore makes view echo whatever is in the file without generation. It is less of a 'decode' and more of a 'generate if missing' flag. But maybe I am wrong.


jkbonfield:
Yes it is a bit poorly worded.  I agree with the changes except for the removal of "It can be particularly useful when combined with a file encoded using store_md=1 and store_nm=1", which I think is a useful hint as it is a common problem with round-tripping CRAM.

PR #902, opened 2018-07-24
Add checks on return value of bam_hdr_write, bam_write1

daviesrob:
Should have been there all along, but also fixes warnings that appear as a result of samtools/htslib#743 which added an HTS_RESULT_USED to these functions.


jmarshall:
`print_error_errno()`.

jmarshall:
Also “Failed to write **BCF** header” will be a little confusing if I'm writing a VCF file.

Issue #861, opened 2018-06-01
Fragment qualities for QT (and other quality) tags in stats output


dkj:
#### Is your feature request related to a problem? Please specify.
No.

#### Describe the solution you would like.
Just as `samtools stats` provides `FFQ` and `LFQ` sections for the qualities of the first and last fragments I'd appreciate similar sections for the qualities found in aux tags especially those in `QT`

dkj:
Repeat/similar to #795

PR #904, opened 2018-07-26
Add stats figures for auxiliary barcode tags

### LINKED ISSUE #795: samtools stats to report quality of bases per cycle for index (barcode) reads

valeriuo:
Compute base composition proportions for **BC**, **CR**, **OX**, **RX** tags.
Compute quality histograms for **QT**, **CY**, **BZ**, **QX** tags.
Fixes #795, #861


stats.c
```diff
@@ -688,6 +716,142 @@ void update_checksum(bam1_t *bam_line, stats_t *stats)
...
+                maxqual = &stats->maxqual_rx;
+                acgtno = stats->acgtno_rx;
+                quals = stats->quals_rx;
+                separator = &stats->separator_rx;
+            } else {
+                return 2;
+            }
+
+            if (barcode_len > nbases)
+                error("Barcodes differ in length at sequence '%s'. Aborting!\n", bam_get_qname(bam_line));
```
jkbonfield: This is an important check to avoid buffer overflows on the acgt_no and quals_no calloced buffers.  However it also means we can't run stats on mixed data sets if they have different barcode lengths.  I've no idea if that matters?  @dkj care to comment?  I seem to recall the possibility of different barcode lengths for 5' and 3' reads.

If in doubt, maybe it should simply be an option as to whether these stats are collected, so it can be disabled if needs be.

stats.c
```diff
@@ -688,6 +716,142 @@ void update_checksum(bam1_t *bam_line, stats_t *stats)
...
+    if (bc) {
+        char* barcode = bam_aux2Z(bc);
+        if (barcode) {
+            uint32_t barcode_len = strlen(barcode), nbases;
+            int *maxqual;
+            acgtno_count_t *acgtno;
+            uint64_t *quals;
+            int32_t *separator;
+
+            if (!strncmp(barcode_tag, "BC", 2)) {
```
jkbonfield: There is a lot of duplicated code for BC, CR, OX and RX.  I'm wondering if we need an array keyed on the ID.  Even at a 2d array of [128][128] it's still not a huge amount of memory (assuming we likewise malloc the actual data arrays the first time they're observed).  This would permit it to be adjustable and to work with whatever local tags people have created.  Example data with https://github.com/grenaud/deML for example uses XI/YI and XJ/YJ.

stats.c
```diff
@@ -688,6 +716,142 @@ void update_checksum(bam1_t *bam_line, stats_t *stats)
...
+                    acgtno[i].g++;
+                    break;
+                case 'T':
+                    acgtno[i].t++;
+                    break;
+                case 'N':
+                    acgtno[i].n++;
+                    break;
+                default:
+                    if (*separator > 0) {
```
jkbonfield: Separator is initialised to -1 and 0 can be a legal index (well, illegal probably, but it could be set to that with malformed data and the loop above would put it there).   So maybe this needs to be `if (*separator >= 0) {`.

PR #908, opened 2018-07-30
Fastq/fasta threading for output file.

### LINKED ISSUE #906: Default compression level for samtools fastq ?
```
  -c                   compression level [0..9] to use when creating gz or bgzf fastq files
```

The default is not mentioned.

jkbonfield:
Also fixes #906

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -780,7 +780,7 @@ static void bam2fq_usage(FILE *to, const char *command)
     if (fq) fprintf(to,
 "  -v INT               default quality score if not given in file [1]\n"
 "  -i                   add Illumina Casava 1.8 format entry to header (eg 1:N:0:ATCACG)\n"
-"  -c                   compression level [0..9] to use when creating gz or bgzf fastq files\n"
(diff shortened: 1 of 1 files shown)


jkbonfield:
Note: following samtools meeting, we proposed changing it so -c enabled compression (rather than being filename based).  I've given up on this for now as it's just too invasive and needs a major refactor of the code.

Specifically stdout ends up being gzipped even if we write nothing to it, so the common idioms of samtools fastq end up dumping binary to the terminal (most people don't specify things like -0 /dev/null to discard them).

If we're interested in gzipping 1.fq and 2.fq but not gzipping the things we discard, then it's a misfeature anyway to gzip before redirecting to /dev/null.  Frankly the entire program wants a new CLI such that no output is written unless explicitly asked for rather than vice versa.  It also needs to cope with specifying the same filename twice (so we can do `-1 - -2 -` to explicitly request 1 & 2 to stdout or a joint file, but that in turn really needs a dozen structure members renaming in order to keep it sane.

Given the complexity, that isn't this PR, but something new for another time.

PR #913, opened 2018-08-07
Markdup stats are easier to parse (valid YAML)

ryanrichholt:


whitwham:
I am inclined to accept this unless anyone has any strong objections to the change in format.

PR #920, opened 2018-08-13
add option to quickcheck to allow unmapped data

### LINKED ISSUE #914: samtools quickcheck returns error for unmapped BAM
1.9

Linux 3.2.0-105-generic
x86_64
gcc (GCC) 4.9.2

`samtools quickcheck` will return error when given an unmapped BAM file that is otherwise fine. It is failing to find targets in the header. I can see the reason for this, but maybe an option for when one expects unmapped data so we can skip this check?
```
$ samtools quickcheck -v 000033Fp01.bam
000033Fp01.bam had no targets in header.
```

mcshane:
Closes #914


samtools.1
```diff
@@ -1202,6 +1202,9 @@ regarding check results to be printed to stderr.
 .B -q
 Quiet mode: disables warning messages on stderr about files that fail.
 If both -q and -v options are used then the appropriate level of -v takes precedence.
+.TP
+.B -u
+Expect unmapped input data, so do require targets in the header.
```
daviesrob: Did you mean do **not** require targets here?

mcshane: Yep. Pushing a correction.

daviesrob:
Squashed and merged (apologies for force push on your branch).

PR #921, opened 2018-08-13
Update inputs for `reheader --in-place` docstring

### LINKED ISSUE #918: Clarify scope of `samtools reheader --in-place` in the documentation
Yes.

According to the [online documentation](http://www.htslib.org/doc/samtools.html), the `in-place` option for `samtools reheader` only applies to CRAM files ("This only works on CRAM files..."). However, the [docstring](https://github.com/samtools/samtools/blame/64852009960cb863b598ade1eb74f6bf3c3fa470/bam_reheader.c#L444) implies that the `in-place` operation works with both BAM and CRAM files ("Modify the bam/cram file directly").
It would be helpful to clarify in the documentation/docstring which of the two is correct.

juliangehring:
Updates the docstring of the `samtools reheader --in-place` option and clarifies that this only works with CRAM files (fixes #918).


daviesrob:
Thanks for the patch, docstring is now fixed.

PR #923, opened 2018-08-14
Spelling correction and CHK section tolerance

betteridiot:
Line 93: 1.8 had misspelled line 87, but this was fixed in version 1.9. However, 1.9 also misspelled 'Qualities" on line 93


betteridiot:
Also, see https://github.com/betteridiot/samtools/commit/9f2625187eed7d4c3aa0a2966d3fc1cd0bcccbd2 for adding CHK section tolerance.

daviesrob:
It might have been better to make a separate pull request for 9f26251, as this is now semi-merged...

You need to do more than just `CHK`, as 1.9 added some more keys.  It also needs to handle `FBC`, `FRL`, `LBC` and `LRL`.

daviesrob:
I'll merge the CHK part, and we'll look at the other ones here.  They're similar to some of the other stats, so it should be easy enough to work out what to do with them.

PR #924, opened 2018-08-14
Switched the comma in line 1187 to `and`

betteridiot:
In line 1187, the comma after `plot-bamstats` was being captured by the sanity check on Line 338. By switching the comma to `and`, this allows for the regex in the sanity check to still pass.


betteridiot:
I formatted that output to follow samtools stats output format as close as possible

daviesrob:
Thanks, I've squashed and merged as c03ee41.

PR #926, opened 2018-08-16
Minimal changes to allow plot-bamstats to work with new LBC FBC LRL F…

jenniferliddle:
Minimal changes to allow plot-bamstats to work with new LBC FBC LRL FRL and barcode tags sections.
By 'work with' I mean 'not fall over when it finds'.
If we actually want to graph the data in the new sections, we'll need a new PR.


daviesrob:
Possibly a bit more minimal than I was after.

I've pushed a revised version that does a bit more merging, and fixes a couple of other bugs.

jkbonfield:
Rebased and merged as c4ebb569bc3fdd2da6f619226181345e23c4d827

PR #928, opened 2018-08-23
Samtools stats fix for "none" barcodes

valeriuo:
Skip "none" barcodes.
Don't print the same error message multiple times.

### DIFF
--- a/stats.c
+++ b/stats.c
@@ -65,6 +65,7 @@ DEALINGS IN THE SOFTWARE.  */
 #define BWA_MIN_RDLEN 35
 #define DEFAULT_CHUNK_NO 8
 #define DEFAULT_PAIR_MAX 10000
+#define ERROR_LIMIT 200
 // From the spec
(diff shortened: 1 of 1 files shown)


stats.c
```diff
@@ -770,8 +771,9 @@ static void collect_barcode_stats(bam1_t* bam_line, stats_t* stats) {
                 break;
             default:
                 if (*separator >= 0) {
-                    if (*separator != i) {
-                        fprintf(stderr, "Barcode separator for tag %s is in a different position at sequence '%s'\n", barcode_tag, bam_get_qname(bam_line));
+                    if (*separator != i && !error_raised) {
```
daviesrob: This will fail to `continue` if the error has already been raised.  I think it needs to be:
```
if (*separator != i) {
    if (!error_raised) { ... }
    continue;
}
```

valeriuo: Thanks for spotting that! `continue` there was useless, anyway. 
The question now is whether we still want to process the rest of the barcode (and its associated quality tag) or  skip to the next tag.

daviesrob:
I think the limit to 200 may have got lost somewhere. Missing commit...?

PR #942, opened 2018-09-06
Make ax_with_curses.m4 test if -ltinfo is needed for linking

### LINKED ISSUE #940: /usr/bin/ld: bam_tview_curses.o: undefined reference to symbol 'cbreak'
Hi,
While trying to compile latest samtool-1.9 i get the following error

```
gcc -L/home/david/miniconda3/lib -L./lz4 -L/home/david/miniconda3/lib -o samtools bam_index.o bam_plcmd.o sam_view.o bam_cat.o bam_md.o bam_reheader.o bam_sort.o bedidx.o bam_rmdup.o bam_rmdupse.o bam_mate.o bam_stat.o bam_color.o bamtk.o bam2bcf.o bam2bcf_indel.o sample.o cut_target.o phase.o bam2depth.o padding.o bedcov.o bamshuf.o faidx.o dict.o stats.o stats_isize.o bam_flags.o bam_split.o bam_tview.o bam_tview_curses.o bam_tview_html.o bam_lpileup.o bam_quickcheck.o bam_addrprg.o bam_markdup.o tmp_file.o ./lz4/lz4.o libbam.a libst.a htslib-1.9/libhts.a -lpthread -lz -lm -lbz2 -llzma -ldeflate -lncurses -lm -lz  -lpthread
/usr/bin/ld: bam_tview_curses.o: undefined reference to symbol 'cbreak'
//usr/lib/libtinfo.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
Makefile:144: recipe for target 'samtools' failed
make: *** [samtools] Error 1
david@localhost ~/tools/samtools $ 
```

Seems related to ncurses. I have installed the latest ncurses-6.1 through conda .

Any idea ?

$ uname -a
Linux 4.10.0-38-generic 42 16.04.1-Ubuntu SMP Tue Oct 10 16:32:20 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
 $ gcc --version
gcc (Ubuntu 7.3.0-21ubuntu1~16.04) 7.3.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

daviesrob:
Some symbols for ncurses (and possibly others) actually come from libtinfo. Normally it's arranged that these are included when linking just with -lncurses, but this can break down in some cases (when the --no-add-needed linker option is used, and libncurses.so is a link to the real library instead of a linker script that adds the missing -ltinfo).
    
To make the tests in ax_with_curses.m4 resistant to this problem, add a test to see if -ltinfo is needed to get the cbreak() symbol, which is one of the ones affected.  If it is, -ltinfo gets appended to CURSES_LIB.

Fixes #940, and also configure failures where it finds the library but the header checks fail.  These checks do more than just look for the header - they actually try to link a program, and at least one of the symbols needed comes from libtinfo.

### DIFF
--- a/m4/ax_with_curses.m4
+++ b/m4/ax_with_curses.m4
@@ -220,6 +220,13 @@ AC_DEFUN([AX_WITH_CURSES], [
             AC_DEFINE([HAVE_NCURSESW], [1], [Define to 1 if the NcursesW library is present])
             AC_DEFINE([HAVE_CURSES],   [1], [Define to 1 if a SysV or X/Open compatible Curses library is present])
 
+            # Normally linking ncursesw gets tinfo as well, but that doesn't
+            # always happen.  Check to see if it's needed as well.
+            AC_SEARCH_LIBS(cbreak, tinfo)
+            AS_CASE([$ac_cv_search_cbreak],
+                    [-l*],[CURSES_LIB="$CURSES_LIB $ac_cv_search_cbreak"
+                           LIBS="$LIBS $ac_cv_search_cbreak"])
+
             AC_CACHE_CHECK([for working ncursesw/curses.h], [ax_cv_header_ncursesw_curses_h], [
                 AC_LINK_IFELSE([AC_LANG_PROGRAM([[
                         @%:@define _XOPEN_SOURCE_EXTENDED 1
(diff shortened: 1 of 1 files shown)


jmarshall:
I've tested this in three scenarios: with ncurses⇒tinfo well-built shared libraries; with the conda shared libraries; and with only static libraries being available. This patch doesn't affect the first (:+1:); rescues the current conda situation; but configuring still fails when only _libncurses.a_/_libtinfo.a_ are present.

Conda will likely soon move to the first scenario (conda-forge/ncurses-feedstock#44). Maybe that makes this _ax_with_curses.m4_ fix moot, or maybe it encourages extending it to handle the static scenario too…

kyleabeauchamp:
Thanks for being on top of this and digging in to the related conda issues!

whitwham:
Replaced by #1170

PR #950, opened 2018-09-14
Fix incorrect macro definition on Windows

daviesrob:
_POSIX_C_SOURCE should have been _XOPEN_SOURCE.
The value given to _POSIX_C_SOURCE caused the strdup() and M_PI definitions to disappear.

Not sure quite how Appveyor failed to spot this, but I think it may be because it builds with htslib in a subdirectory of samtools.  It failed for me when they were in side-by-side directories.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -129,7 +129,7 @@ case $host_alias in
     # This also sets __USE_MINGW_ANSI_STDIO which in turn makes PRId64,
     # %lld and %z printf formats work.  It also enforces the snprintf to
     # be C99 compliant so it returns the correct values (in kstring.c).
-    CPPFLAGS="$CPPCFLAGS -D_POSIX_C_SOURCE=600"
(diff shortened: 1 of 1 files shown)


jmarshall:
Nice catch. I guess this explains samtools/htslib#707…

daviesrob:
Yes.  To add to the confusion, there are at least two sets of include files you can get with MSYS2.  If you got them from `mingw-w64-x86_64-headers-git` (possibly via `mingw-w64-x86_64-toolchain`) then the build worked.  This is the version that Appveyor uses.  If you got them from `msys2-runtime-devel` you ended up with cygwin-style headers which are a bit more strict about POSIX versions and the build failed.

I only noticed because I happened to download the msys2 version.

valeriuo:
Closed by 5708485f861c096d50471a70bb996561a0514181

Issue #951, opened 2018-09-16
Adding gcc-8 case on Travis CI


junaruga:
Thank you for maintaining `samtools`.
I would like to add the latest version GCC gcc-8 test case on Travis.

## Background

Right now the used gcc version in this repository is GCC 4.8, that is default setting of Travis CI.

I think it is meaningful to test both old version gcc-4.8 and latest version gcc-8 for people who use `samtools` on newer os platform.

I am going to send the pull-request soon as a proposal.

## References

If you want to know more about adding gcc-8 on Travis, here is the official document.

https://docs.travis-ci.com/user/languages/c/#gcc-on-linux
https://docs.travis-ci.com/user/installing-dependencies/#adding-apt-sources-1

We can refer some cases.

### Case: Ruby

Ruby is using Travis syntax as much as possible to add gcc-8 case.

https://github.com/ruby/ruby/blob/trunk/.travis.yml
https://travis-ci.org/ruby/ruby

Interesting feature is that they are doing a hack in `configure.ac` not to set `AR=gcc-ar-8 NM=gcc-nm-8 RANLIB=gcc-ranlib-8` environment, predicting the values from `CC=gcc-8`.
https://github.com/ruby/ruby/commit/f677ba8

### Case: nanopolish

Analysis tool for Oxford Nanopore sequencing data.
This tool is using shell commands to add gcc-8 case.

https://github.com/jts/nanopolish/blob/master/.travis.yml
https://travis-ci.org/jts/nanopolish

### Case: Trinity

This tools is using shell commands to add gcc-8 case too.
You can refer gcc-8 case on both linux and osx here.

https://github.com/trinityrnaseq/trinityrnaseq/blob/master/.travis.yml
https://travis-ci.org/trinityrnaseq/trinityrnaseq

kyleabeauchamp:
On a related note Bioconda + Conda-forge are in the process of migrating everything to GCC 7.2.  So in a month or so I expect the Bioconda packages to be done on a new-ish compiler.

junaruga:
@kyleabeauchamp thank you for the useful note.

In the pull-request #952 , I used "ubuntu-toolchain-r-test" repository.
That means we can use below gcc set in the list.
https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test

When we want to test on GCC 7.2 aligning with Bioconda environment, if we add the case `gcc-7` in `.travis.yml`, GCC 7.2 is used.

junaruga:
[...]

PR #952, opened 2018-09-16
Add linux gcc-8 case to Travis CI.

junaruga:
This pull-request is related to https://github.com/samtools/samtools/issues/951 .

In my repository, I tested it. Here is the result.
You can check how it looks like.
https://travis-ci.org/junaruga/samtools/builds/429213754

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,29 @@ matrix:
       sudo: required
       env: CFLAGS="-fsanitize=address" LDFLAGS="-fsanitize=address"
     - os: linux
-      compiler: gcc
+      compiler: gcc-8
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-8
+      env:
+        - CC=gcc-8
+        - AR=gcc-ar-8
     # An unoptimised C99 build, for detecting non-static inline functions
     - os: linux
-      compiler: gcc
-      env: CFLAGS="-std=gnu99 -O0"
+      compiler: gcc-8
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-8
+      env:
+        - CC=gcc-8
+        - AR=gcc-ar-8
+        - CFLAGS="-std=gnu99 -O0"
     - os: osx
       compiler: clang
 


daviesrob:
Thanks for the pull request.  Generally I think this is a good idea, although I do have a few observations:

- We should upgrade [htslib's Travis configuration ](https://github.com/samtools/htslib/blob/develop/.travis.yml) to try gcc-8 as well.  We don't want to rely on samtools to test parts of HTSlib.
- We might want to make gcc-8 the default and have the older compiler as the special case (if we bother with it at all).  Gcc-8 spots more problems than gcc-4, so it would be a better first choice.
- We don't want to run too many tests, so Travis finishes in a reasonable time.  This is more of an issue for HTSlib as it runs more cases.
- Samtools and HTSlib do get tested against gcc-8, albeit on Windows, by Appveyor.  I've also been using gcc 8.1 for a while, so there shouldn't be many gcc-8 related problems left by now.

junaruga:
@daviesrob thanks for the observations. I basically agree for that.

> We should upgrade htslib's Travis configuration to try gcc-8 as well. We don't want to rely on samtools to test parts of HTSlib.

Yes, we need to update `htslib`'s configuration too. Actually that was my next plan after `samtools`.
So, after `samtools`, I can send pull-request to `htslibs` as well.

> We might want to make gcc-8 the default and have the older compiler as the special case (if we bother with it at all). Gcc-8 spots more problems than gcc-4, so it would be a better first choice.

I can agree with it.

Current test cases are

```
1. linux & clang
2. linux & gcc-4.8
3. linux & gcc-4.8 with "C99 and -O0"
4. osx & clang
```
And below test cases are your preference as a first choice right?
The case `osx & gcc-8` can be included as next choice?

```
1. linux & clang
2. linux & gcc-8
3. linux & gcc-8 with "C99 and -O0"
4. osx & clang
```

> We don't want to run too many tests, so Travis finishes in a reasonable time. This is more of an issue for HTSlib as it runs more cases.

Alright. That makes sense.

> Samtools and HTSlib do get tested against gcc-8, albeit on Windows, by Appveyor. I've also been using gcc 8.1 for a while, so there shouldn't be many gcc-8 related problems left by now.

That makes sense. But as I am not familiar with Windows and Appveyor, I would like to ask someone for the work.

Maybe I will update this PR tonight or tomorrow.

junaruga:
I rebased with below way!

```
1. linux & clang
2. linux & gcc-8
3. linux & gcc-8 with "C99 and -O0"
4. osx & clang
```

junaruga:
@daviesrob ping.

daviesrob:
Sorry for the delay, this has been waiting on a companion in htslib (see samtools/htslib#781)

junaruga:
@daviesrob alright! Thanks for the working of `htslib`.

junaruga:
`htslib` with `gcc-8` is merged! Now it's time to merge for this PR.

valeriuo:
Merged

PR #960, opened 2018-10-09
Minor enhancements to bam_cat

jenniferliddle:
Now supports --output-fmt-option and --output-fmt command line options.
Added @PG line to output file

### DIFF
--- a/bam_cat.c
+++ b/bam_cat.c
@@ -47,6 +47,7 @@ Illumina.
 #include "htslib/cram.h"
 #include "htslib/khash.h"
 #include "samtools.h"
+#include "sam_opts.h"
 
 KHASH_MAP_INIT_STR(s2i, int)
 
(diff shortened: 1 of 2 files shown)


bam_cat.c
```diff
@@ -604,13 +648,13 @@ int main_cat(int argc, char *argv[])
     switch (hts_get_format(in)->format) {
```
valeriuo: The output file format is still decided by the format of the first input file, which makes the `--output-fmt` option useless.

jenniferliddle: Done

bam_cat.c
```diff
@@ -313,6 +342,7 @@ int cram_cat(int nfn, char * const *fn, const bam_hdr_t *h, const char* outcram)
     cram_set_option(out_c, CRAM_OPT_VERSION, vers);
     //fprintf(stderr, "Creating cram vers %s\n", vers);
 
+    if (addNewPG(new_h, arg_list)) return -1;
```
valeriuo: You should switch to using the new header API method `bam_hdr_add_pg`.

valeriuo:
Merged as 3e81cc099fb4d485019f3a73d66139d74ec08d05 into https://github.com/samtools/samtools/pull/998

PR #985, opened 2018-11-25
Free memory when failing to set CRAM options

fbreitwieser:
Currently, bam2depth just exits with `1` when there is an error setting
`CRAM_OPT_REQUIRED_FIELDS` or `CRAM_OPT_DECODE_MD`.

This patch sets the status to `EXIT_FAILURE` when there is an error, and gives
control to `depth_end`, where files are closed and memory is freed. 

Additionally, `EXIT_FAILURE` is returned instead of `1` in some other cases, and `print_error_errno` is used instead of `fprintf`.

### DIFF
--- a/bam2depth.c
+++ b/bam2depth.c
@@ -97,7 +97,7 @@ static int usage() {
     fprintf(stderr, "omitted by default; see the -a option.\n");
     fprintf(stderr, "\n");
 
-    return 1;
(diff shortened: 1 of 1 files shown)


jkbonfield:
Thanks for the fix and apologies for the slow action; I'd forgotten about this.  Rebased and merged it as b3fe80a after a minor white space fix.

PR #992, opened 2018-12-11
Add command samtools coverage

### LINKED ISSUE #596: average coverage and coverage plots
For our read mappings, we always check for average coverage and it's nice to have a plot of coverage as a function of position as well. Lots of people have written their own scripts for this, but it would be helpful if this was standardized in `samtools` and `plot-bamstats`, especially since these produce very useful reports that are just missing a couple things like this.

Many thanks and regards

fbreitwieser:
This pull request adds the new command  samtools coverage` that prints a tabular format of the average coverage and percent coverage for each sequence, as well as number of aligned reads, average mapQ and baseQ. Using the parameter '-m', it plots a histogram of the
percent covered along the genome.

Fixes #596.

### DIFF
--- a/coverage.c
+++ b/coverage.c
@@ -1,6 +1,6 @@
 /* coverage.c -- samtools coverage subcommand
- 
-   Author: Florian P Breitwieser <florian.bw@gmail.com>
+
+Author: Florian P Breitwieser <florian.bw@gmail.com>
 
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -42,21 +42,21 @@ DEALINGS IN THE SOFTWARE.  */
 #include <unistd.h>
 
 #if defined (__has_include)
-  #if __has_include(<sys/ioctl.h>)
-    #include <sys/ioctl.h>
-    #define has_ioctl
-  #elif __has_include(<windows.h>)
-    #include <windows.h>
-    #define has_windows
-  #endif
+#if __has_include(<sys/ioctl.h>)
+#include <sys/ioctl.h>
+#define has_ioctl
+#elif __has_include(<windows.h>)
+#include <windows.h>
+#define has_windows
+#endif
 #else
-  #if defined (__MSYS__) || defined (__CYGWIN__) || defined (_WIN32)
-    #include <windows.h>
-    #define has_windows
-  #else
-    #include <sys/ioctl.h>
-    #define has_ioctl
-  #endif
+#if defined (__MSYS__) || defined (__CYGWIN__) || defined (_WIN32)
+#include <windows.h>
+#define has_windows
+#else
+#include <sys/ioctl.h>
+#define has_ioctl
+#endif
 #endif
 
 #include "htslib/sam.h"
@@ -120,38 +120,38 @@ int read_file_list(const char *file_list, int *n, char **argv[]);
 
 static int usage() {
     fprintf(stdout, "Usage: samtools coverage [options] in1.bam [in2.bam [...]]\n\n"
-                    "Input options:\n"
-                    "  -b, --bam-list FILE     list of input BAM filenames, one per line\n"
-                    "  -l, --min-read-len INT  ignore reads shorter than INT bp [0]\n"
-                    "  -q, --min-MQ INT        base quality threshold [0]\n"
-                    "  -Q, --min-BQ INT        mapping quality threshold [0]\n"
-                    "  --rf <int|str>          required flags: skip reads with mask bits unset []\n"
-                    "  --ff <int|str>          filter flags: skip reads with mask bits set \n"
-                    "                                      [UNMAP,SECONDARY,QCFAIL,DUP]\n"
-                    "Output options:\n"
-                    "  -m, --histogram         show histogram instead of tabular output\n"
-                    "  -M, --Histogram         show histogram with full UTF8 mode for finer resolution\n"
-                    "  -o, --output FILE       write output to FILE [stdout]\n"
-                    "  -H, --no-header         don't print a header in tabular mode\n"
-                    "  -w, --n-bins INT        number of bins in histogram [terminal width - 40]\n"
-                    "  -r, --region REG        show specified region. Format: chr:start-end. \n"
-                    "  -h, --help              help (this page)\n");
+            "Input options:\n"
+            "  -b, --bam-list FILE     list of input BAM filenames, one per line\n"
+            "  -l, --min-read-len INT  ignore reads shorter than INT bp [0]\n"
+            "  -q, --min-MQ INT        base quality threshold [0]\n"
+            "  -Q, --min-BQ INT        mapping quality threshold [0]\n"
+            "  --rf <int|str>          required flags: skip reads with mask bits unset []\n"
+            "  --ff <int|str>          filter flags: skip reads with mask bits set \n"
+            "                                      [UNMAP,SECONDARY,QCFAIL,DUP]\n"
+            "Output options:\n"
+            "  -m, --histogram         show histogram instead of tabular output\n"
+            "  -A, --ascii             show only ASCII characters in histogram\n"
+            "  -o, --output FILE       write output to FILE [stdout]\n"
+            "  -H, --no-header         don't print a header in tabular mode\n"
+            "  -w, --n-bins INT        number of bins in histogram [terminal width - 40]\n"
+            "  -r, --region REG        show specified region. Format: chr:start-end. \n"
+            "  -h, --help              help (this page)\n");
 
     fprintf(stdout, "\nGeneric options:\n");
     sam_global_opt_help(stdout, "-.--.-");
 
     fprintf(stdout, 
             "\nSee manpage for additional details.\n"
-                    "  rname       Reference name / chromosome\n"
-                    "  startpos    Start position\n"
-                    "  endpos      End position (or sequence length)\n"
-                    "  numreads    Number reads aligned to the region (after filtering)\n"
-                    "  covbases    Number of covered bases with depth >= 1\n"
-                    "  coverage    Proportion of covered bases [0..1]\n"
-                    "  meandepth   Mean depth of coverage\n"
-                    "  meanbaseq   Mean baseQ in covered region\n"
-                    "  meanmapq    Mean mapQ of selected reads\n"
-                    );
+            "  rname       Reference name / chromosome\n"
+            "  startpos    Start position\n"
+            "  endpos      End position (or sequence length)\n"
+            "  numreads    Number reads aligned to the region (after filtering)\n"
+            "  covbases    Number of covered bases with depth >= 1\n"
+            "  coverage    Proportion of covered bases [0..1]\n"
+            "  meandepth   Mean depth of coverage\n"
+            "  meanbaseq   Mean baseQ in covered region\n"
+            "  meanmapq    Mean mapQ of selected reads\n"
+           );
 
     return EXIT_SUCCESS;
 }
@@ -222,21 +222,21 @@ void print_tabular_line(FILE *file_out, const bam_hdr_t *h, const stats_aux_t *s
             stats->summed_coverage / region_len,
             stats->summed_coverage > 0? stats->summed_baseQ/(double) stats->summed_coverage : 0,
             stats->n_selected_reads > 0? stats->summed_mapQ/(double) stats->n_selected_reads : 0
-            );
+           );
 }
 
 void print_hist(FILE *file_out, const bam_hdr_t *h, const stats_aux_t *stats, const uint32_t *hist, 
-                const int hist_size, const bool full_utf) {
+        const int hist_size, const bool full_utf) {
     int i, col;
     bool show_percentiles = false;
     const int n_rows = 10;
     const char * const * BLOCK_CHARS = full_utf? BLOCK_CHARS8 : BLOCK_CHARS2;
     const int blockchar_len = full_utf? 8 : 2;
     /*
-    if (stats->beg == 0) {
-        stats->end = h->target_len[stats->tid];
-    }
-    */
+       if (stats->beg == 0) {
+       stats->end = h->target_len[stats->tid];
+       }
+       */
     double region_len = stats->end - stats->beg;
 
     // Calculate histogram that contains percent covered
@@ -249,7 +249,7 @@ void print_hist(FILE *file_out, const bam_hdr_t *h, const stats_aux_t *stats, co
 
     char buf[100];
     fprintf(file_out, "%s (%sbp)\n", h->target_name[stats->tid], readable_bps(h->target_len[stats->tid], buf));
-    
+
     double row_bin_size = max_val / (double) n_rows;
     for (i = n_rows-1; i >= 0; --i) {
         double current_bin = row_bin_size * i;
@@ -333,7 +333,7 @@ int main_coverage(int argc, char *argv[]) {
     bool opt_print_tabular = true;
     bool opt_print_histogram = false;
     bool *covered_tids;
-    bool opt_full_utf = false;
+    bool opt_full_utf = true;
 
     FILE *file_out = stdout;
 
@@ -351,7 +351,7 @@ int main_coverage(int argc, char *argv[]) {
         {"min-BQ", required_argument, NULL, 'Q'},
         {"min-bq", required_argument, NULL, 'Q'},
         {"histogram", no_argument, NULL, 'm'},
-        {"Histogram", no_argument, NULL, 'M'},
+        {"ascii", no_argument, NULL, 'A'},
         {"output", required_argument, NULL, 'o'},
         {"no-header", required_argument, NULL, 'H'},
         {"n-bins", required_argument, NULL, 'w'},
@@ -364,41 +364,41 @@ int main_coverage(int argc, char *argv[]) {
     // parse the command line
     int c;
     while ((c = getopt_long(argc, argv, "o:L:q:Q:hHw:r:b:mM", lopts, NULL)) != -1) {
-            switch (c) {
-                case 1:
-                        if ((required_flags = bam_str2flag(optarg)) < 0) {
-                             fprintf(stderr,"Could not parse --rf %s\n", optarg); return EXIT_FAILURE;
-                         }; break;
-                case 2:
-                        if ((fail_flags = bam_str2flag(optarg)) < 0) {
-                             fprintf(stderr,"Could not parse --ff %s\n", optarg); return EXIT_FAILURE;
-                         }; break;
-                case 'o': opt_output_file = optarg; opt_full_width = false; break;
-                case 'L': opt_min_len = atoi(optarg); break;
-                case 'q': opt_min_baseQ = atoi(optarg); break;
-                case 'Q': opt_min_mapQ = atoi(optarg); break;
-                case 'w': opt_n_bins = atoi(optarg); opt_full_width = false;
-                          opt_print_histogram = true; opt_print_tabular = false; 
-			              break;
-                case 'r': opt_reg = optarg; break;   // parsing a region requires a BAM header (strdup unnecessary)
-                case 'b': opt_file_list = optarg; break;
-                case 'm': opt_print_histogram = true; opt_print_tabular = false; break;
-                case 'M': opt_full_utf = true; 
-                          opt_print_histogram = true; opt_print_tabular = false; 
-                          break;
-                case 'H': opt_print_header = false; break;
-                case 'h': return usage();
-                default:  if (parse_sam_global_opt(c, optarg, lopts, &ga) == 0) break;
-                      /* else fall-through */
-                case '?': 
-                   if (optopt == 'c')
-                      fprintf (stderr, "Option -%c requires an argument.\n", optopt);
-                    else if (isprint (optopt))
-                      fprintf (stderr, "Unknown option `-%c'.\n", optopt);
-                    else
-                      fprintf (stderr, "Unknown option character `\\x%x'.\n", optopt);
-                    return 1;
-            }
+        switch (c) {
+            case 1:
+                if ((required_flags = bam_str2flag(optarg)) < 0) {
+                    fprintf(stderr,"Could not parse --rf %s\n", optarg); return EXIT_FAILURE;
+                }; break;
+            case 2:
+                if ((fail_flags = bam_str2flag(optarg)) < 0) {
+                    fprintf(stderr,"Could not parse --ff %s\n", optarg); return EXIT_FAILURE;
+                }; break;
+            case 'o': opt_output_file = optarg; opt_full_width = false; break;
+            case 'L': opt_min_len = atoi(optarg); break;
+            case 'q': opt_min_baseQ = atoi(optarg); break;
+            case 'Q': opt_min_mapQ = atoi(optarg); break;
+            case 'w': opt_n_bins = atoi(optarg); opt_full_width = false;
+                      opt_print_histogram = true; opt_print_tabular = false; 
+                      break;
+            case 'r': opt_reg = optarg; break;   // parsing a region requires a BAM header (strdup unnecessary)
+            case 'b': opt_file_list = optarg; break;
+            case 'm': opt_print_histogram = true; opt_print_tabular = false; break;
+            case 'A': opt_full_utf = false; 
+                      opt_print_histogram = true; opt_print_tabular = false; 
+                      break;
+            case 'H': opt_print_header = false; break;
+            case 'h': return usage();
+            default:  if (parse_sam_global_opt(c, optarg, lopts, &ga) == 0) break;
+                          /* else fall-through */
+            case '?': 
+                          if (optopt == 'c')
+                              fprintf (stderr, "Option -%c requires an argument.\n", optopt);
+                          else if (isprint (optopt))
+                              fprintf (stderr, "Unknown option `-%c'.\n", optopt);
+                          else
+                              fprintf (stderr, "Unknown option character `\\x%x'.\n", optopt);
+                          return 1;
+        }
     }
     if (optind == argc && !opt_file_list) 
         return usage();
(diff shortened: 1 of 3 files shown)


coverage.c
```diff
@@ -0,0 +1,590 @@
...
+
+
+    h = data[0]->hdr; // easy access to the header of the 1st BAM
+    covered_tids = calloc(h->n_targets, sizeof(bool));
+    stats_aux_t *stats = calloc(1, sizeof(stats_aux_t));
+
+    bool reg_sets_end = false;
+    if (opt_reg) {
+        stats->beg = data[0]->iter->beg; // and to the parsed region coordinates
+        stats->end = data[0]->iter->end;
```
jkbonfield: This works only provided we fully specify the reference.  Eg 20:1000000-20000000 will show 10M to 20M on the X-axis.  However if we just do "20" then it'll go from 1 to INT_MAX (2.147Gb).  It should check for these cases and lookup in the header to find the sequence extents.  There is a `target_len` array in bam_hdr_t for this (indexed on tid).

coverage.c
```diff
@@ -0,0 +1,590 @@
...
+                    memset(hist, 0, opt_n_bins*sizeof(uint32_t));
+            }
+
+            stats->tid = tid;
+            covered_tids[tid] = true;
+            if (!reg_sets_end)
+                stats->end = h->target_len[tid];
+
+            if (compute_histogram) 
+                stats->bin_width = (stats->end-stats->beg) / opt_n_bins;
```
jkbonfield: This needs a check to ensure bin_width never hits zero.  Maybe adjust opt_n_bins if it does to guarantee it cannot happen.

Currently with a tiny region the current_bin calculation below gets a divide by zero and crashes.

coverage.c
```diff
@@ -0,0 +1,590 @@
...
+    int32_t tid;    // chromosome ID, defined by header
+    int beg;
+    int end;
+    int bin_width;
+} stats_aux_t;
+
+int read_file_list(const char *file_list, int *n, char **argv[]);
+
+static int usage(int status) {
+    fprintf(stderr, "\nUsage: samtools coverage [options] in1.bam [in2.bam [...]]\n"
```
jkbonfield: Ideally usage statements should go to stdout unless they are infact an error (eg someone tried "samtools coverage -x").  Possibly stdout always is good enough given the "coverage: invalid option -- 'x'" will be stderr anyway.

I know a lot of our other commands use stderr, but we also have this listed in the test harness as "known failures" for us to fix one day.

coverage.c
```diff
@@ -0,0 +1,590 @@
...
+                    );
+
+    return status;
+}
+
+static int version(int status) {
+    fprintf(stderr,
+    "┌┐ ┌─┐┌┬┐┌─┐┌─┐┬  ┬\n"
+    "├┴┐├─┤││││  │ │└┐┌┘\n"
+    "└─┘┴ ┴┴ ┴└─┘└─┘ └┘   v%s\n", VERSION);
```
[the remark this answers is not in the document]
jmarshall: As a subcommand integrated into samtools, does this really need to retain its own version number? Does `samtools --version` not suffice?

jkbonfield:
I am unsure of whether the \uxxxx unicode syntax arrived in C99 or C11 as sources differ (gcc says C99, stackoverflow implies C11, although I'm more inclined to trust gcc).  Older compilers are whinging at this, but still appear to work.

I am wondering if there is a better alternative to using unicode block symbols, eg `#`, with some ifdefs to detect sufficient C version compatibility to permit unicode support when and only when the appropriate -stdc=c99 type option has been specified (or when the compiler is new enough that it is the default).

coverage.c
```diff
@@ -0,0 +1,590 @@
...
+                    "  1. Reference name / chromosome\n"
+                    "  2. Start position\n"
+                    "  3. End position (or sequence length)\n"
+                    "  4. Number reads aligned to the region\n"
+                    "  5. Number of covered bases with depth >= 1\n"
+                    "  6. Percent of covered bases\n"
+                    "  7. Average coverage depth\n"
+                    "  8. Average baseQ in covered region\n"
+                    "  9. Average mapQ of selected reads\n"
+                    );
```
sjackman: I'd recommend outputting a one-line TSV header.
```
rname startpos endpos numreads numcov percentcov meandepth meanbaseq meanmapq
```

coverage.c
```diff
@@ -0,0 +1,590 @@
...
+        stats->summed_mapQ += data[i]->summed_mapQ;
+        data[i]->n_reads = 0;
+        data[i]->n_selected_reads = 0;
+        data[i]->summed_mapQ = 0;
+    }
+}
+
+// UTF8 specifies block characters in eights going from \u2581 (lower one eight block) to \u2588 (full block)
+//   https://en.wikipedia.org/wiki/Block_Elements
+const char* BLOCK_CHARS[8] = {"\u2581", "\u2582", "\u2583", "\u2584", "\u2585", "\u2586", "\u2587", "\u2588"};
```
sjackman: This code should work portably:
```c++
	/** Unicode bar characters. */
	static const char* BLOCK_CHARS[10] = {
		" ", "_",
		"\342\226\201", // 9601
		"\342\226\202", // 9602
		"\342\226\203", // 9603
		"\342\226\204", // 9604
		"\342\226\205", // 9605
		"\342\226\206", // 9606
		"\342\226\207", // 9607
		"\342\226\210", // 9608
	};
```
See https://github.com/bcgsc/abyss/blob/d4b4b5d3091d90a4967180d987bd7168dbf04585/Common/Histogram.cpp#L71-L82

jmarshall:
My purchased copy of the C99 IS :smile: confirms that \uxxxx is a C99 feature. (C11 introduces `u8"foo"` to make a UTF-8-encoded string, which is perhaps confounding what you found on SO.) [GCC 5](https://www.gnu.org/software/gcc/gcc-5/changes.html) is the first GCC that defaults to ≥C99; before that GCC would complain about this notation without `-std=c99` or so.

Whether the compiler supports the \uxxxx notation is a separate question from whether your terminal can display the Unicode characters produced. Even if the program is compiled with a compiler that doesn't understand \u, the desired characters can probably be displayed on just about any terminal in use today.

The `\u` notation causes the requested Unicode character to be put into the string in an appropriate representation for the platform. @sjackman's `"\342\226\201"` hard-codes the UTF-8 representation of the corresponding Unicode character. These are the same thing on sensible platforms that use UTF-8, but not on other platforms—if there are any. (Does Windows still use UTF-16 in this context? I have no idea. But if so, `"\342\226\201"` would be less portable…)

So my suggestion would be something like

```c
#if __STDC_VERSION__ >= 199901L
#define VERTICAL_LINE "\u2502" // BOX DRAWINGS LIGHT VERTICAL

// LOWER ONE EIGHTH BLOCK … FULL BLOCK
static const char *const BLOCK_CHARS[] = { "\u2581", …, "\u2588" };

#else
// Fall back to explicit UTF-8 encodings of the same characters
#define VERTICAL_LINE "\xE2\x94\x82"
static const char *const BLOCK_CHARS[] = { "\xE2\x96\x81", …, "\xE2\x96\x88" };
#endif
```

where `VERTICAL_LINE` is a `#define` so that it can be concatenated into the various `printf` strings that `\u2502` currently somewhat obscurely appears in. And raw UTF-8 is more familiar (and google-able!) in hex than in octal, hence the `\x`s.

sjackman:
In which C/C++ standard was `"\xFF"' introduced?

jkbonfield:
That's positively ancient.  No sure when, but likely K&R era.

jkbonfield:
Anyway, I think this is the best approach for dealing with inline unicode.  If anyone with windows experience can inform us of whether utf16 is still a thing and if so whether it also supports inline utf8 printing and "does the right thing", then it'll help us to simplify it.

This isn't major functionality anyway as it's only the pictorial view.  If it doesn't work in some whacky terminal / OS combo then so be it; the table form permits you to load it into a spread-sheet and roll your own plot.

sjackman:
You're right that the hexadecimal escape sequence goes back to at least C89. No idea then why I used octal when I wrote this code. `¯\_(ツ)_/¯`
https://en.cppreference.com/w/c/language/escape

jmarshall:
First edition K&R (1978) has only octal for character and string constants. I'm pretty sure `"\xFF"` was available in the pre-Standard K&R compiler I used in the late 80s. It was already there in the first ever C  Standard (C89) and the [C89 Rationale](https://www.lysator.liu.se/c/rat/c1.html#3-1-3-4) notes “Hexadecimal escape sequences, beginning with \x, have been adopted, with precedent in several existing implementations”.

sjackman:
Heh. Nice find in the rationale.

fbreitwieser:
Thanks a lot for going through the code and the all comments, @jkbonfield, @jmarshall and @sjackman ! And also for approving the inclusion upon adding the changes. I'm reading though the comments now and will address each of them in the next days. 

I was working on a bit of a cleanup and fixes and cleanups after I sent the the pull request to samtools, which are in the repo https://github.com/fbreitwieser/bamcov. I may merge some changes which are appropriate from there.

Regarding the UTF8 block characters, one issue that came up for a Windows Putty user was that only half and full blocks were shown, for the eight's and quarter's it just showed the dreaded white square □. Maybe the hard-coded representation fixes that? I'll try. Otherwise should it just fall back to using just half and full blocks by default (which I did in the bamcov repo)?

sjackman:
I'd be surprised if Putty didn't support Unicode. I think perhaps if you change the Putty font, it may work. See https://serverfault.com/questions/378352/can-putty-be-configured-to-display-the-following-utf-8-characters

sjackman:
```
                    "  1. Reference name / chromosome\n"
	                    "  2. Start position\n"
	                    "  3. End position (or sequence length)\n"
	                    "  4. Number reads aligned to the region\n"
	                    "  5. Number of covered bases with depth >= 1\n"
	                    "  6. Percent of covered bases\n"
	                    "  7. Average coverage depth\n"
	                    "  8. Average baseQ in covered region\n"
	                    "  9. Average mapQ of selected reads\n"
```
Is start position always `1`, or is it the first covered base?
Is end position always the sequence length, or is it the last covered base?
Perhaps add "Number of bases aligned to the region". Number of aligned bases divided by number of covered bases yields mean depth.
Change `Average` to `Mean` in three cases.
I suggest proportion of covered bases (0–1) rather than percent (0–100).
Could you please report median depth of coverage? I find it a very useful metric when sequencing depth is uneven.

One example of this (though there's many) is aligning whole genome sequencing data to the mitochondrial genome of an organism that does not yet have a nuclear genome reference. There's probably more common examples.

fbreitwieser:
Re Putty - I think you are right it's the font. Rather annoying of Putty to not use a font by default that supports UTF8.

Regarding the questions:
- Is start position always 1, or is it the first covered base?
- Is end position always the sequence length, or is it the last covered base?

If no region is specified, the start position is always 1 and the end position is always the sequence length. When a region is specified, they are region begin and end, resp.. 

- Perhaps add "Number of bases aligned to the region". Number of aligned bases divided by number of covered bases yields mean depth.

Can do that. Is there a special use for this number over the mean depth?

- Change Average to Mean in three cases.

Will do, thanks for the suggestion.

- I suggest proportion of covered bases (0–1) rather than percent (0–100).
- Could you please report median depth of coverage? I find it a very useful metric when sequencing depth is uneven. One example of this (though there's many) is aligning whole genome sequencing data to the mitochondrial genome of an organism that does not yet have a nuclear genome reference. There's probably more common examples.

Thanks for the good suggestions! I'd like to add the median depth. It will need to keep the depths at each position in memory, though, unless we go for an approximate solution (eg count-min-sketch).

coverage.c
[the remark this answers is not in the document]
fbreitwieser: You're right, I'll leave out the version command.

Regarding the man page - working on it.

coverage.c
```diff
@@ -0,0 +1,590 @@
...
+
+    int *n_plp;
+    bam_hdr_t *h = NULL; // BAM header of the 1st input
+
+    bool opt_print_header = false;
+    bool opt_print_tabular = true;
+    bool opt_print_histogram = false;
+    bool *covered_tids;
+
+    int current_tid = -1;
```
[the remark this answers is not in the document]
fbreitwieser: Fixed

sjackman:
>> Perhaps add "Number of bases aligned to the region". Number of aligned bases divided by number of covered bases yields mean depth.

> Can do that. Is there a special use for this number over the mean depth?

The only one I can think of is that total aligned bases divided by the total number of sequenced bases gives you an estimate of the proportion of sequencing that is on target. (I've never actually done this.) Feel free to ignore this request.

coverage.c
fbreitwieser: Good point, made usage print to stdout and always return EXIT_SUCCESS.

coverage.c
fbreitwieser: Should the header be outputted by default? I'm for it, but I think most tools currently don't.  `samtools view` uses the `-h` flag to show the header, but also the SAM header is more involved than a simple TSV header. 

Other question: What should I change the percent-covered column name when reporting a number between 0 and 1? `proportioncov`? Can't think of a good name

sjackman: > Should the header be outputted by default? I'm for it, but I think most tools currently don't. samtools view uses the -h flag to show the header, but also the SAM header is more involved than a simple TSV header.

I would output the header by default. An option to disable the header would be useful (perhaps `--no-header`). `samtools coverage | tail +n2` works as well.

> Other question: What should I change the percent-covered column name when reporting a number between 0 and 1? proportioncov? Can't think of a good name

I'd suggest simply `coverage`. To me, genome coverage is the proportion of the genome that is covered, whereas depth of coverage is, well, the depth. =)
I'd suggest changing `numcov` to `coveredbases` or perhaps just `covbases`.

jkbonfield:
I can confirm the Putty issue with the open square (unknown) symbols appearing.  Sadly.

coverage.c
fbreitwieser: Fixed by ensuring that a bin is at minimum one base

coverage.c
fbreitwieser: Outputted the header by deault. Also changed the column names as suggested

coverage.c
fbreitwieser: Fixed w/ @jmarshall 's `\xFF` version

fbreitwieser:
Thanks again for the review everyone. Sorry it took a while, but implemented and tested most of the requested changes. The default histogram mode should now also work on Putty, and a 'full resolution mode' can be invoked by `-U`. Haven't added the man page, yet. Let me know if there are further comments!

fbreitwieser:
Re all eight block characters vs not - Even though the vertical resolution is reduced by a factor of 4, it does not matter too much in my opinion. First plot is with 2 block chars and should work with default putty font, second with 8:

image

coverage.c
```diff
@@ -0,0 +1,632 @@
...
+static const char *const BLOCK_CHARS8[8] = {"\u2581", "\u2582", "\u2583", "\u2584", "\u2585", "\u2586", "\u2587", "\u2588"};
+// In some terminals / with some fonts not all UTF8 block characters are supported (e.g. Putty). Use only half and full block for those
+static const char *const BLOCK_CHARS2[2] = {"\u2584", "\u2588"};
+
+#else
+
+// Fall back to explicit UTF-8 encodings of the same characters
+#define VERTICAL_LINE "\xE2\x94\x82"
+
+static const char *const BLOCK_CHARS8*[8] = {
```
daviesrob: Looks like there's a typo here.  The older version of gcc on travis tries to compile this and says:
```
gcc -g -O2 -I. -Ihtslib -I./lz4  -c -o coverage.o coverage.c
coverage.c:93:38: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘*’ token
 static const char *const BLOCK_CHARS8*[8] = { 
```

fbreitwieser: Thanks, will fix that

fbreitwieser:
> Could the default instead by to use UTF? UTF is pretty widely supported these days. The option can be to request ASCII rather than the default UTF.

I'm open for discussions about this. However, with full UTF, the histogram would not display correctly for all Putty users with default config, which I suppose is still a sizeable portion. Of course I also like the full UTF8 look better :).

sjackman:
Sounds like a Putty bug to me. Here's the page to report bugs to Putty:
https://www.chiark.greenend.org.uk/~sgtatham/putty/feedback.html
PuTTY Known Bugs and Wish List: https://www.chiark.greenend.org.uk/~sgtatham/putty/wishlist/
utf8-default: Make UTF-8 the default line character set: https://www.chiark.greenend.org.uk/~sgtatham/putty/wishlist/utf8-default.html

fbreitwieser:
They say in the feature request that it's fixed, though. I just tried putty via Wine, and with four different fonts (Courier New [default], fixedsys, Andale Mono, PT Mono) it didn't display the UTF characters, but with Menlo it did.

sjackman:
What's Putty's default code page? Is it still ISO-8859-1?

fbreitwieser:
It says it is UTF-8, but I think the issue is the default fonts don't have the required characters

coverage.c
```diff
@@ -0,0 +1,632 @@
...
+#include <ctype.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>  // variadic functions
+#include <limits.h>  // INT_MAX
+#include <math.h>    // round
+#include <stdbool.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
```
valeriuo: `ioctl.h` is not included in the MSYS2 toolchain, so it will not build on Windows.
Appveyor doesn't flag this, because it builds on `develop` for pull requests (there is still some work to be done on the config script).

fbreitwieser: Thanks for catching this. I use it to get terminal width, when the width of the histogram is the whole width. I'll check for alternatives on Windows.

sjackman: The environment variable `COLUMNS`, when it's present, otherwise default to 80.

fbreitwieser: Just tried with `getenv("COLUMNS")`. In principle that would be a great solution, but it seems that I need to call `export COLUMNS` in the shell for the `COLUMNS` to be set in the program. Is there a workaround for that?

sjackman: > The process that is spawned from the shell won't see any variables that are not exported, and at least OSX bash does not export COLUMNS by default.

Huh. You're right. That's odd. I learned something.

```console
$ echo $COLUMNS
80
$ env | grep COLUMNS
$
```

jmarshall: Me too — I always assumed $LINES/$COLUMNS were exported too. Back to `ioctl` with appropriate ifdefs I guess!

fbreitwieser: Just setup a MSYS2 environment to test; for me it compiles and works with the ioctl header. Are the instructions on how to build samtools w/ MSYS2? Also tried GetConsoleScreenBufferInfo w/ MSYS2, but doesn't give me the correct values so far.

I'm not cross-compiling usually, so excuse my ignorance: `_WIN32` is not defined when compiling with gcc in MSYS2. How should I check the system?

valeriuo: I should have added a few more details. MSYS2 comes in two flavours:

- the MSYS2 line, which is a fork of Cygwin, is more POSIX-like and links all programs with a compatibility library (`msys-2.0.dll`). This is what you have probably installed. It usually ships **gcc 7** and the `msys2-runtime-devel` package, which includes `ioctl.h` (`pacman -Fs ioctl.h`).
- the MINGW line provides the `mingw-w64-{i686,x86_64}-toolchain`, which includes **gcc 8** and links with the native Windows libraries. And this is what Appveyor uses. You can see that at line 13 in the build [log](https://ci.appveyor.com/project/valeriuo/samtools-dp6yo/builds/22378373):

`sh -lc "pacman -S --noconfirm --needed base-devel mingw-w64-x86_64-toolchain mingw-w64-x86_64-zlib mingw-w64-x86_64-bzip2 mingw-w64-x86_64-xz mingw-w64-x86_64-curl"`

`GetConsoleScreenBufferInfo` seems to be working only with `cmd` consoles. 

Anyway, the `#include` and the code have to be guarded by `#if defined(__MSYS__) || defined (__CYGWIN__)`. 
`_WIN32` comes with the MINGW toolchain.

fbreitwieser: Thanks a lot for the details @valeriuo ! That `GetConsoleScreenBufferInfo` only works with `cmd` consoles complicates it, but the feature is anyway not the most important - histogram width can still be defined manually. I'll submit a new version.

fbreitwieser: What's the minimum GCC version that should be supported? What do you think of `#if defined(__has_include) && __has_include(<sys/ioctl.h>)` as test? `__has_include` was added in GCC 5.

jmarshall: Using `__has_include` would be trying too hard. (But see the macros in HTSlib's _hts_defs.h_ for similar things that **are** worthwhile.)

You could use `configure` to see if ioctl was available, but that would be trying too hard too.

You mostly just want it to **build** on windows. So `#ifndef __WINDOWS_WHATEVER_` use ioctl `#else` just use `$COLUMNS` or (`#if WINDOWS`) `GetConsoleScreenBufferInfo` or a hardcoded 80.

fbreitwieser: Sounds good. Just pushed a version that checks with `__has_include` if it's available, and check w/ ifdefs otherwise. Can refactor that to not use `__has_include`. In the current version, `getenv("COLUMNS")` is checked first, then ioctl if it has it or windows GetConsoleScreenBufferInfo. Might be better to check for ioctl/GetConsoleScreenBufferInfo first, and do getenv if those are not available.

I also changed the options to have full UTF mode (with eight block characters) and the legacy mode (with two block characters) on a more equal footing: `samtools coverage -m` and `samtools coverage --histogram` show legacy mode, `samtools coverage -M` and `samtools coverage --Histogram` show the full UTF set. Also switched to using full terminal width for the histogram by default. I think that's desirable but am open to discussion.

sjackman: For the option names, I suggest…
```
samtools coverage -m
samtools coverage --histogram
samtools coverage -m --ascii
samtools coverage --histogram --ascii
```

sjackman: An environment variable would be helpful to set the default.
```sh
export SAMTOOLS_ASCII=1
```
in which case there should also be a `--utf8` option to override the environment variable.

daviesrob: I think the `__has_include` is a bit over the top.  Usually for this sort of thing we just check for the normal Windows defines, and otherwise assume the normal unix-y stuff is available.

`TIOCGWINSZ` seems fairly universal, but as ioctls are somewhat system-dependent it might be worth putting in an `#ifdef TIOCGWINSZ` check in to ensure it's really there.

jkbonfield: I'm not sure on using an environment variable to control this behaviour.  I'm not aware of other subcommands where command line options are alternatively given via the environment instead.

Not saying it's necessarily a bad idea, but if we do it I think we should be consistent and think it through rather than winging it for this one sub-command, eg with sensible name spaces such as SAMTOOLS_COVERAGE_OPTS=-m or similar, and permit it to simply be an extension to argc/argv.

sjackman: For reproducibility, I'm against allowing environment variables to affect behaviour other than simple presentation issues. One advantage of `SAMTOOLS_ASCII` over say `SAMTOOLS_COVERAGE_OPTS` is that it could be used by other components of `samtools` other than `coverage` that run into the same UTF/ASCII issue.

coverage.c
```diff
@@ -0,0 +1,671 @@
...
+
+    if (opt_n_bins <= 0 || opt_full_width) {
+        // get number of columns of terminal
+        const char* ccolumns = getenv("COLUMNS");
+        int columns = 0;
+        if (ccolumns == NULL) {
+#if defined has_windows
+            CONSOLE_SCREEN_BUFFER_INFO csbi;
+            GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &csbi);
+	    columns = csbi.srWindow.Right - csbi.srWindow.Left + 1;
```
valeriuo: You still need a default value for columns (80 works fine) for MINGW consoles, as `GetConsoleScreenBufferInfo` doesn't return the correct values in this case and the output looks like

coverage

coverage.c
```diff
@@ -0,0 +1,671 @@
...
+                columns = w.ws_col;
+#endif
+        } else {
+            columns = atoi(ccolumns); // atoi(NULL) returns 0
+        }
+        if (columns > 60) {
+            opt_n_bins = columns - 40;
+        } else {
+            opt_n_bins = 50;
+	}
```
valeriuo: Also, please change the tabs to 4 spaces and tidy up the indentation.

fbreitwieser:
Hi, sorry for the delay, was very busy the last months. Will add the changes and push a cleaned version later today!

coverage.c
fbreitwieser: Checking the return status of `GetConsoleScreenBufferInfo` now, should capture this case.

fbreitwieser: Let me know if there are still isues

coverage.c
```diff
@@ -0,0 +1,672 @@
...
+        {"n-bins", required_argument, NULL, 'w'},
+        {"full-width", required_argument, NULL, 'W'},
+        {"region", required_argument, NULL, 'r'},
+        {"help", no_argument, NULL, 'h'},
+        { NULL, 0, NULL, 0 }
+    };
+
+    // parse the command line
+    int c;
+    while ((c = getopt_long(argc, argv, "o:L:q:Q:hHw:r:b:mM", lopts, NULL)) != -1) {
```
valeriuo: You need to add `A` and `W` to this list. Also, what is `M` for?

coverage.c
```diff
@@ -0,0 +1,672 @@
...
+        if ( read_file_list(opt_file_list, &n_bam_files, &argv) ) {
+            print_error_errno("coverage", "Cannot open file list \"%s\".", opt_file_list);
+            return EXIT_FAILURE;
+        }
+        optind = 0;
+    }
+    else
+        n_bam_files = argc - optind; // the number of BAMs on the command line
+
+    data = (bam_aux_t **)calloc(n_bam_files, sizeof(bam_aux_t*)); // data[i] for the i-th BAM file
```
valeriuo: Check for memory allocation errors.

valeriuo:
@fbreitwieser 
I [rebased](https://github.com/valeriuo/samtools/commits/covhist) your work on top of the latest develop and added a [commit](https://github.com/valeriuo/samtools/commit/60e33ad0434affc4e826236980255ffc9cc6bc4b) with various fixes. If you agree to them, I will merge everything and close this PR.

fbreitwieser:
@valeriuo , thanks so much for doing those changes and fixes! Look good to me! Sorry I haven't been more responsive - started a new job and did not have much time to work on it. Looking forward to having the functionality added to samtools!

valeriuo:
Merged into develop

jmarshall:
Three remaining not-initialised-at-the-top variables produced an astonishing number of warnings: see [e.g. this travis log](https://travis-ci.org/samtools/samtools/jobs/565862331).

I've taken the liberty of pushing a commit fixing them and encapsulating this PR's commits with an explicit merge commit.

daviesrob:
Samtools' travis configuration current lacks some of the extras that htslib has, like `-Werror` and the script to check for trailing white-space.  I'll look into adding them.

PR #1007, opened 2019-02-08
Reheader external

### LINKED ISSUE #999: samtools reheader: modify header with external program

It would be nice if `samtools reheader` would be able to:
  - read files from stdin
  - let the header be processes (modified) by an external program (which accepts the header on stdin and writes the modified header to stdout)
  - write this modified header to stdout (in BAM or CRAM format) before writing the reads of the original BAM/CRAM file 

This would be very helpful in cases where you need to modify the header, when you don't want to write the output of the previous filtering steps to a file.

```bash
samtools view -bu -q 10 sample.bam \
    | samtools reheader -c modify_header.sh - \
    | samtools sort -O BAM > sample_q10.sorted.bam
```

valeriuo:
Allow an external application to modify the header of the target file and pass back the result to _samtools_. The external app will read the original header through **stdin** and output the modified header on **stdout**.

Fixes #999.

### DIFF
--- a/bam_reheader.c
+++ b/bam_reheader.c
@@ -1,7 +1,7 @@
 /*  bam_reheader.c -- reheader subcommand.
 
     Copyright (C) 2010 Broad Institute.
-    Copyright (C) 2012-2015 Genome Research Ltd.
+    Copyright (C) 2012-2019 Genome Research Ltd.
 
     Author: Heng Li <lh3@sanger.ac.uk>
 
@@ -43,7 +43,7 @@ DEALINGS IN THE SOFTWARE.  */
  * the header.    No checks are made to the validity.
  */
 int bam_reheader(BGZF *in, bam_hdr_t *h, int fd,
-                 const char *arg_list, int add_PG)
+                 const char *arg_list, int add_PG, int skip_header)
 {
     BGZF *fp = NULL;
     ssize_t len;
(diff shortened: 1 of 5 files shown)


bam_reheader.c
```diff
@@ -453,30 +467,104 @@ int main_reheader(int argc, char *argv[])
...
         }
     }
 
-    if (argc - optind != 2)
+    if ((argc - optind != 2 || external) && (argc - optind != 1 || !external))
         usage(stderr, 1);
 
-    { // read the header
+    if (external) {
+        in = sam_open(argv[optind], inplace?"r+":"r");
```
jmarshall: If you open the file with `hopen()` followed by `hts_hopen()`, you will already have the `hFILE*` and won't have a desire for an `hts_hfile()` function.

jmarshall:
1. The conventional option letter for stuff like this is `-c` / `--command`. cf bash, su.

2. If you organised this via a function like

```c
bam_hdr_t *rewrite_header(bam_hdr_t *input_header) {
    if (g_external_command_mode) return do_popen(g_command, input_header);
    else return do_sam_hdr_read(g_header_filename);
}
```

that was called from inside each of the `bam/cram_reheader` functions, you wouldn't have any need to rewind the input.

3. It looks like the external program is always given the header in SAM format. IMHO that's the right thing to do, but it should be documented.

valeriuo:
Thanks for the review and suggestions.

Issue #51, opened 2013-05-22
Error message improvement (in general)


jmarshall:
This is a tracking bug for error messages and diagnostics.  Diagnostics need to:
1. Exist where appropriate.  There are a number of places where e.g. an incorrect reference sequence name is silently ignored and you just get no results rather than an indication of detectable user error.
2. Except for warnings, lead to a non-zero exit status.  So failure can be detected in scripts.
3. Contain all relevant information.  For example, currently most errors propagated from system calls omit `perror()` or an equivalent, which makes it harder for the user to track down the problem:
   
   ```
   $ samtools sort example.bam /usr/cantwritehere
   [samopen] SAM header is present: 25 sequences.
   [sort_blocks] fail to create file /usr/cantwritehere.bam.
   ```
   
    Samtools knows that the problem is "Permission denied" and it should be printing that out so that the user knows too.

PR #1013, opened 2019-02-20
Check for gzopen() failure and print error messages

jmarshall:
Ensure that all instances of `gzopen()` have appropriate error checking. In a couple of cases, lift the `gzopen` to above some allocations so that the error path doesn't need to undo the allocations. Down in _misc/*.c_ the infrastructure is lacking and we don't go to great lengths to e.g. return a non-zero status.

Fixes another few instances of #51.

Noticed in bedcov by @rotifergirl (see https://twitter.com/Julie_B92/status/1097867250642628608).


daviesrob:
Thanks.  More work needed here, but it's a start.

Bedcov getting into an infinite loop suggests something lacking in `kseq` too.

jmarshall:
Yes, `kseq` and `kstream` basically have no error checking — IMHO the way to deal with this is to exterminate places where they're used. See also samtools/htslib@71c03b88edf892a93eae7bb7932a6da84858480a's commit message.

I'm sure _bedcov.c_ has better ways to read BED files nowadays, but this was the small quick fix…

PR #1019, opened 2019-03-07
Use the new muti-region generating method in stats

valeriuo:
Fix a memory deallocation conflict introduced by https://github.com/samtools/htslib/pull/836

### DIFF
--- a/stats.c
+++ b/stats.c
@@ -2321,7 +2321,7 @@ int main_stats(int argc, char *argv[])
     char *bam_fname = NULL;
     char *bam_idx_fname = NULL;
     char *group_id = NULL;
-    int sparse = 0, has_index_file = 0;
+    int sparse = 0, has_index_file = 0, ret = 1;
     sam_global_args ga = SAM_GLOBAL_ARGS_INIT;
 
     stats_info_t *info = stats_info_init(argc, argv);
@@ -2409,6 +2409,13 @@ int main_stats(int argc, char *argv[])
         free(info);
         return 1;
     }
+
+    if (has_index_file && !(bam_idx_fname = argv[optind++])) {
+        fprintf(stderr, "No index file provided\n");
+        free(info);
+        return 1;
+    }
+
     if (ga.nthreads > 0)
         hts_set_threads(info->sam, ga.nthreads);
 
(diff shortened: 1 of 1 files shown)


stats.c
```diff
@@ -2426,80 +2433,45 @@ int main_stats(int argc, char *argv[])
...
-                        if ( all_stats->nregions && all_stats->regions ) {
-                            while (sam_itr_multi_next(info->sam, iter, bam_line) >= 0) {
-                               if (info->split_tag) {
-                                   curr_stats = get_curr_split_stats(bam_line, split_hash, info, targets);
-                                   collect_stats(bam_line, curr_stats, read_pairs);
-                               }
-                               collect_stats(bam_line, all_stats, read_pairs);
-                            }
+                if ( all_stats->nregions && all_stats->regions ) {
+                    while (sam_itr_next(info->sam, iter, bam_line) >= 0) {
```
daviesrob: Need to check for return value < -1 here

stats.c
```diff
@@ -2426,80 +2433,45 @@ int main_stats(int argc, char *argv[])
...
                     }
-                } else {
-                    fprintf(stderr, "Creation of the region list failed.");
                 }
-
-                hts_idx_destroy(bam_idx);
+                hts_itr_multi_destroy(iter);
             } else {
-                fprintf(stderr, "Random alignment retrieval only works for indexed BAM files.\n");
+                fprintf(stderr, "Multi-region iterator could not be created\n");
```
daviesrob: Probably want to return an error here

stats.c
```diff
@@ -2426,80 +2433,45 @@ int main_stats(int argc, char *argv[])
...
             }
-
-            bed_destroy(region_hash);
+            hts_idx_destroy(bam_idx);
         } else {
-            fprintf(stderr, "Creation of the region hash table failed.\n");
+            if (has_index_file)
+                fprintf(stderr, "Invalid index file '%s'\n", bam_idx_fname);
+            fprintf(stderr, "Random alignment retrieval only works for indexed files\n");
```
daviesrob: Missing else?
Also need to return an error.

valeriuo: > Missing else?

No, just that, when the user specifies the index file, a hint could be given about the validity of the file.

PR #1020, opened 2019-03-08
Don't break people's names across (source file) lines

jmarshall:
Minor, but some people care about this.

The man pages will be reflowed when they're displayed which might break some names across lines again, but at least this keeps these linguistic units together when the files are being edited.

* @daviesrob appears sometimes as Rob and sometimes as Robert… 🤔 

* The _AUTHORS_ file has not kept up. It might benefit from being updated from _doc/samtools.1_'s AUTHOR section.

* It is observed once again (see also https://github.com/samtools/samtools/pull/894#issuecomment-409490067) that the AUTHOR sections on the individual man pages arguably unfairly de-emphasise those who have contributed significantly and generally to samtools but who have not invented new subcommands.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -103,8 +103,8 @@ Release 1.9 (18th July 2018)
    - Fixed compilation problems when using C_INCLUDE_PATH. (#870; #817 reported
      by Robert Boissy)
 
-   - Fixed --version when built from a Git repository. (#844, thanks to John
-     Marshall)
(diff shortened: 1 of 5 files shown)


daviesrob:
Merged.  Yes, AUTHORS is very outdated and needs quite a few adjustments.  It may be the subject of a future PR.

jmarshall:
Thanks Rob.

PR #1033, opened 2019-04-10
fixed extern declaration

d-smirnov:
samtools did not compile if configured with --without-curses flag


daviesrob:
Looks like this was missed when the index file locations feature was added in 2dd7d3.

Thanks for the patch.

jmarshall:
Has anyone actually tried to use this option with interactive tview? How come `curses_tv_init()` didn't gain an `fn_idx` parameter?

daviesrob:
Indeed, the `-X` option seems somewhat broken in most modes.  It seems there's a lack of `tview` coverage in the test harness, too.

Issue #939, opened 2018-09-03
samtools fastq ignoring -f/-F


finswimmer:
Hello,

I'm running samtools v1.9.

```
$ samtools --version
samtools 1.9
Using htslib 1.9
Copyright (C) 2018 Genome Research Ltd.
```

According to the examples given on the [web manual page](http://www.htslib.org/doc/samtools.html) and the comment about bugs there, I understood that `samtools fastq` extract all alignments nevertheless which FLAG is set. So the number of reads in the final `fastq` file, should be the same number as in the `bam` file. If one would get rid of secondary aligned reads, the option `-F 0x100` have to be set.

My `bam` file was generated by `bwa` and sorted by name:

`$ bwa mem -R '@RG\tID:1\tPL:ILLUMINA\tSM:1' -t 3 -M genome.fa Sample1_L001_R1_001.fastq.gz Sample1_L001_R2_001.fastq.gz | samtools sort -n -o Sample1.bam -`

Number of reads in `Sample1_L001_R1_001.fastq.gz`:

```
$ zcat Sample1_L001_R1_001.fastq.gz|wc -l 
7946560
```

So the total number of reads should be:

`Total_Reads = 2x (7946560/4) = 3973280`

Checking total number of reads in `bam`:
```
$ samtools view -c Sample1.bam
3975850
```

Checking number of reads in `bam` without secondary alignments:
```
$ samtools view -c -F 0x100 Sample1.bam
3973280
```

Now I create `fastq` without any option:

`$ samtools fastq Sample1.bam|bgzip -c > all.fastq.gz`

Checking read numbers:
```
$ zcat all.fastq.gz|wc -l 
15893120
```

So the number of reads is `15893120/4 = 3973280`. So it seems the secondary alignment were removed.

So maybe the default was changed and the manual page wasn't update? To confirm this, I try to convert only secondary alignments to `fastq`:

```
$ samtools fastq -f 0x100 Sample1.bam|bgzip -c > secondary.fastq.gz
[M::bam2fq_mainloop] discarded 0 singletons
[M::bam2fq_mainloop] processed 0 reads
```

No reads were extracted.

At this time it seems to me that `samtools fastq`

- extract unique reads by default
- ignoring the filter FLAGs

Is it a bug? Is the manual not correct? Or do I misunderstood something?

fin swimmer

daviesrob:
[...]

PR #1042, opened 2019-04-24
Samtools fastq -o option and other fixes.

jkbonfield:
- -1 FILE -2 FILE with the same filename now works properly instead of clobbering each other.

- -o FILE is now a synonym for -1 FILE -2 FILE to avoid needing > FILE (with -s and -0 to junk places).

- Stdout/stderr usage message improvements

- When reading from redirected file or pipe, "-" is no longer explicitly needed as input filename.  (This problem still affects many other samtools commands.)

- Finally for my own sanity, we no longer need to remember the fasta / fastq commands are hidden along side view.  It has its own file.


bam_fastq.c
```diff
@@ -0,0 +1,1019 @@
...
+    }
+
+    int i, j;
+    for (i = 0; i < 3; ++i) {
+        if (opts->fnr[i]) {
+	    for (j = 0; j < i; j++)
+		if (opts->fnr[j] && strcmp(opts->fnr[j], opts->fnr[i]) == 0)
+		    break;
+	    if (j == i) {
+		state->fpr[i] = open_fqfile(opts->fnr[i], state->compression_level, &state->p);
```
valeriuo: Please replace the tabs with spaces.

valeriuo:
Fixes #939

bam_fastq.c
jkbonfield: Good spot.  Fixed.

It seems this is a result of upgrading my OS causing emacs on the old servers to no longer work properly (it dislikes something in my upgraded config apparently).  Grr.

jkbonfield:
While tab hunting I also spotted a recent tab introduction (c15e884e) in bam_tview_curses.c.  While there I noticed the horrid mishmash of indentation and bracket styles too. 

It's not really part of this issue, but for laziness given the simplicity of the changes I added it to this PR.

bam_fastq.c
```diff
@@ -0,0 +1,1019 @@
...
+        in_isnt_tty = 0; // as we specified "-" explicitly anyway
+
+    if (argc - optind + in_isnt_tty == 0) {
+        fprintf(stderr, "No input file specified.\n");
+        bam2fq_usage(in_isnt_tty ? stderr : stdout, argv[0]);
+        free_opts(opts);
+        return false;
+    }
+
+    if (argc - optind + in_isnt_tty > 1) {
```
valeriuo: This condition doesn't work with MSYS2 consoles, which don't support `isatty`. Maybe we can drop it for `_WIN32` environments. Also, `fasta` and `fastq` should probably be added to [this list](https://github.com/samtools/samtools/blob/b217a91c501a2026e753f0b6c75456d246ae340f/test/test.pl#L885).

`	next if ($^O =~ /^msys/ && $subcommand =~ /^(dict|sort|stats|view)$/);`

jkbonfield: Updated, can you please test whether the MSYS console "make check" is now passing? Thanks.

I decided my code was a misfeature.  It was trying to be clever, but we don't do that elsewhere anyway.  Eg "cat file1.sam | samtools view file2.sam" doesn't whinge, it just views file2.sam and ignores the input.  This now does the same, instead of grumbling about too many arguments.

bam_fastq.c
```diff
@@ -0,0 +1,1020 @@
...
+        free_opts(opts);
+        return false;
+    }
+
+    int in_isnt_tty = !isatty(STDIN_FILENO);
+    if (in_isnt_tty && argc > optind && strcmp(argv[optind], "-") == 0)
+        in_isnt_tty = 0; // as we specified "-" explicitly anyway
+
+    if (argc - optind + in_isnt_tty == 0) {
+        fprintf(stderr, "No input file specified.\n");
```
jmarshall: `samtools fastq` by itself is the canonical way to ask for the usage, so it shouldn't produce a _No input file specified_ error message (or return EXIT_FAILURE).

daviesrob: The _No input file specified_ predated this PR, but it's now gone.  It should also now return EXIT_SUCCESS if just called as `samtools fastq` (but maybe not on Windows...)

I've also fixed a couple of dependencies in the Makefile that got lost when the fastq code was extracted from `sam_view.c`.

jmarshall: Thanks × 2. Oops — it appeared on the GH web view to be new, but that's just because James moved the code to a new file, duh.

PR #1049, opened 2019-04-30
Use the temporary files interface with sort

### LINKED ISSUE #1035: samtools sort  fails if there are existing tmp files
Hi,

It appears that `samtools sort` fails if there existing `tmp.*` files in the directory (from a previous failed run, for example) with the following message:

```
samtools sort: failed to create temporary file "/lab/.../bwa/347.4______347.4_RG1.bam.tmp.0000.bam": File exists
```

Is there a way to specify force overwrite existing tmp files?

```
samtools 1.9
Using htslib 1.9
```

valeriuo:
The temporary files will be automatically deleted after `samtools sort` ends, even in case of a crash.
Name conflict resolution is provided by the temporary file creation logic.
Fixes #1035

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,13 +230,13 @@ check test: samtools $(BGZIP) $(TEST_PROGRAMS)
 
 
 test/merge/test_bam_translate: test/merge/test_bam_translate.o test/test.o libst.a $(HTSLIB)
-	$(CC) $(ALL_LDFLAGS) -o $@ test/merge/test_bam_translate.o test/test.o libst.a $(HTSLIB_LIB) $(ALL_LIBS) -lpthread
+	$(CC) $(ALL_LDFLAGS) -o $@ test/merge/test_bam_translate.o test/test.o libst.a tmp_file.o $(LZ4OBJS) $(HTSLIB_LIB) $(ALL_LIBS) -lpthread
 
 test/merge/test_rtrans_build: test/merge/test_rtrans_build.o libst.a $(HTSLIB)
-	$(CC) $(ALL_LDFLAGS) -o $@ test/merge/test_rtrans_build.o libst.a $(HTSLIB_LIB) $(ALL_LIBS) -lpthread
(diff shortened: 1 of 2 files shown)


raivivek:
@valeriuo Wondering if there's been any discussion on getting this merged? Thanks!

Makefile
```diff
@@ -230,13 +230,13 @@ check test: samtools $(BGZIP) $(TEST_PROGRAMS)
 
 
 test/merge/test_bam_translate: test/merge/test_bam_translate.o test/test.o libst.a $(HTSLIB)
-	$(CC) $(ALL_LDFLAGS) -o $@ test/merge/test_bam_translate.o test/test.o libst.a $(HTSLIB_LIB) $(ALL_LIBS) -lpthread
+	$(CC) $(ALL_LDFLAGS) -o $@ test/merge/test_bam_translate.o test/test.o libst.a tmp_file.o $(LZ4OBJS) $(HTSLIB_LIB) $(ALL_LIBS) -lpthread
```
jmarshall, line 233: Instead of adding these here, just move them from `AOBJS` to `LIBST_OBJS` — sharing object files between samtools and these test programs is exactly what _libst.a_ is for.

valeriuo:
@raivivek It's highly unlikely the solution will be merged in this form. Although it ticks most of the boxes in terms of functionality, the current implementation of temporary file back-end uses LZ4 compression, which creates much larger files than what we currently generate with BGZF. This is not acceptable, so we will have to change the way the temporary files are compressed.

ghuls:
@valeriuo Are there plans to use zstd compression instead (to have better speed and BGZF and smaller filesize than L

daviesrob:
I have tried using zstd.  I does a reasonable job of compressing the files but suffers from the same problem as the existing solution, which is that it becomes stalled on I/O when the number of temporary files becomes too large.

I have an ongoing project to rewrite sort so that it runs more efficiently.  Meanwhile I'll close this PR, as unfortunately this experiment didn't result in the hoped-for improvement on the existing version.

raivivek:
Thanks @daviesrob. I appreciate the continued push on this front.

PR #1055, opened 2019-05-28
Expand the man page's pileup format documentation

jmarshall:
Rewrite the single paragraph summary as a longer list of columns describing the format more fully, and add descriptions of the optional extra columns.

Grouping by sample (SM) identifiers was only ever for VCF/BCF output. Remove SM info from the first paragraph: text pileup output is only grouped by input file.

### DIFF
--- a/doc/samtools-mpileup.1
+++ b/doc/samtools-mpileup.1
@@ -63,38 +63,17 @@ samtools mpileup
 
 .SH DESCRIPTION
 .PP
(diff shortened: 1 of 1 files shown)


daviesrob:
Looks good, thanks.

jmarshall:
Re grouping by sample: For the record, see also [this 2013 samtools-help posting](https://sourceforge.net/p/samtools/mailman/message/31720086/) from Heng:

> Only the BCF output groups samples. The text output does not.

Issue #1047, opened 2019-04-29
samtools fastq -i does not produce Casava format


dinvlad:
#### Are you using the latest version of [samtools]
`samtools 1.9`

#### Please describe your environment.
- OS: `Linux 4.14.94+`
- machine architecture: `x86_64`
- compiler: `gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516`

#### Please specify the steps taken to generate the issue, the command you are running and the relevant output.
According to [Samtools doc](http://www.htslib.org/doc/samtools.html), `samtools fastq -1 f1.fastq -2 f2.fastq -i in.bam` (note the `-i`) should
> add Illumina Casava 1.8 format entry to header (eg 1:N:0:ATCACG)

However, all reads in the resulting FASTQ files _don't_ contain `1:N:0:<BARCODE>` tag. This is despite the fact that `samtools view -H` is in the form
```
@HD     VN:1.6  SO:queryname
@RG     ID:A00XX.1      SM:Sample1   LB:Sample1   PL:illumina     PU:A00XXX:NNN:HH1VVYYZZ.1.GGCCTTAA-ATCGATCG     CN:BI
```
All reads in the resulting FASTQs look like
```
@A00XXX:NNN:HH1VVYYZZ:1:1101:10000:10000
...
```
while they should look like
```
@A00XXX:NNN:HH1VVYYZZ:1:1101:10000:10000 1:N:0:GGCCTTAAATCGATCG 
...
```

daviesrob:
To get the `-i` option to work, you need to supply both the `--index-format` and `--i1` options (and maybe also `--i2` if you used dual indexing).  So something like this should work (`-n` is also used here to get rid of the `/1` and `/2` on the read names):

```
samtools fastq -1 name_R1.fastq.gz -2 name_R2.fastq.gz --i1 name_I1.fastq.gz --index-format "i*" -n -i name.bam
```

Arguably having to supply `--i1` shouldn't be strictly necessary with `-i`, but `fastq` will currently insist on having it.  If you don't actually want the barcode fastq file, you can always work around this by using `--i1 /dev/null`.

dinvlad:
[...]

PR #1059, opened 2019-06-05
Bug fix to samtools fastq

jenniferliddle:
Allow santools fastq to have the -i and --index-format options without
having to specify index files.

Fix bug whereby only the second index of a dual index tag was
being put into the -i (Casava) field.

This fixes the problems mentioned in #1047


daviesrob:
I'm not sure if this works properly yet.  `samtools fastq -i --index-format 'i4i4' ...` doesn't write out the extra tags, but `samtools fastq -i -i1 index1.fq --index-format 'i4i4' ...` does.  Also `samtools fastq -i ...` without `--index-format` does not complain about the lack of that option while it will if `-i1` is used.

When I do get them, the tags look like `1:N:0:ATCA-CGAT` on both reads, but [the bcl2fastq user manual](https://support.illumina.com/content/dam/illumina-support/documents/documentation/software_documentation/bcl2fastq/bcl2fastq2-v2-20-software-guide-15051736-03.pdf) says it should start with `2:` on read 2.  Unfortunately it doesn't say exactly what it does with dual-index runs, so I can't tell if the barcode bit is correct.  Do we have the output from bcl2fastq on one of these runs to we can see what it actually does?

Also (and somewhat more fundamentally) it looks like `bcl2fastq` might be using the expected barcode sequence from the manifest instead of the one that was actually read.  If that is true then using the value in the BC:Z tag from the read would not always give the right answer, and instead the BC: value stored in the header for the read group should be used instead (if available).

jenniferliddle:
> I'm not sure if this works properly yet. `samtools fastq -i --index-format 'i4i4' ...` doesn't write out the extra tags, but `samtools fastq -i -i1 index1.fq --index-format 'i4i4' ...` does.

So it doesn't. Now fixed.

> Also `samtools fastq -i ...` without `--index-format` does not complain about the lack of that option while it will if `-i1` is used.

Added a check to complain about this.

> When I do get them, the tags look like `1:N:0:ATCA-CGAT` on both reads, but [the bcl2fastq user manual](https://support.illumina.com/content/dam/illumina-support/documents/documentation/software_documentation/bcl2fastq/bcl2fastq2-v2-20-software-guide-15051736-03.pdf) says it should start with `2:` on read 2.

So it does! It was always supposed to use 1: or 2: This would appear to be a bug which went in with a PR from June of last year, but nobody noticed it. Now fixed.

> Also (and somewhat more fundamentally) it looks like `bcl2fastq` might be using the expected barcode sequence from the manifest instead of the one that was actually read. If that is true then using the value in the BC:Z tag from the read would not always give the right answer, and instead the BC: value stored in the header for the read group should be used instead (if available).

The BC:Z tag from the read is the *actual* barcode found, rather than the expected one, and therefore, I would argue, is the "right" one. Also, as far as I can see, we are not putting a BC value in the read group line of the header.

PR #1060, opened 2019-06-06
Allow some header reordering in tests

daviesrob:
Needed so that samtools tests will continue to work through the HTSlib header API merge.

Some fixes to test files, e.g. spaces instead of tabs in addrprg tests.

Change how test_count_rg and test_filter_header redirect stderr so messages from e.g. address sanitizer are less likely to be lost.  Has to be done by flushing stderr then switching the underlying file descriptor as MinGW doesn't allow assigning to stderr, which makes closing it and reopening with fdopen() difficult.


test/split/test_filter_header_rg.c
```diff
@@ -28,6 +28,59 @@ DEALINGS IN THE SOFTWARE.  */
...
+    for (l = hdr1 + nl1; *l != '\0'; l += strcspn(l, "\n")) ++l, ++count1;
+    for (l = hdr2 + nl2; *l != '\0'; l += strcspn(l, "\n")) ++l, ++count2;
+    if (count1 != count2) return false;
+    
+    lines1 = malloc(count1 * sizeof(*lines1));
+    if (!lines1) return false;
+    lines2 = malloc(count2 * sizeof(*lines2));
+    if (!lines2) return false;
+
+    for (i = 0, l = hdr1 + nl1; *l != '\0'; l += strcspn(l, "\n"))
```
valeriuo: Are you skipping the first line intentionally? 
Currently, `lines1[0]` starts with the second line (`@SQ..`) and `lines1[2]` is an empty string.

To include the first line, we could have something like:
```
    for (i = 0, l = hdr1; *l != '\0'; l += strcspn(l, "\n")+1)
        lines1[i++] = l;
    for (i = 0, l = hdr2; *l != '\0'; l += strcspn(l, "\n")+1)
        lines2[i++] = l;
```
Otherwise, all the tests pass with the new header API and I happy to merge it.

test/split/test_filter_header_rg.c
```diff
@@ -28,6 +28,59 @@ DEALINGS IN THE SOFTWARE.  */
...
+    const char *l;
+    const char **lines1, **lines2;
+    int res = 0;
+
+    // First line should be @HD
+    if (strncmp(hdr1, "@HD\t", 4) != 0) return false;
+    if (strncmp(hdr2, "@HD\t", 4) != 0) return false;
+    nl1 = strcspn(hdr1, "\n");
+    nl2 = strcspn(hdr2, "\n");
+    if (nl1 != nl2 || memcmp(hdr1, hdr2, nl1) != 0) return false;
```
valeriuo: Ok, I see now that the first lines are compared here.

valeriuo:
Merged as cb011eba7e7f0311ad3f3a6736d1fb0946c201cd

PR #1061, opened 2019-06-07
fix typo

russHyde:
"supplemenary alignments" -> "supplementary alignments" in `samtools markdup --help`


daviesrob:
Thanks for the fix.

Issue #774, opened 2018-02-08
tools that write BAM/CRAM should have option to index


keiranmraine:
Tools that (can) write a sorted alignment file should have the ability to generate the index file while writing the output.

It would be preferable to have this than use `tee` to achieve this via pipes.

Reason: biobambam2 has this for most tools but the build time is quite considerable, samtools now has most of the functionality I need (for downstream processing) so I'm looking at reducing complexity of tools in docker images.

daviesrob:
Agreed, it would be useful.  It would need changes to HTSlib though, as I don't think it supports indexing on-the-fly at the moment.

keiranmraine:
Alternatively the ability to `--passthrough` the stream on tools (on view useful to do this even when writing to file), e.g.

```
samtools view -S -C -T ref.fa --passthrough -o new.cram some.sam | \
  samtools index --passthrough - new.cram.crai | \
    md5sum > new.cram.md5
```

(I know `-S` is ignored, just for extra clarity)

daviesrob:
It should be possible to add a passthrough mode to `samtools index`, although we would effectively be implementing tee in it.  It would be easier than adding indexing to everything, which although useful would require ABI changes.

For this particular case you could try [teepot](https://github.com/wtsi-npg/teepot) which is mostly like tee, but better at handling cases where one of the output pipes either pauses for a long time or stops reading completely.

```
samtools view -S -C -T ref.fa some.sam | \
  teepot new.cram >(samtools index - new.cram.crai) >(md5sum)
```

keiranmraine:
Ah, I see this works without any changes... thank you for the (re)education

jkbonfield:
Depends on samtools/htslib#718, but will also need a small samtools tweak to use the new htslib code.

nh13:
@jkbonfield bump

jkbonfield:
Good point.  The htslib PR to do this was written a year ago but only arrived in Feb so I kind of forgot that there was samtools follow-up work waiting on it.  I'll try and remind myself what the grand plan was!

jkbonfield:
It's a work in progress now so expect PR soon.  I decided to use the --long-opts global options style as we need indices specifying for every program that outputs.

[...]

PR #1062, opened 2019-06-07
Auto index

jkbonfield:
1. Fixed the travis build so it behaves as claimed in the comment - an *optimised* build doing address sanitizer. 

2. Added a global --write-index option to enable the auto-indexing code.  This requires the auto_index PR of htslib merging first.

### DIFF
--- a/bam_addrprg.c
+++ b/bam_addrprg.c
@@ -223,7 +223,7 @@ static void usage(FILE *fp)
             "  -r STRING @RG line text\n"
             "  -R STRING ID of @RG line in existing header to use\n"
             );
-    sam_global_opt_help(fp, "..O..@");
+    sam_global_opt_help(fp, "..O..@.");
 }
 
 static bool parse_args(int argc, char** argv, parsed_opts_t** opts)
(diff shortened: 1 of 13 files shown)


jkbonfield:
NB: fixes #774

jkbonfield:
Note I restarted tests and Travis now passes thanks to the htslib merge, but AppVeyor dies:

> The command failed [256]: C:/projects/samtools/samtools view -h --write-index -O sam,level=5 -o C:/projects/samtools/test/dat/auto_indexed.sam C:/projects/samtools/test/dat/mpileup.1.sam
> out:
> err:samtools auto_index: failed to open index "C:/projects/samtools/test/dat/auto_indexed.sam.csi" for writing

This error comes from sam_utils.c due to `sam_idx_init` returning < 0.    I'm wondering if it's something odd with C: in the filename, but I'm curious how it passes htslib, given that is the bit that does the writing of the index, but not samtools.

Furthermore it passes AppVeyor when run from my own fork: https://ci.appveyor.com/project/jkbonfield/samtools/builds/25278919

I'll restart it incase something broken in the merge, but this has me stumped atm.

daviesrob:
Added support for `##idx##` delimiter that came in with samtools/htslib#861, but see also samtools/htslib#884 which is needed to prevent samtools from incorrectly outputting sam all the time when this is used.

bam_addrprg.c
```diff
@@ -438,12 +438,17 @@ static bool init(const parsed_opts_t* opts, state_t** state_out) {
...
 
-static bool readgroupise(state_t* state)
+static bool readgroupise(parsed_opts_t *opts, state_t* state)
 {
     if (sam_hdr_write(state->output_file, state->output_header) != 0) {
         print_error_errno("addreplacerg", "[%s] Could not write header to output file", __func__);
         return false;
     }
+    char *idx_fn = NULL;
+    if (opts->ga.write_index && opts->output_name && strcmp(opts->output_name, "-") != 0) {
```
valeriuo: `if (opts->ga.write_index)` should be enough, as `auto_index` checks internally for a null output file name and for `stdout`.

padding.c
```diff
@@ -597,10 +597,23 @@ int main_pad2unpad(int argc, char *argv[])
         ret = 1;
         goto depad_end;
     }
+    if (ga.write_index) {
```
valeriuo: Here, only the flag is checked, as it probably should be.

valeriuo:
Rebased and pushed to develop

PR #1070, opened 2019-07-02
Use `#` symbol to mark deletions on the reverse strand

### LINKED ISSUE #1054: Feature request: In mpileup output, differentiating between deletions on forward and reverse strands
In the `samtools mpileup` output, deletions which stretch out over multiple bases are marked with asterisks (`*`), regardless of whether they are on forward or reverse strand.

**Problem:** This makes it hard to count the strand-specific read depth (including deletions) from the mpileup output. (Especially when the `-l` and/or `-r` command line option is specified and the output may start in the middle of a deletion so that the first line of the deletion which would contain the deleted bases is not included in the output.)

Example output from command `samtools mpileup -l somefile.bed`:

```
chr1	10440	N	10	**c$***AcCa	HFBBBDDIA:
```
Here the line contains five asterisks, with no distinction whether they are on forward or reverse strand.

**Proposed solution:** It would be useful if there would be a way to see whether the deletion is on forward or reverse strand, for example by outputting different character depending on whether it is on forward or reverse strand. (Since this affects the output, perhaps there should be a separate command line option to enable this behavior.)

Hypothetical output from `samtools mpileup -l somefile.bed --enable-proposed-feature`:

```
chr1	10440	N	10	##c$***AcCa	HFBBBDDIA:
```
Here the deletions on the reverse strand were marked with hypothetical `#` and deletions from the forward strand with the usual `*`.

valeriuo:
Fixes #1054


jmarshall:
Lots of programs parse mpileup output. Did you consider making this optional?

valeriuo:
Yes, I thought about adding a long option, if preserving the current output convention is a must.

doc/samtools-mpileup.1
```diff
@@ -129,7 +129,7 @@ Forward	Reverse	Meaning
 \&.\fR dot	,\fR comma	Base matches the reference base
 ACGTN	acgtn	Base is a mismatch to the reference base
 >	<	Reference skip (due to CIGAR \(lqN\(rq)
-*	*	Deletion of the reference base (CIGAR \(lqD\(rq)
+*	*(#)	Deletion of the reference base (CIGAR \(lqD\(rq)
```
jmarshall: Thanks. A couple of comments on the man page:

Where options affect the pileup output, I put a pointer to them in the pileup description — something like (don't just apply this as the `.TE` will get duplicated, but you get the idea :smile:)

```suggestion
*      *\fR/\fB#       Deletion of the reference base (CIGAR \(lqD\(rq)
 .TE

Deleted bases are shown as \(lq*\(rq on both strands
unless \fB--reverse-del\fR is used.

```

doc/samtools-mpileup.1
```diff
@@ -257,6 +257,12 @@ Disable read-pair overlap detection.
 Include customized index file as a part of arugments. See
 .B EXAMPLES
 section for sample of useage.
+.TP
+.B --reverse-del
+Mark the deletions on the reverse strand with the character
+.B #
```
jmarshall: Use `.BR # ,` to avoid the space between `#` and `,`, and similarly for the fullstop after `the usual *`.

This should probably be listed under _Output Options_ (and similarly in the Usage string).

valeriuo:
Thanks. Amended.

bam_plcmd.c
```diff
@@ -949,6 +949,7 @@ static void print_usage(FILE *fp, const mplp_conf_t *mplp)
 "  -O, --output-BP         output base positions on reads\n"
 "  -s, --output-MQ         output mapping quality\n"
 "      --output-QNAME      output read names\n"
+"      --reverse-del       use '#' character to mark deletions on the reverse strand"
```
jmarshall: `\n` and shorten to fit in 80 characters:

```suggestion
"      --reverse-del       use '#' character for deletions on the reverse strand\n"
```

doc/samtools-mpileup.1
```diff
@@ -142,6 +142,10 @@ reference bases represented similarly.  (Subsequent pileup lines will
 contain \(lq*\(rq for this read indicating the deleted bases.)
 .IP \(bu 2
 If this is the last position covered by the read, a \(lq$\(rq character.
+.IP \(bu 2
```
jmarshall: Each of these subbullet items is a character(s) that appears in the output. So this would be better beside the table it relates to (and without its own bullet) rather than lost down the bottom of the list of items.

Issue #1064, opened 2019-06-10
Samtools to output RX tags for Unique Molecular Indexes (UMI)


apallav:
Hi,

I would like to know if there is a way in samtools algo to recognize and output UMI / uniq molecular index tags from bam files. ex:RX:Z tags. 

Could this be added as feature request if none is available currently?

Thanks,

Aparna

valeriuo:
By default, `samtools view` outputs all tags attached to a read, starting from column 12.

There is a merged but unreleased pull request (#1001) that filters the reads based on tags, including **RX**. However, the filter only works with a supplied tag value, e.g. 
`samtools view -d RX:AGTCGATA alignment.bam`

But it's not very clear that this feature is what you are after. If you just want to list all the values of a tag present in a file, you can go with a grep filter: 
`samtools view alignment.bam | grep -oE "RX:Z:[acgtnACGTN-]+ | sort -u"`

apallav:
> By default, `samtools view` outputs all tags attached to a read, starting from column 12.
> 
> There is a merged but unreleased pull request (#1001) that filters the reads based on tags, including **RX**. However, the filter only works with a supplied tag value, e.g.
> `samtools view -d RX:AGTCGATA alignment.bam`
> 
> But it's not very clear that this feature is what you are after. If you just want to list all the values of a tag present in a file, you can go with a grep filter:
> `samtools view alignment.bam | grep -oE "RX:Z:[acgtnACGTN-]+ | sort -u"`

I am sorry for not being clear. I would like to see if it is possible for samtools mpileup to give out tag information  just like it gives out Fastq header tags. 

I could use samtools view alignment.bam | grep -oE "RX:Z:[acgtnACGTN-]+ | sort -u" to pipe in to samtools mpileup but that info is lost in the mpile up non vcf or vcf output formats. 

I specifically work with non-vcf formats from samtools / mpileup.

Thanks for your help,

-A

valeriuo:
Closed by https://github.com/samtools/samtools/pull/1073

PR #1073, opened 2019-07-08
[mpileup] Fields and tag values

### LINKED ISSUE #1052: mpileup output option to report flags of reads
I would like to use duplicate reads to call variants, but I want to be able to filter to require at least N non-duplicate supporting alt-allele reads.  To implement this filtering I was attempting to use mpileup to determine how many alt-allele reads were non-duplicates, but to do that I need the DUP flag for each read, which is not available in the mpileup output.

I'd like an option in the samtools mpileup command to report the FLAGS for each read.

valeriuo:
Add option `--output-extra` to extend the output of mpileup with columns of user selected read fields and tag values. The field and tag names have to be provided as a comma-separated all caps string argument, following the option. The filed names follow the naming convention from the [Sequence Alignment/Map Format Specification](https://github.com/samtools/hts-specs/blob/master/SAMv1.pdf) (page 6). Field value columns are always displayed before tag value columns.

Usage example:

`samtools mpileup --output-extra FLAG,RG,QNAME,NM in.bam`

Includes #1070 
Fixes #1052, #1054, #1064


daviesrob:
I think it would be useful to have an option to set the separator, so people who want to dump out tags like 'SA' have a way of dealing with the presence of commas in the tag itself.

Currently columns in empty pileup lines get printed as '*', but missing tags in otherwise present lines get printed as an empty string.  It might be better to print them as '*' as well.  And possible this should be settable too, in case someone wants to print a tag that's likely to have a single '*' in it.

We should also remove the restriction in the man page on tag names being all caps, as lower-case tags are allowed.  I don't think there's any actual restriction on this in the code.

jmarshall:
It was not commented on by reviewers at the time, but this PR changed the expected output for test _a17.out_, which happens to test the combination of extra columns from `mpileup -sO`. The change reordered the two extra columns.

What was the reason for this change? Note that _a17.out_ no longer matches the mpileup format documentation (previously merged in PR #1055), and see also [this report of samtools 1.9 vs 1.10 incompatibility](https://twitter.com/CFlensburg/status/1278857379400966145).

PR #1076, opened 2019-07-12
add another argument to function sam_header2key_val to solve a bug might be triggered

vanNul:
In the original code of sam_header2key_val, when the last line of bam header parsed, it return NULL to its caller, if the tag we want to get is just in the last line, and we use a while loop to iterate through sam_header2key_val and test its return value as whether this while loop will continue or not, then in the while loop, we will never got the last line record checked.

In my version, I add another argument eof to sam_header2key_val, which indicate the eof of bam header, and use its value as another sentinel of the while loop, then we can simply use just one while loop to iterate all lines in bam header now.

I have modified all the example and source code using this modification, I think it is ugly somehow, but haven't got a better idea to use just one while loop only to iterate all lines and got all wanted key-value checked in the while loop.

### DIFF
--- a/sam_header.c
+++ b/sam_header.c
@@ -722,7 +722,7 @@ void *sam_header2key_val(void *iter, const char type[2], const char key_tag[2], 
         *_key = key->value;
         *_value = value->value;
         if(!l->next){
-            eof = true;
+            *eof = true;
         }
         return l->next;
     }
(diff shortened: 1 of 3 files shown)


valeriuo:
This code was not used and will not be in the future. HTSlib will ship a new [header API](https://github.com/samtools/htslib/pull/812) in the next release and `sam_header.{h,c}` will be dropped altogether.

In the meantime, feel free to implement the functionality of `init_group_id` with the new header methods, which have been merged into the develop branch of HTSlib.

vanNul:
Wow, that is wonderful, glad to hear that

PR #1084, opened 2019-08-09
Fixed invalid assertion.

jkbonfield:
b may not be set to NULL if the "b_score > score[rp]" condition later on is not met.  I don't understand what this is attempting to do, but given b isn't always NULL we shouldn't assert that it is.  Possibly we need another fix?

Local example file for @jenniferliddle at `~jkb/lustre/minhash/NA12878.chr10.10M.cram`

NB: this file isn't sorted by name, but actually I don't care either as I'm not splitting into separate files and I'm not processing anything in pairs.  I just want to output sequence in original orientation (for which samtools fastq should work amiably).  So this is input that's not expected by the tool, but obviously it shouldn't crash.  (Samtools 1.9 doesn't so this is new functionality that appeared in 0f8febc1)

### DIFF
--- a/bam_fastq.c
+++ b/bam_fastq.c
@@ -913,9 +913,9 @@ static bool bam2fq_mainloop(bam2fq_state_t *state, bam2fq_opts_t* opts)
     bam1_t* b = NULL;
 
     while (true) {
-        assert(b == NULL);
-        b = bam_init1();
-        if (b == NULL ) {
+        if (!b)
+            b = bam_init1();
+        if (b == NULL) {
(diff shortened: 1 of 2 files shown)


valeriuo:
Lines [931-935](https://github.com/samtools/samtools/blob/76e7ed6e5ebd51c7f1f8cee88f4c2371afc2f6b0/bam_fastq.c#L931-L935) could also be changed from 

```c
if (!at_eof && filter_it_out(b, state)) {
    bam_destroy1(b);
    b = NULL;
    continue;
}
```
to
```c
if (!at_eof && filter_it_out(b, state)) continue;
```
as `b` can now be reused.

PR #998, opened 2019-01-15
Use the new header API

valeriuo:
Depends on htslib PR https://github.com/samtools/htslib/pull/812
Fixes #856
Fixes #1079

PR #1087, opened 2019-08-16
Add a new `@PG` line when producing a new SAM/BAM/CRAM file

### LINKED ISSUE #1086: PG header records should be added for all samtools commands
Samtools should record in a PG header record (what and) that it is manipulating the SAM/BAM/CRAM data it outputs so that the provenance of data is better recorded in the PG records.

samtools to add a PG record by default whenever it outputs SAM/BAM/CRAM.
This could be enhanced with a command-line option to disable added the PG record.

valeriuo:
The following commands: `sort`, `merge`, `fixmate`, `collate`, `depad`, `split`, `reheader`, `cat`, `calmd`, `phase`, `addreplacerg` add by default a new `@PG` line to the header of the produced SAM/BAM/CRAM file. The new `@PG` line includes the tags `ID`, `PN`, `PP`, `VN` and/or `CL`.
For all these commands, this behaviour can be suppressed using the long option `--no-PG`.

Includes #998
Fixes #1086

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -239,11 +239,11 @@ check test: samtools $(BGZIP) $(TEST_PROGRAMS)
 test/merge/test_bam_translate: test/merge/test_bam_translate.o test/test.o libst.a $(HTSLIB)
 	$(CC) $(ALL_LDFLAGS) -o $@ test/merge/test_bam_translate.o test/test.o libst.a $(HTSLIB_LIB) $(ALL_LIBS) -lpthread
 
-test/merge/test_rtrans_build: test/merge/test_rtrans_build.o libst.a $(HTSLIB)
-	$(CC) $(ALL_LDFLAGS) -o $@ test/merge/test_rtrans_build.o libst.a $(HTSLIB_LIB) $(ALL_LIBS) -lpthread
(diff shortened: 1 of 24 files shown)


test/merge/test_rtrans_build.c
```diff
@@ -26,6 +26,11 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include "../../bam_sort.c"
 
+// For tests, just return a constant that can be embedded in expected output.
+const char *samtools_version(void) {
```
jmarshall: Better to adjust the _Makefile_ instead: add _test/test.o_ to the link recipes for _test/merge/test_rtrans_build_ and _test/merge/test_trans_tbl_init_ — similarly to the existing _test/merge/test_bam_translate_ recipe.

valeriuo: Thanks. Done.

daviesrob:
Can you add the `--no-PG` option to the various man pages as well, please?

daviesrob:
Squashed, rebased and removed `htslib/cram.h` includes which are no longer needed following samtools/htslib#914

PR #1089, opened 2019-08-20
Make get_rg_sample() use header API and fix tview -s option

daviesrob:
Makes `samtools tview -s` work as advertised in the manual page.

It's been broken since commit c8c745e which deliberately commented out the filter due to changes resulting from the HTSlib/samtools split.  Some later commits tried to restore it, but didn't actually work properly.  This gets it back to the pre-c8c745e state.

### DIFF
--- a/bam_tview.c
+++ b/bam_tview.c
@@ -34,27 +34,66 @@ DEALINGS IN THE SOFTWARE.  */
 #include "samtools.h"
 #include "sam_opts.h"
 
-khash_t(kh_rg)* get_rg_sample(const char* header, const char* sample)
+static void destroy_rg_hash(khash_t(kh_rg)* rg_hash)
 {
-    khash_t(kh_rg)* rg_hash = kh_init(kh_rg);
-    // given sample id return all the RD ID's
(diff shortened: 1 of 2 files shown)


bam_tview.c
```diff
@@ -34,27 +34,65 @@ DEALINGS IN THE SOFTWARE.  */
...
+        if (strcmp(sample, id_val.s) != 0) {
+            // No match, try SM:
+            r = sam_hdr_find_tag_pos(header, "RG", i, "SM", &sm_val);
+            if (r < -1) goto memfail;
+            if (r < 0 || strcmp(sample, sm_val.s) != 0)
+                continue;
+        }
+
+        // Found a match, add ID to rg_hash
+        kh_put(kh_rg, rg_hash, ks_release(&id_val), &r);
```
valeriuo: There is a small probability for a memory leak, in case `kh_put` fails before adding the new ID string to the hash table. `id_val` has already been released, so the pointer is lost.

daviesrob:
Rebased with fix for possible memory leak.

valeriuo:
Merged into develop

PR #1091, opened 2019-08-20
Additions to markdup

whitwham:
Added optional optical duplicate marking.
Additional stats and the option to output to a file.
Write an PG line to the header.
Optional different duplicate criteria.
Other small bug fixes and tweaks.

### DIFF
--- a/bam_markdup.c
+++ b/bam_markdup.c
@@ -77,6 +77,11 @@ typedef struct {
     key_data_t single_key;
 } read_queue_t;
 
+typedef struct {
+    char *name;
+    char type;
+} dup_map_t;
+
 
 
 static khint32_t do_hash(unsigned char *key, khint32_t len);
@@ -143,7 +148,7 @@ static int key_equal(key_data_t a, key_data_t b) {
 
 KHASH_INIT(reads, key_data_t, in_hash_t, 1, hash_key, key_equal) // read map hash
 KLIST_INIT(read_queue, read_queue_t, __free_queue_element) // the reads buffer
-KHASH_MAP_INIT_STR(duplicates, int) // map of duplicates for supplementary dup id
+KHASH_MAP_INIT_STR(duplicates, dup_map_t) // map of duplicates for supplementary dup id
 
 
 /* Calculate the mate's unclipped start based on position and cigar string from MC tag. */
@@ -458,7 +463,7 @@ static void make_single_key(key_data_t *key, bam1_t *bam) {
 
 /* Add the duplicate name to a hash if it does not exist. */
 
-static int add_duplicate(khash_t(duplicates) *d_hash, bam1_t *dupe) {
+static int add_duplicate(khash_t(duplicates) *d_hash, bam1_t *dupe, char *orig_name, char type) {
     khiter_t d;
     int ret;
 
@@ -467,10 +472,19 @@ static int add_duplicate(khash_t(duplicates) *d_hash, bam1_t *dupe) {
     if (d == kh_end(d_hash)) {
         d = kh_put(duplicates, d_hash, strdup(bam_get_qname(dupe)), &ret);
 
-        if (ret > 0) {
-            kh_value(d_hash, d) = 1;
-        } else if (ret == 0) {
-            kh_value(d_hash, d)++;
+        if (ret >= 0) {
+            if (orig_name) {
+                kh_value(d_hash, d).name = strdup(orig_name);
+
+                if (kh_value(d_hash, d).name == NULL) {
+                    fprintf(stderr, "[markdup] error: unable to allocate memory for duplicate original name.\n");
+                    return 1;
+                }
+            } else {
+                kh_value(d_hash, d).name = NULL;
+            }
+
+            kh_value(d_hash, d).type = type;
         } else {
             fprintf(stderr, "[markdup] error: unable to store supplementary duplicates.\n");
             return 1;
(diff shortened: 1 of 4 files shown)


bam_markdup.c
```diff
@@ -948,7 +948,6 @@ static int bam_mark_duplicates(md_param_t *param) {
                 "EXAMINED: %d\n"
                 "PAIRED: %d\n"
                 "SINGLE: %d\n"
-                "DULPICATE PAIR: %d\n"
```
tseemann: it looks like the line has been removed, but not added back with the correct spelling?

whitwham: Oh, for crying out loud!  I even had someone check it for me.  Thank you.

bam_markdup.c
```diff
@@ -914,22 +1286,51 @@ static int bam_mark_duplicates(samFile *in, samFile *out, char *prefix, int remo
...
+        fprintf(fp,
+                "COMMAND: %s\n"
                 "READ: %d\n"
                 "WRITTEN: %d\n"
                 "EXCLUDED: %d\n"
                 "EXAMINED: %d\n"
                 "PAIRED: %d\n"
                 "SINGLE: %d\n"
-                "DULPICATE PAIR: %d\n"
+                "DUPLICATE PAIR: %d\n"
```
whitwham: @tseemann there it is.

bam_markdup.c
```diff
@@ -467,6 +627,200 @@ static int add_duplicate(khash_t(duplicates) *d_hash, bam1_t *dupe) {
...
+    int oxpos = 0, oypos = 0, dxpos = 0, dypos = 0;
+
+
+    original  = bam_get_qname(ori);
+    duplicate = bam_get_qname(dup);
+
+    seps = get_coordinate_positions(original, &oxpos, &oypos);
+
+    if (!(seps == 3 || seps == 4)) {
+        fprintf(stderr, "[markdup] warning: cannot decipher read name %s for optical duplicate marking.\n", original);
```
daviesrob: Might want to limit the number of these that get printed out, or it could get very spammy.

bam_markdup.c
```diff
@@ -467,6 +627,200 @@ static int add_duplicate(khash_t(duplicates) *d_hash, bam1_t *dupe) {
...
+}
+
+
+/* Modified Lander-Waterman equation */
+static inline double lw_equation(double x, double c, double n) {
+    return c / x - 1 + exp(-n / x);
+}
+
+
+/* estimate the library size */
```
daviesrob: As this is essentially the Picard code reimplemented in C, we should put in a comment to say where it came from, and also add them to the copyright at the top of the file.  Something like:

```
Estimate library size derived from Picard DuplicationMetrics.java
  Copyright (c) 2009,2018 The Broad Institute.  MIT license.
```

Luckily they use the same license as us.

bam_markdup.c
```diff
@@ -33,15 +33,38 @@ DEALINGS IN THE SOFTWARE
 #include <ctype.h>
 #include <time.h>
 #include <sys/stat.h>
+#include <math.h>
 #include "htslib/thread_pool.h"
 #include "htslib/sam.h"
 #include "sam_opts.h"
 #include "samtools.h"
+#include "htslib/cram.h"
```
daviesrob: This can be removed as `stringify_argv()` is now in `htslib/sam.h`

bam_markdup.c
whitwham: I disagree.  It should either keep printing out error messages or fail completely.  I don't want it to not work silently.

I should probably mention the optical duplicates depends on the standard read names in the man page.

bam_markdup.c
whitwham: True, though I'm not sure how useful the library estimate is.

bam_markdup.c
whitwham: Okay.

bam_markdup.c
daviesrob: That's a harsh punishment for someone who runs it with the wrong name format.  I presume you get one warning per alignment?

I'd have a static variable to count the number of warnings, and stop printing them after a while.

whitwham: If someone is running with the wrong name format and trying to get optical duplicates then it should probably not work at all.  No point in doing a full run if it is not going to work.

Alternatively, we could not throw an error while it is running and come up with a final total of errors at the end.  This could be useful for people using a mixture of data from different machines.  The downside is that is does not pinpoint what the problem read names are.

jmarshall: I think I did something like that for the bwa check-qnames-match thing, but it must have been an early draft as I can't find it in the source code now. Something like

```c
for (all reads) {
    …
    if (strcmp(read_qname, mate_qname) != 0) {
        nerrors++;
        if (nerrors <= 10) fprintf(stderr, "names don't match %s != %s\n", read_qname, mate_qname);
    }
    …
}

if (nerrors > 10) fprintf(stderr, "And additional name mismatch errors: %llu more errors\n", nerrors - 10);
```

whitwham: I think something like that could work.  It gives examples of failures, a final report on numbers and doesn't produce huge amounts of spam.

bam_markdup.c
tseemann: Great :)

bam_markdup.c
```diff
@@ -467,6 +637,238 @@ static int add_duplicate(khash_t(duplicates) *d_hash, bam1_t *dupe) {
...
+    read_pairs /= 2;
+    duplicate_pairs /= 2;
+
+    if (read_pairs && duplicate_pairs) {
+        unsigned long unique_pairs = read_pairs - duplicate_pairs;
+        double m = 1;
+        double M = 100;
+        int i;
+
+        if (lw_equation(m * (double)unique_pairs, (double)unique_pairs, (double)read_pairs) < 0) {
```
valeriuo: If unique_pairs is 0, this will result in a division by 0.

bam_markdup.c
```diff
@@ -418,6 +456,128 @@ static int make_pair_key(key_data_t *key, bam1_t *bam) {
...
+    char *cig;
+
+    this_ref    = bam->core.tid + 1; // avoid a 0 being put into the hash
+    other_ref   = bam->core.mtid + 1;
+
+    this_coord = unclipped_start(bam);
+    this_end   = unclipped_end(bam);
+
+    if ((data = bam_aux_get(bam, "MC"))) {
+        cig = bam_aux2Z(data);
```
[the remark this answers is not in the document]
whitwham: Good catch.

bam_markdup.c
```diff
@@ -418,6 +456,128 @@ static int make_pair_key(key_data_t *key, bam1_t *bam) {
...
+
+    if (leftmost < 0) {
+        leftmost = 1;
+    } else if (leftmost > 0) {
+        leftmost = 0;
+    } else {
+        // tie breaks
+
+        if (bam->core.pos == bam->core.mpos) {
+            if ( bam->core.flag & BAM_FREAD1) {
```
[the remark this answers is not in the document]
whitwham: Seems perfectly valid according to **The Samtools Developers' Coding Style Guide** but I'll change it so it looks neater.

bam_markdup.c
```diff
@@ -453,10 +614,19 @@ static int add_duplicate(khash_t(duplicates) *d_hash, bam1_t *dupe) {
     if (d == kh_end(d_hash)) {
         d = kh_put(duplicates, d_hash, strdup(bam_get_qname(dupe)), &ret);
 
-        if (ret > 0) {
-            kh_value(d_hash, d) = 1;
-        } else if (ret == 0) {
-            kh_value(d_hash, d)++;
+        if (ret >= 0) {
+            if (orig_name) {
+                kh_value(d_hash, d).name = strdup(orig_name);
```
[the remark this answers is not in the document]
whitwham: Done.

bam_markdup.c
whitwham: Well unique_pairs can only be zero if all the pairs are duplicates.  I'm not sure that can ever happen, even if markdup is fed bad data.  Still, I'll check for it just in case.

daviesrob:
Rebased with tweaks to ensure all commits build without warnings.

bam_markdup.c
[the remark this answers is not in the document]
valeriuo: The same check should be employed for `make_pair_key` at line 360.

PR #1092, opened 2019-08-21
Fix use of internal headers + travis update to spot this

jkbonfield:
We spot dependence on htslib source rather than htslib install by not letting samtools automatically compile htslib itself.

Note there is some nuance with LDFLAGS shenanigans where we cannot override it as expected due to make -e, so the rpath hackery is duplicated (but harmless).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,8 +72,8 @@ script:
     autoreconf -i && \
     (cd $HTSDIR && \
       autoreconf -i && \
-      ./configure --prefix=`pwd`/../inst LDFLAGS="$LDFLAGS -Wl,-rpath,`pwd`/../inst/lib" && \
-      make install \
+      ./configure --prefix=`pwd`/../inst && \
+      make -e install \
     ) && \
     (./configure --enable-werror --with-htslib=`pwd`/inst  LDFLAGS="$LDFLAGS -Wl,-rpath,`pwd`/inst/lib" || \
       (cat config.log; /bin/false) \
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,14 +50,6 @@ addons:
     - liblzma-dev
     - libbz2-dev
 
-# For MacOSX systems
-before_install:
-  - |
-    if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      HOMEBREW_NO_AUTO_UPDATE=1 brew install xz || \
-      ( brew update && brew install xz )
-    fi
-
 before_script:
   # Clone htslib, trying the same branch name in the owners' copy of htslib.
   # If this exists then the user is likely making a joint PR to htslib and
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,11 @@ addons:
 
 # For MacOSX systems
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then HOMEBREW_NO_AUTO_UPDATE=1 brew install xz || ( brew update && brew install xz ); fi
+  - |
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -41,7 +41,7 @@ matrix:
 
 env:
   global:
-    - HTSDIR=./htslib
+    - HTSDIR=./hidden-htslib
```
daviesrob: If you made this `../hidden-htslib` you wouldn't have to change `check_copyright.pl` or `check_spaces.pl`.

.travis.yml
```diff
@@ -63,4 +67,19 @@ before_script:
   # Logic for choosing which to use is in the .travis/clone script.
   - .travis/clone "git://github.com/$(dirname $TRAVIS_REPO_SLUG)/htslib.git" "$HTSDIR" "$TRAVIS_BRANCH"
 
-script: autoreconf -i && (cd htslib && autoreconf -i) && (./configure --enable-werror || (cat config.log; /bin/false)) && if test "x$DO_MAINTAINER_CHECKS" = "xyes" ; then make -e maintainer-check ; fi && make -e && make -e test
+script:
+  - |
+    autoreconf -i && \
+    (cd $HTSDIR && \
+      autoreconf -i && \
+      ./configure --prefix=`pwd`/../inst LDFLAGS="$LDFLAGS -Wl,-rpath,`pwd`/../inst/lib" && \
```
daviesrob: I don't think you need to set rpath for HTSlib.

.travis.yml
```diff
@@ -8,7 +8,7 @@ matrix:
     - os: linux
       compiler: clang
       sudo: required
-      env: CFLAGS="-g -Wall -O3 -fsanitize=address" LDFLAGS="-fsanitize=address"
+      env: CFLAGS="-g -Wall -O3 -fsanitize=address" LDFLAGS="-fsanitize=address -Wl,-rpath,`pwd`/inst/lib"
```
daviesrob: Do you need to set rpath here as it's also being done on the configure line?

.travis.yml
jkbonfield: True, but it doesn't matter either way frankly.  The change isn't exactly an onerous one and I don't like the idea of trying to escape out of the directory structure provided for me by travis.

.travis.yml
jkbonfield: I do if we want to do a make check inside htslib too, although I since removed that change.  I can cull this again and see if it still works - ought to.

.travis.yml
jkbonfield: Yes, it's needed due to `make -e` and I can't get rid of that because we want it for CFLAGS.  I can't get rid of the rpath stuff in the configure line as we need it for the non-clang builds.  It's an annoying duplication, but this is a working kludge.

.travis.yml
```diff
@@ -52,7 +52,11 @@ addons:
 
 # For MacOSX systems
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then HOMEBREW_NO_AUTO_UPDATE=1 brew install xz || ( brew update && brew install xz ); fi
+  - |
+    if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      HOMEBREW_NO_AUTO_UPDATE=1 brew install xz || \
+      ( brew update && brew install xz )
+    fi
```
jmarshall: Rather than reformatting this bit, you could just nuke the whole `before_install` stanza as brewed xz is no longer necessary on macOS, configure or no configure.

jkbonfield: I'll add an additional commit to nuke it, so we can see if travis still works.  I assume they're reasonably up to date.

PR #1093, opened 2019-08-21
Added the ability to sort unmapped data by minhash.

jkbonfield:
This has the effect of clumping similar sequences together, improving
the efficiency of LZ based compression methods.  This is extremely
simplistic with a single kmer and no assembly of minhashes into
contigs.  Hence it's not going to achieve ratios like Minicom (which
uses sketches), but it is fast, uses minimal memory and is achievable
as part of "samtools sort".

Credit for ideas also go to Tony Cox from Illumina.

Some stats on SRR445724.fastq (as unaligned SAM):

```
            bpb         ctime   dtime
Spring      0.4041      1986    374     (tested on seq3, -q bsc -i)
CRAM        1.9494         ?      ?     (default CRAM)
CRAM-M20    0.6306       148     26     (dtime from view -c -@8)
"+Z7s100k   0.5055       677     27     (archive mode for V3.0)
v31-small   0.5017      +498     46     (scramble -V3.1 -Xsmall -s100000 -Z -7 -t8)
```

Total file sizes:
```
-rw-r--r-- 1 jkb team117 3214122305 Aug 14 15:27 /tmp/_1.cram (orig)
-rw-r--r-- 1 jkb team117 2611100075 Aug 14 15:30 /tmp/_2.cram (minhash20)
-rw-r--r-- 1 jkb team117 2404005175 Aug 14 15:44 /tmp/_3.cram ("+Z7s100k)
-rw-r--r-- 1 jkb team117 2090644964 Aug 14 16:21 /tmp/_4.cram (V3.1 small)
-rw-r--r-- 1 jkb team117 2004008960 Aug 14 16:52 SRR445724.spring
```
Not super compression unless we engage slow modes, but a big win for
not much CPU expenditure.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -38,6 +38,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <getopt.h>
 #include <assert.h>
 #include <pthread.h>
+#include <inttypes.h>
 #include "htslib/ksort.h"
 #include "htslib/hts_os.h"
 #include "htslib/khash.h"
@@ -98,6 +99,7 @@ KLIST_INIT(hdrln, char*, hdrln_free_char)
 
 static int g_is_by_qname = 0;
 static int g_is_by_tag = 0;
+static int g_is_by_minhash = 0;
 static char g_sort_tag[2] = {0,0};
 
 static int strnum_cmp(const char *_a, const char *_b)
@@ -134,8 +136,11 @@ typedef struct {
 } heap1_t;
 
 static inline int bam1_cmp_by_tag(const bam1_tag a, const bam1_tag b);
+static inline int bam1_cmp_by_minhash(const bam1_tag a, const bam1_tag b);
 
 // Function to compare reads in the heap and determine which one is < the other
+// Note, unlike the bam1_cmp_by_X functions which return <0, 0, >0 this
+// is strictly 0 or 1 only.
 static inline int heap_lt(const heap1_t a, const heap1_t b)
 {
     if (!a.entry.bam_record)
@@ -146,6 +151,9 @@ static inline int heap_lt(const heap1_t a, const heap1_t b)
     if (g_is_by_tag) {
         int t;
         t = bam1_cmp_by_tag(a.entry, b.entry);
+        if (t != 0) return t > 0;
+    } else if (g_is_by_minhash) {
+        int t = bam1_cmp_by_minhash(a.entry, b.entry);
         if (t != 0) return t > 0;
     } else if (g_is_by_qname) {
         int t, fa, fb;
@@ -1699,6 +1707,12 @@ static int bam_merge_simple(int by_qname, char *sort_tag, const char *out,
     ks_heapmake(heap, heap_size, heap);
     while (heap->pos != HEAP_EMPTY) {
         bam1_t *b = heap->entry.bam_record;
+        if (g_is_by_minhash && b->core.tid == -1) {
+            // Remove the cached minhash value
+            b->core.pos = -1;
+            b->core.mpos = -1;
+            b->core.isize = 0;
+        }
         if (sam_write1(fpout, hout, b) < 0) {
             print_error_errno(cmd, "failed writing to \"%s\"", out);
             goto fail;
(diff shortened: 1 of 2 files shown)


bam_sort.c
```diff
@@ -1803,23 +1842,26 @@ typedef struct {
...
         goto fail;
     }
-    if (sam_hdr_write(fp, (sam_hdr_t *)h) != 0) goto fail;
+    if (sam_hdr_write(fp, h) != 0) goto fail;
     if (n_threads > 1) hts_set_threads(fp, n_threads);
     for (i = 0; i < l; ++i) {
-        if (sam_write1(fp, (sam_hdr_t *)h, buf[i].bam_record) < 0) goto fail;
+        bam1_t *b = buf[i].bam_record;
+        if (clear_minhash && g_is_by_minhash && b->core.tid == -1)
+            b->core.pos = b->core.isize = -1;
```
jmarshall: Why isn't this `b->core.pos = -1; b->core.isize = 0;`?

(In fact, what's this for? Simply to improve compression by flattening some noise? To undo the “unmapped read gets its mate's tid&pos” thing? It wouldn't work for the latter I think. A comment might be handy…)

Also seems like it'd be clearer for the `write_buffer` parameter to be `clear_unmapped`, the test here to be `if (clear_unmapped && b->core.tid == -1)` (so no `g_foo` used in this function) and the single non-zero argument to `write_buffer()` being adjusted accordingly — i.e., lift this `clear_minhash && g_is_by_minhash` logic out of this buffer writing function.

doc/samtools-sort.1
```diff
@@ -109,6 +112,17 @@ suffix.
 To prevent sort from creating a huge number of temporary files, it enforces a
 minimum value of 1M for this setting.
 .TP
+.B "-M "
+Sort unmapped reads (those in chromosome "*") by their sequence
+minimiser (Schleimer et al., 2003; Roberts et al., 2004), also reverse
+complementing as appropriate.  This has the effect of collating some
+similar data together, improving the compressability of the unmapped
```
jmarshall: compressibility

bam_sort.c
jkbonfield: You're correct, it probably should be `isize = 0`.  I was getting confused with the counting from 0 vs 1 and applying to both incorrectly.

It's there because I've modified them elsewhere to hold the two halves of the minimiser.  This is just putting them back to what they would have been, although leaving this data intact is something we wish to discuss at the file formats conference call.

As for the parameter name, it does exactly what it says on the tin and IMO `clear_unmapped` is less clear.

jmarshall: > It's there because I've modified them elsewhere to hold the two halves of the minimiser.

Well, now that I know what's in the tin, this is clearer! I only skimmed the core minimiser code, so a comment here would be appreciated.

I still think lifting `clear_minhash && g_is_by_minhash` out of the function to the one significant call site would be clearer (and remove two whole instructions from the inner loop).

jkbonfield: I'd like to hope the compiler is capable of removing the loop invariants. :-)

Agreed though.  I've adjusted it and will push a new version as soon as I've done some trial runs.  Thanks for the review.

jkbonfield:
Rebased on develop and updated.  I can squash it later, but left as-is for now so the changes from review are visible.

jkbonfield:
Try 2, this time with correct branch copy: rebased off develop and updated with review fixes (thanks).

dkj:
What should be in the `@HD` header record?
- `SO` field of `unknown`, `unsorted`, or even "pseudo" `coordinate`?
- `GO` field of `none` (or "pseudo" `reference`?) would be important if `SO:unknown` is used as I suspect many tools will presume that to be template collated - so info explicitly stating that this is not the case would be valuable.

jkbonfield:
It already set SO to coordinate, so it wouldn't be giving tools any hints that the data is name collated (which it isn't).

I decided to keep SO as coordinate.  It's correct for the general purpose case of aligned + unaligned.  For purely unaligned it's neither correct nor incorrect - just maybe misleading (but that was already true even without adding -M opt).

For subsort I chose minhash.  I don't know what it should be, but there's nothing appropriate, it's free for us to choose, and naming it after the option seems the best choice.

I haven't done anything with GO and it's a controlled vocabulary so I cannot do this.  I'm not sure if we even permit both SO and GO together unless they're clearly duplicates.  Eg SO:queryname and GO:name.  (I've no idea why the vocabulary differs between SO and GO, but that dates all the way back to SAM pre-TeXifcation so it's lost in the mists of time.)

bam_sort.c
```diff
@@ -2012,18 +2064,215 @@ static int ks_radixsort(size_t n, bam1_tag *buf, const sam_hdr_t *h)
...
+}
+
+/*!
+ * @abstract Modifies a single base in the bam structure.
+ *
+ * @param s    Seq buffer, from bam_get_seq(bam1_t*)
+ * @param i    Position in sequence, starting from zero
+ * @param b    Base in nt16 nomenclature (see seq_nt16_table)
+ */
+#define bam_set_seqi(s,i,b) ((s)[(i)>>1] = ((s)[(i)>>1] & (0xf0 >> ((~(i)&1)<<2))) | ((b)<<((~(i)&1)<<2)))
```
daviesrob: If this is added to a future HTSlib, we might want to use this name.  So possibly better to use `set_seqi` here to avoid future clashes?

bam_sort.c
```diff
@@ -2012,18 +2064,215 @@ static int ks_radixsort(size_t n, bam1_tag *buf, const sam_hdr_t *h)
...
+//--- Start of candidates to punt to htslib
+/*!
+ * @abstract
+ * Extracts the sequence (in current alignment orientation) from
+ * a bam record and places it in buf, which is nul terminated.
+ *
+ * @param b     The bam structure
+ * @param buf   A buffer at least b->core.l_qseq+1 bytes long
+ */
+void bam_to_seq(bam1_t *b, char *buf) {
```
daviesrob: Should be `static void ...` (also other functions below).

bam_sort.c
jkbonfield: I moved it to htslib then rather than rename it, as it's a genuinely useful "function".

Personally I think coding style consistency matters, so it's matching the rest in being a #define.  IMO should stay so until one of us gets time to rewrite all those macros as static inlines, which can be another PR.

bam_sort.c
jkbonfield: Done.

In a second commit for sake of consistency I've updated all the other samtools functions (of which there are *MANY*) given this is now our new coding style.  However it may cause problems for people who manipulate internals of samtools via their own nefarious means.  Can't have it both ways!

daviesrob:
045112d should be its own pull request.

PR #1094, opened 2019-08-23
Fix gcc 8.1.0 sprintf overflow warning

daviesrob:
Warning came when compiling with `gcc -g -gdwarf-2 -O3` using gcc version 8.1.0.  It's fairly easy to silence by adjusting the sizes of the buffers involved.

Also ensure that `readable_bps()` can't go past the end of its `units` array.  While unlikely on normal data, it might be possible to get this to happen when the 64-bit position changes have gone into HTSlib.


valeriuo:
Merged into develop

PR #1095, opened 2019-08-26
Account for RG, PG, or CO not being present in merged headers

mp15:
Given that there is a subset of files in existence that do not contain at least 1 of either RG, PG or CO we should handle this.


daviesrob:
Adjusted to use new `ks_c_str()` function, for a slightly simpler fix.

PR #1097, opened 2019-08-27
Add PG lines with view

valeriuo:
Add a `@PG` line when producing a new SAM/BAM/CRAM file with `samtools view`. 
Avoid this behaviour with the `--no-PG` option.

### DIFF
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -349,6 +349,9 @@ of input.
 Include customized index file as a part of arugments. See
 .B EXAMPLES
 section for sample of useage.
+.TP
+.B --no-PG
+Do not add a @PG line to the header of the output file.
 
 .SH EXAMPLES
 .IP o 2
--- a/sam_view.c
+++ b/sam_view.c
@@ -261,14 +261,14 @@ static inline int check_sam_write1(samFile *fp, const sam_hdr_t *h, const bam1_t
 
 int main_samview(int argc, char *argv[])
 {
-    int c, is_header = 0, is_header_only = 0, ret = 0, compress_level = -1, is_count = 0, has_index_file = 0;
+    int c, is_header = 0, is_header_only = 0, ret = 0, compress_level = -1, is_count = 0, has_index_file = 0, no_pg = 0;
     int64_t count = 0;
     samFile *in = 0, *out = 0, *un_out=0;
     FILE *fp_out = NULL;
     sam_hdr_t *header = NULL;
     char out_mode[5], out_un_mode[5], *out_format = "";
     char *fn_in = 0, *fn_idx_in = 0, *fn_out = 0, *fn_list = 0, *q, *fn_un_out = 0;
-    char *fn_out_idx = NULL, *fn_un_out_idx = NULL;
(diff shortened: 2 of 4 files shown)


test/test.pl
```diff
@@ -1589,7 +1589,7 @@ sub bam_compare
 
     my $b1 = open_bgunzip($opts, $bam1);
     my $b2 = open_bgunzip($opts, $bam2);
-    do {
+    do {{
```
jmarshall: ?

doc/samtools-view.1
```diff
@@ -349,6 +349,9 @@ of input.
 Include customized index file as a part of arugments. See
 .B EXAMPLES
 section for sample of useage.
+.TP
+.BI --no-PG
```
jmarshall: `.BI` is for when there's an option and an argument: `.BI --boldme italicme`. There's only bold here, so `.B --no-PG`.

sam_view.c
```diff
@@ -34,6 +34,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <inttypes.h>
 #include <getopt.h>
 #include "htslib/sam.h"
+#include "htslib/cram.h"
```
jmarshall: Needed for…?

sam_view.c
```diff
@@ -261,14 +262,14 @@ static inline int check_sam_write1(samFile *fp, const sam_hdr_t *h, const bam1_t
 
 int main_samview(int argc, char *argv[])
 {
-    int c, is_header = 0, is_header_only = 0, ret = 0, compress_level = -1, is_count = 0, has_index_file = 0;
+    int c, is_header = 0, is_header_only = 0, ret = 0, compress_level = -1, is_count = 0, has_index_file = 0, no_pg = 0;
```
jmarshall: The code for all these --no-PG options would be clearer if the variable was `add_pg` starting at 1 and reset to 0 by the option.

sam_view.c
```diff
@@ -293,6 +294,7 @@ int main_samview(int argc, char *argv[])
 
     static const struct option lopts[] = {
         SAM_OPT_GLOBAL_OPTIONS('-', 0, 'O', 0, 'T', '@'),
+        {"no-PG", no_argument, NULL, 1},
```
jmarshall: It would be good to have a `--no-pg` alias as well for typing ease.

In fact, I wonder whether it would be practical to move this into `sam_global_args` rather than duplicating the code everywhere…

sam_view.c
valeriuo: No longer needed. Will remove.

test/test.pl
valeriuo: For using `last` in a `do-while` statement.

daviesrob: This doesn't actually work as intended because the `last` is simply ignored.  Of course it eventually gets to the end of the file, but it takes longer than it really needs to.  This can be shown by feeding the function different never-ending streams, like `<(yes)` and `/dev/zero`, which put it into an endless loop.

The best solution is to add `&& !$fail` to the while clause.

valeriuo: Yes, that is a better solution.

daviesrob:
Rebased with minor fix to bam_compare().

PR #1100, opened 2019-09-03
Various minor tidy ups

jmarshall:
* Fix a few dependencies corresponding to recent `#include` changes
* Mention `--output-extra` columns in the Pileup Format man page
(and a couple of formatting/typo fixes)
* Rename some recently-added test data files so they are git-ignored and `make testclean`ed


daviesrob:
Thanks very much.

Issue #1077, opened 2019-07-21
flagstat machine readable output format


SichongP:
#### Is your feature request related to a problem? Please specify.
I'm trying to consolidate flagstat output for multiple bam files in my workflow and generate a report with a clean table for all stats. I could achieve this with a simple script and regex but obviously this is prone to failing should there be any updates in the output formatting for flagstat in the future. As such I'm wondering if there is a way to make flagstat output a more standardized and machine readable format (a csv or yaml perhaps)?
If not, may I request such a feature? 
#### Describe the solution you would like.
An option for flagstat when specified output will be in a csv or yaml format that can be easily consolidated.

tseemann:
This would be useful.  I would choose TSV over CSV, and JSON over YAML, but I am all for this.

daviesrob:
Well, it's output hasn't changed for years (four to be precise) but it wouldn't be hard to make it print other formats.

I too prefer JSON to YAML.

raivivek:
If no one's working on this feature yet, I'd like to take a shot at it. I'd also find it personally useful. Let me know if that's alright.

daviesrob:
@raivivek Feel free to have a go, it shouldn't be that difficult.

I think you'd want to use `--output-fmt` for the long option, with maybe `-O` (capital O) for the short option.  So you would need to change `SAM_OPT_GLOBAL_OPTIONS` to:
```c
SAM_OPT_GLOBAL_OPTIONS('-', 0, 'O', '-', '-', '@')
```
and also add `case 'O':` to the options switch statement to handle the format selection.  It's important to do that because `parse_sam_global_opt()` doesn't understand formats like `json`, so you need to bypass it for this option.

It would be best to move the `printf` statements out of `bam_flagstat()` so you can make a separate function for each output format.  You could also turn them into `fprintf` so you can have a `-o` or `--output` option to set the output file name.

Note that samtools code uses four spaces for indentation.  Tabs and trailing spaces are not allowed.  Also, don't forget to add tests and update the `samtools-flagstat.1` manual page.

raivivek:
Great, thanks for the details. That gives me a good start.

raivivek:
@daviesrob Also, I was wondering. Do you have any specific debugging practices that you use? I'm using the usual "change -> make -> run/test", but can't seem to find any general development instructions. Thanks!

raivivek:
[...]

PR #1106, opened 2019-09-11
Add TSV/JSON output format for `samtools flagstat`

raivivek:
Work in progress for #1077. Tests, if any, to be added.

### DIFF
--- a/bam_stat.c
+++ b/bam_stat.c
@@ -131,60 +131,72 @@ static void out_fmt_default(bam_flagstat_t *s)
 static void out_fmt_json(bam_flagstat_t *s) {
     char b0[16], b1[16];
     printf("{\n \"QC-passed reads\": { \n"
-                 "  \"total\": \"%lld\", \n"
-                 "  \"secondary\": \"%lld\", \n"
-                 "  \"supplementary\": \"%lld\", \n"
-                 "  \"duplicates\": \"%lld\", \n"
-                 "  \"mapped\": \"%lld (%s)\", \n"
-                 "  \"paired in sequencing\": \"%lld\", \n"
-                 "  \"read1\": \"%lld\", \n"
-                 "  \"read2\": \"%lld\", \n"
-                 "  \"properly paired\": \"%lld\", \n"
-                 "  \"with itself and mate mapped\": \"%lld\", \n"
-                 "  \"singletons\": \"%lld (%s)\", \n"
-                 "  \"with mate mapped to a different chr\": \"%lld\", \n"
-                 "  \"with mate mapped to a different chr (mapQ >= 5)\": \"%lld\" \n"
+                 "  \"total\": %lld, \n"
+                 "  \"secondary\": %lld, \n"
+                 "  \"supplementary\": %lld, \n"
+                 "  \"duplicates\": %lld, \n"
+                 "  \"mapped\": %lld, \n"
+                 "  \"mapped %%\": %s, \n"
+                 "  \"paired in sequencing\": %lld, \n"
+                 "  \"read1\": %lld, \n"
+                 "  \"read2\": %lld, \n"
+                 "  \"properly paired\": %lld, \n"
+                 "  \"properly paired %%\": %s, \n"
+                 "  \"with itself and mate mapped\": %lld, \n"
+                 "  \"singletons\": %lld, \n"
+                 "  \"singletons %%\": %s, \n"
+                 "  \"with mate mapped to a different chr\": %lld, \n"
+                 "  \"with mate mapped to a different chr (mapQ >= 5)\": %lld \n"
                  " },"
             "\n \"QC-failed reads\": { \n"
-                 "  \"total\": \"%lld\", \n"
-                 "  \"secondary\": \"%lld\", \n"
-                 "  \"supplementary\": \"%lld\", \n"
-                 "  \"duplicates\": \"%lld\", \n"
-                 "  \"mapped\": \"%lld (%s)\", \n"
-                 "  \"paired in sequencing\": \"%lld\", \n"
-                 "  \"read1\": \"%lld\", \n"
-                 "  \"read2\": \"%lld\", \n"
-                 "  \"properly paired\": \"%lld\", \n"
-                 "  \"with itself and mate mapped\": \"%lld\", \n"
-                 "  \"singletons\": \"%lld (%s)\", \n"
-                 "  \"with mate mapped to a different chr\": \"%lld\", \n"
-                 "  \"with mate mapped to a different chr (mapQ >= 5)\": \"%lld\" \n"
+                 "  \"total\": %lld, \n"
+                 "  \"secondary\": %lld, \n"
+                 "  \"supplementary\": %lld, \n"
+                 "  \"duplicates\": %lld, \n"
+                 "  \"mapped\": %lld, \n"
+                 "  \"mapped %%\": %s, \n"
+                 "  \"paired in sequencing\": %lld, \n"
+                 "  \"read1\": %lld, \n"
+                 "  \"read2\": %lld, \n"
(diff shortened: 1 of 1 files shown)


bam_stat.c
```diff
@@ -97,15 +97,130 @@ static void usage_exit(FILE *fp, int exit_status)
...
+    printf("%lld + %lld properly paired (%s : %s)\n", s->n_pair_good[0], s->n_pair_good[1], percent(b0, s->n_pair_good[0], s->n_pair_all[0]), percent(b1, s->n_pair_good[1], s->n_pair_all[1]));
+    printf("%lld + %lld with itself and mate mapped\n", s->n_pair_map[0], s->n_pair_map[1]);
+    printf("%lld + %lld singletons (%s : %s)\n", s->n_sgltn[0], s->n_sgltn[1], percent(b0, s->n_sgltn[0], s->n_pair_all[0]), percent(b1, s->n_sgltn[1], s->n_pair_all[1]));
+    printf("%lld + %lld with mate mapped to a different chr\n", s->n_diffchr[0], s->n_diffchr[1]);
+    printf("%lld + %lld with mate mapped to a different chr (mapQ>=5)\n", s->n_diffhigh[0], s->n_diffhigh[1]);
+}
+
+static void out_fmt_json(bam_flagstat_t *s) {
+    char b0[16], b1[16];
+    char *str = "{ \"QC-passed reads\": {\
```
daviesrob: Is there any reason why you chose to use a separate variable for the format string (apart from tidiness)?

It's generally a bad idea to do this as it stops the compiler from checking that the format string matches the types of the parameters passed in.  For example, trying to compile:

```c
#include <stdio.h>
int main(int argc, char **argv) {
    long long int answer = 42;
    const char *fmt = "The answer is %d\n";

    printf(fmt, answer);

    printf("The answer is %d\n", answer);
}
```

results in:
```
$ cc test.c
test.c: In function ‘main’:
test.c:8:28: warning: format ‘%d’ expects argument of type ‘int’, but argument 2 has type ‘long long int’ [-Wformat=]
     printf("The answer is %d\n", answer);
                           ~^
                           %lld

```

So it accepted the first `printf` but not the second, even though both have the same bug.

Also, the spaces due to the indentation following each backslashed newline is ending up in the JSON output.  The best way to do multi-line strings in C is to use the fact that `"AB" "CD"` is equivalent to `"ABCD"` and put each line in its own set of double-quotes, ending up with something like this:
```c
printf("{ \"QC-passed reads\": {"
       " \"total\": \"%lld\","
       // Bit in middle skipped
       " \"with mate mapped to a different chr (mapQ >= 5)\": \"%lld\" } "
       "}\n",
       s->n_reads[0],
       // ... variables skipped ...
       s->n_diffhigh[1]);
```

You might also want to consider putting a `\n` on the end of each line so you get a pretty-printed JSON file.  This would both make it easier to read and allow simple tools like `awk` or `grep` to easily extract individual lines.  Any reasonable JSON parser should be able to handle the extra white-space.

bam_stat.c
```diff
@@ -97,15 +97,130 @@ static void usage_exit(FILE *fp, int exit_status)
...
+        s->n_pair_map[1],
+        s->n_sgltn[1], percent(b1, s->n_sgltn[1], s->n_pair_all[1]),
+        s->n_diffchr[1],
+        s->n_diffhigh[1]
+    );
+}
+
+static void out_fmt_tsv(bam_flagstat_t *s) {
+    char b0[16], b1[16];
+    printf("%lld \t %lld \t total (QC-passed reads + QC-failed reads)\n", s->n_reads[0], s->n_reads[1]);
```
daviesrob: I also just noticed, this should be `"%lld\t%lld\ttotal (QC-passed reads + QC-failed reads)\n"`, as currently they're strictly space-tab-space separated values.

raivivek: Thanks for pointing out.

bam_stat.c
raivivek: 1 -- Your intuition is correct. I mostly used for tidiness, but your idea is valid. I revised the code.
2 -- Agreed. I updated the code.

bam_stat.c
```diff
@@ -97,15 +97,129 @@ static void usage_exit(FILE *fp, int exit_status)
...
+    printf("%lld + %lld properly paired (%s : %s)\n", s->n_pair_good[0], s->n_pair_good[1], percent(b0, s->n_pair_good[0], s->n_pair_all[0]), percent(b1, s->n_pair_good[1], s->n_pair_all[1]));
+    printf("%lld + %lld with itself and mate mapped\n", s->n_pair_map[0], s->n_pair_map[1]);
+    printf("%lld + %lld singletons (%s : %s)\n", s->n_sgltn[0], s->n_sgltn[1], percent(b0, s->n_sgltn[0], s->n_pair_all[0]), percent(b1, s->n_sgltn[1], s->n_pair_all[1]));
+    printf("%lld + %lld with mate mapped to a different chr\n", s->n_diffchr[0], s->n_diffchr[1]);
+    printf("%lld + %lld with mate mapped to a different chr (mapQ>=5)\n", s->n_diffhigh[0], s->n_diffhigh[1]);
+}
+
+static void out_fmt_json(bam_flagstat_t *s) {
+    char b0[16], b1[16];
+    printf("{\n \"QC-passed reads\": { \n"
```
daviesrob: I've a few more comments on the JSON output, now it's a bit easier to read (thanks).

As [JSON](https://www.json.org/) values can be numbers, I think it would be better to use them instead of strings.  It may help avoid unnecessary conversions, depending on what is reading the data.

For ease of parsing the values with percentages, it might be better to split them into separate entries, so instead of:
```json
{
  "mapped": "32 (94.12%)", 
}
```
it could be like this (note the `%` is dropped from the value to turn it into a number):
```json
{
   "mapped": 32,
   "mapped %": 94.12
}
```
This does leave the problem of what to do when `total` is zero so you can't calculate a percentage.  Currently it uses `N/A` which would turn the value into a string.  Possibly more convenient options might be to output `0.0` so it stays a number, or maybe `null` to indicate a missing value.  The `0.0` option could be easily done using:
```c
s->n_reads[0] ? (float)s->n_mapped[0] / s->n_reads[0] * 100.0 : 0.0
```
`null` or `"N/A"` would require a json-specific version of the `percent()` function.

bam_stat.c
```diff
@@ -114,12 +228,15 @@ int bam_flagstat(int argc, char *argv[])
...
-        SAM_OPT_GLOBAL_OPTIONS('-', 0, '-', '-', '-', '@'),
+        SAM_OPT_GLOBAL_OPTIONS('-', 0, 'O', '-', '-', '@'),
         {NULL, 0, NULL, 0}
     };
 
-    while ((c = getopt_long(argc, argv, "@:", lopts, NULL)) >= 0) {
+    while ((c = getopt_long(argc, argv, "@:O:", lopts, NULL)) >= 0) {
         switch (c) {
+        case 'O':
+          strcpy(out_fmt, optarg);
```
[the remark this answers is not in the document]
raivivek: Thanks! Updated.

bam_stat.c
raivivek: @daviesrob I see. So the idea is to have the JSON output to be as reliant to the spec as possible. That makes sense. With that, I suppose using `null` makes more sense even though for consistency among output formats 'N/A' might be more useful.

daviesrob:
You need to actually use your `percent_json()` function:

```
bam_stat.c:96:20: error: 'percent_json' defined but not used [-Werror=unused-function]
```

The continuous integration tests turn on `-Werror` to ensure the package builds with no warnings.

daviesrob:
Thanks, I've now squashed and merged this.

PR #1109 contains a manual page update to go with this.

PR #1107, opened 2019-09-13
Prepare for 64-bit positions

daviesrob:
This does not add 64-bit position support, it just ensures samtools will build without warnings (using gcc/clang -Wall) when built against both 32-bit and 64-bit position branches of HTSlib.

Makes various printf()s use PRId64 for positions.  The values are cast to (int64_t) so both 32-bit and 64-bit versions work.

Fix `isize` calculation in bam_mate.c so it sign-extends correctly.


valeriuo:
Merged into develop as cbe1246a9c6729e2d0df2b57777f63f7dae771f4

PR #1121, opened 2019-10-14
Fixes an optical stats error.

whitwham:
Read paired with an unmapped mate now treated as single in the optical duplicate counts.


valeriuo:
If the mate is unmapped, shouldn't the condition at line [1193](https://github.com/samtools/samtools/blob/796cf226ff23771c754e9976fda85efa20414737/bam_markdup.c#L1193) be:
`if ((bp->p->core.flag & BAM_FPAIRED) && (bp->p->core.flag & BAM_FMUNMAP))` ?

whitwham:
No because we want to see if the single read is matched against a pair.  A pair in this case being defined as a read flagged as a pair with a mapped mate.

PR #1124, opened 2019-10-16
Add --verbosity INT global option for setting hts_verbose

jmarshall:
Currently the only way for users to crank up `hts_verbose` for debugging purposes is to use `htsfile -vvv`. It would be good to be able to debug e.g. networking and S3 access problems by cranking up the logging in `samtools` invocations too.

This adds a `--verbosity INT` global option. As this pretty much has to be a long option, it seems better to have an option that sets the level directly rather than make people type stuff like

    samtools view --verbose --verbose --verbose --verbose s3://bucket/blah.bam

(It might be worth in future also adding a less flexible `--verbose` global option that just sets it to 6 or so…)

### DIFF
--- a/bam2depth.c
+++ b/bam2depth.c
@@ -90,7 +90,7 @@ static int usage() {
     fprintf(stderr, "   -Q <int>            mapping quality threshold [0]\n");
     fprintf(stderr, "   -r <chr:from-to>    region\n");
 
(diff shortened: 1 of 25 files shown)


jkbonfield:
Thanks.  Good to tidy up the old code and get something we've needed for ages in there too.

PR #1128, opened 2019-11-05
Optical Duplicate Fixes

whitwham:
Checks for optical duplicates when multiple duplicates occur.
Multiple duplicates of single original can now be tagged with that originals read name.
Fixed broken PG option.
In stats changed entry name from DUPLICATE OPTICAL to DUPLICATE PAIR OPTICAL for more clarity.

### DIFF
--- a/bam_markdup.c
+++ b/bam_markdup.c
@@ -1211,6 +1211,7 @@ static int bam_mark_duplicates(md_param_t *param) {
         if (in_read->b->core.tid >= 0) { // -1 for unmapped reads
             if (in_read->b->core.tid < prev_tid ||
                ((in_read->b->core.tid == prev_tid) && (in_read->b->core.pos < prev_coord))) {
+                fprintf(stderr, "[markdup] error: not in coordinate sorted order.\n");
                 goto fail;
             }
         }
--- a/bam_markdup.c
+++ b/bam_markdup.c
@@ -857,7 +857,7 @@ static inline int optical_retag(md_param_t *param, khash_t(duplicates) *dup_hash
     int ret = 0;
     uint8_t *data;
 
-    // remove any existing do tag
+    // remove any existing dt tag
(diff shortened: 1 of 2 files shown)


bam_markdup.c
```diff
@@ -852,6 +853,195 @@ static int mark_duplicates(md_param_t *param, khash_t(duplicates) *dup_hash, bam
 }
 
 
+static inline int optical_retag(md_param_t *param, khash_t(duplicates) *dup_hash, bam1_t *b, int paired, long *optical_single, long *optical_pair) {
+    int ret = 0;
+    uint8_t *data;
+
+    // remove any existing do tag
```
jkbonfield: "do" typo of "dt"?

bam_markdup.c
```diff
@@ -852,6 +853,195 @@ static int mark_duplicates(md_param_t *param, khash_t(duplicates) *dup_hash, bam
...
+            uint8_t *data;
+
+            // at this stage all duplicates should have a do tag
+            if ((data = bam_aux_get(current->b, "do")) != NULL) {
+                // see if we need to change the tag
+                char *old_name = bam_aux2Z(data);
+
+                if (old_name) {
+                    if (strcmp(old_name, ori_name) != 0) {
+                        bam_aux_del(current->b, data);
```
jkbonfield: Note there are bam_aux_update functions now, but I'm not convinced they're worth using yet in terms of efficiency as they look rather like delete and append anyway (but even more complex).

Something to consider, but also definitely not a show stopper.

bam_markdup.c
whitwham: Yes it is.  I use (and remove) 'do' and 'dt' tags.  Must have got them mixed up here.

bam_markdup.c
whitwham: Okay, I think this is something to remember for next time.

valeriuo:
Merged into develop

PR #1133, opened 2019-11-08
Improved error recovery in faidx write_output()

daviesrob:
Fix bug where the wrong pointer was passed to free if `write_line()` failed on `qual`.

Add checks to fail early if `fai_fetch64()` or `fai_fetchqual64()` return `NULL` pointers, and return `EXIT_FAILURE` if they failed for any reason other than a missing region (previously the pointer would be ignored and the function would eventually return `EXIT_SUCCESS`).  In the case of `fai_fetchqual64()`, it's assumed that any missing regions will have been reported by `fai_fetch64()` already, so all `fai_fetchqual64()` errors can be treated as failure.

### DIFF
--- a/faidx.c
+++ b/faidx.c
@@ -134,8 +134,15 @@ static int write_output(faidx_t *faid, FILE *file, const char *name, const int i
                         const int length, const int rev,
                         const char *pos_strand_name, const char *neg_strand_name,
                         enum fai_format_options format) {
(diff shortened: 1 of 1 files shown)


daviesrob:
Oops, didn't notice that the same checks for `seq_len` are already in `write_line()`.  Will replace this with a commit for just the `free()` bug.

PR #1135, opened 2019-11-12
Fixed gcc9.2 warning on fprintf.

whitwham:
Removed the attempt to print out a null value.  Also put a check on unguarded memory allocations.


jkbonfield:
Fixed with minor edit as 419f8ac826cefe7e655f43c2a3e7ea56181bc284

PR #1136, opened 2019-11-12
Updates copyright message for 2019.

whitwham:
Added 2019 to copyright dates on files committed in 2019.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Makefile for samtools, utilities for the Sequence Alignment/Map format.
 #
-#    Copyright (C) 2008-2017 Genome Research Ltd.
+#    Copyright (C) 2008-2019 Genome Research Ltd.
 #    Portions copyright (C) 2010-2012 Broad Institute.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
--- a/bam.c
+++ b/bam.c
@@ -1,6 +1,6 @@
 /*  bam.c -- BAM format.
 
-    Copyright (C) 2008-2013, 2015 Genome Research Ltd.
+    Copyright (C) 2008-2013, 2015, 2019 Genome Research Ltd.
     Portions copyright (C) 2009-2012 Broad Institute.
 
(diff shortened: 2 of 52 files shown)


daviesrob:
Added man pages and pre-2019 updates that really should have been done earlier.

tseemann:
I believe that doing an open-ended year is legally correct.
This will solve the annual update problem.
eg. `(C) Copyright Mrs Sam Tools 2010-`

whitwham:
Our own internal guidelines say otherwise.  Since GRL pay our wages we really have to go by their interpretation of copyright law.

tseemann:
Agreed - you need to follow your employer's rules.

Just be aware that if you forget to git update on 1 Jan 2020 then I am stealing all the code as it is no longer copyright by GRL :-P

jkbonfield:
I know you're just joking of course and you're welcome to reuse the code given the liberal license. :-)

I think the point of the copyright assertions is to date the last substantial modification as copyright often lasts a certain time from authorship (or in some cases time since death of author, but I'm unsure where source code fits in this scheme of things).  This is why some people consider open ended dates to be watering down things - there is no clear definition of when the copyright is being asserted from, meaning it may perhaps be interpreted as counting down from the first date instead of the date of last modification.

When all said and done though it's largerly up to courts to decide things so giving them more accurate information to go on is rarely a bad thing, even though in reality things like git logs are a far better log than a basic date (but remember we ship tar balls too).

jmarshall:
As usual, the [GNU Maintainer Guide](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html) is a good summary of best practices in this area (and IFAIK pretty similar to GRL's conventions). Samtools et al to date have effectively considered individual source files to be individual works and not done ¶8's simpler “just bump **all** the copyright notices with the first non-trivial change in January” thing.

Me, I'm making jokes about which will come first: 2020 or an update to the footer at the bottom of <http://htslib.org/> :smile:

PR #1145, opened 2019-11-26
Fix ks_radixsort() on x86_64 gcc 4.4.7 -O2

daviesrob:
Fixes `make test` failure on Centos 6.  In an optimised build, its gcc swapped the order of stores to overlapping destinations in `ks_radixsort()`, causing part of the data to be overwritten.  Fixed by reworking the code so the stores no longer overlap.

Problem appeared when `ks_radixsort()` was upgraded to support 64-bit positions.


tseemann:
Wow - subtle bug!
Is there a way to help the C compiler to not optimise this incorrectly?
Does the `restrict` keyword in C99 help?

jkbonfield:
The compiler on Centos 6 is gcc 4.4 IIRC.  A nice idea, but it's highly unlikely it supports restrict.

bam_sort.c
```diff
@@ -1853,15 +1854,24 @@ static int ks_radixsort(size_t n, bam1_tag *buf, const sam_hdr_t *h)
     for (; max_tid > 0; max_tid >>= 8) tid_bytes++;
     assert(pos_bytes + tid_bytes < sizeof(buf[0].u.pos_tid));
 
+    tid_shift_l = pos_bytes * 8;
+    tid_shift_r = 64 - pos_bytes;
```
jkbonfield: Shouldn't this be `64 - tid_shift_l`?

daviesrob: Good spot.  Should now be fixed with an enhanced test to catch the bug.

PR #1151, opened 2019-12-10
Fix test suite on OpenBSD

jmarshall:
Samtools compiles fine on OpenBSD but there are test suite failures due to two separate causes:

1. `/bin/sh` on OpenBSD is ksh and it expands `{a,b,c}` brace notation somewhat differently from bash, which can be avoided by reformatting the awk expression to be a little clearer (it's not intended to be a shell brace expression at all!).

2. The `samtools merge -c -p` tests fail due to [OpenBSD's `srand48()`](https://www.mail-archive.com/tech@openbsd.org/msg21465.html) seeding a [stronger random number sequence](https://lwn.net/Articles/625506/) that differs from the prescribed POSIX sequence. There are various ways to fix this, but the best way may be to use the existing _htslib/hts_os.h_ to provide tools to get OpenBSD to also produce the (weaker, but for HTS purposes we probably value cross-platform repeatability more) standard POSIX sequences. Cf PR samtools/htslib#1002.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1426,7 +1426,7 @@ int bam_merge(int argc, char *argv[])
         return 1;
     }
 
-    srand48(random_seed);
+    hts_srand48(random_seed);
     if (!(flag & MERGE_FORCE) && strcmp(argv[optind], "-")) {
         FILE *fp = fopen(argv[optind], "rb");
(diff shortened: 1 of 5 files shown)


daviesrob:
Thanks, now rebased and merged.

PR #1156, opened 2019-12-18
Add support for multi-region merge

### LINKED ISSUE #768: samtools merge - add support for multiple regions
Samtools merge is used in [npg_ranger](https://github.com/wtsi-npg/npg_ranger) implementation of an htsget-compliant server.At the moment [htsget](https://github.com/samtools/hts-specs/blob/master/htsget.md) spec defines support for only one genomic region, but extending support to multiple regions is being discussed. If this is adopted, being able to call a merge for multiple regions would be crucial.

valeriuo:
`samtools merge` now accepts a BED file as a command line argument, supplied with the **-L** option, which allows the user to select more genomic regions in which to perform the merge of the alignment files. 
This new option extends the functionality of **-R** option, which only allowed one merge region to be specified, and the two options are mutually exclusive.

Fixes #768


bam_sort.c
```diff
@@ -911,6 +912,36 @@ int* rtrans_build(int n, int n_targets, trans_tbl_t* translation_tbl)
...
+        new_rl[i].min_beg = rl[i].min_beg;
+        new_rl[i].max_end = rl[i].max_end;
+
+        new_rl[i].reg = rl[i].reg;
+        new_rl[i].intervals = calloc(new_rl[i].count, sizeof(hts_pair_pos_t));
+        if (!new_rl[i].intervals) {
+            hts_reglist_free(new_rl, i);
+            return NULL;
+        }
+        int j;
```
daviesrob: Would be quicker and possibly simpler to do `memcpy(new_rl[i].intervals, rl[i].intervals, sizeof(new_rl[i].intervals[0]) * count)` here.

bam_sort.c
```diff
@@ -1098,58 +1128,95 @@ int bam_merge_core2(int by_qname, char* sort_tag, const char *out, const char *m
...
+
+            hts_idx_destroy(reg_idx);
         }
+
         free(rtrans);
         rtrans = NULL;
     } else {
         for (i = 0; i < n; ++i) {
-            if (hdr[i] == NULL) {
+            if (hts_get_format(fp[i])->format != sam) {
```
daviesrob: It might be possible to simplify this.  Back in commit 3fdca5b6e5 (when this test was added) the input header was destroyed for non-SAM files before getting here, so it used `sam_itr_queryi(NULL, HTS_IDX_REST, 0, 0)` to avoid having to use `hdr[i]` when getting records.  Now `hdr[i]` still exists, so `sam_read1()` should work for everything and it should be possible to just say `iter[i] = NULL;` for all file types.

test/merge/merge.bed
```diff
@@ -0,0 +1,4 @@
+ref1	1	4
+ref1	20	26
+ref1	32	45
+ref3	0	1
```
daviesrob: Can you add the missing newline on the end please?  Also for the input SAM files.

Makefile
```diff
@@ -131,7 +131,7 @@ print-version:
 .c.o:
 	$(CC) $(CFLAGS) $(ALL_CPPFLAGS) -c -o $@ $<
 
-LIBST_OBJS = sam_opts.o sam_utils.o
+LIBST_OBJS = sam_opts.o sam_utils.o bedidx.o
```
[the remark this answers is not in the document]
jmarshall: Also if you're adding something to `LIBST_OBJS` then you should remove it from `AOBJS`.

PR #1164, opened 2020-01-09
New dict option to add AN tag by adding/removing 'chr' from SN

jmarshall:
New option for the `dict` command that produces a very simple first draft of AN alternative names for the reference sequences, hitting the 90% requirement of just adding/removing `chr`. In lowercase only. And would require hand-editing for the 0.1% of people with reference sequences called `chrom1`. But it hits the use case for just about everybody, who have a `1`/`chr1` problem.


mcshane:
I've seen `chrM` <=> `MT` quite a lot, but would add some complexity to support?

jmarshall:
You're right, it wouldn't be hard to special-case that one too.

I'd be slightly nervous about species where `M` meant something else (some bird with even weirder sex chromosomes? some microbe labelled `A`,`B`,`C`,… rather than `I`,`II`,`III`,…?), but probably there's no such thing and if there is it's down in the noise level…

jmarshall:
Now emits e.g. `AN:M,chrMT,MT` for mitochondria.

daviesrob:
Thanks.  Out of interest, I asked @raskoleinonen what names have been used for mitochondria.  He gave the following list of "Mitochondria names used by INSDC/NCBI/ENA assembly database (all names with at least one occurrence)":

Name | Count
----|----
MT | 1795
mtDNA | 117
I | 34
mitogenome | 26
MIT | 15
1 | 13
15 | 12
M | 5
2 | 2
7 | 2
MITO | 2

So it looks like "MT" is by far the most popular.

PR #1165, opened 2020-01-10
Fix samtools build on Illumos

### LINKED ISSUE #653: Changes required to build on illumos (née OpenSolaris)

jmarshall:
Fix the build on OpenIndiana with Sun Studio 12. The main problem is a substandard awk implementation which makes `gmake test` fail, which we avoid by using `AC_PROG_AWK`so we instead get nawk on Illumos.

Also misc/ace2sam needs to pull in kstring.o on this platform, so link it against libhts.

Also Illumos doesn't have the `TIOCGWINSZ` ioctl, so conditionalise its use via #ifdef. Disabling this code leads to a similar effect to the ioctl() call failing, so still leads to columns being handled sensibly.

Fixes #653 and supersedes the outdated PRs #654 and #663. As noted on those PRs, at least on OpenIndiana you may need to configure --without-ncursesw.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@
 
 CC       = gcc
 AR       = ar
+AWK      = awk
 CPPFLAGS =
 #CFLAGS   = -g -Wall -O2 -pedantic -std=c99 -D_XOPEN_SOURCE=600
 CFLAGS   = -g -Wall -O2
@@ -231,8 +232,8 @@ check test: samtools $(BGZIP) $(TEST_PROGRAMS)
 	test/merge/test_bam_translate test/merge/test_bam_translate.tmp
 	test/merge/test_rtrans_build
 	test/merge/test_trans_tbl_init
-	cd test/mpileup && ./regression.sh mpileup.reg
-	cd test/mpileup && ./regression.sh depth.reg
+	cd test/mpileup && AWK="$(AWK)" ./regression.sh mpileup.reg
+	cd test/mpileup && AWK="$(AWK)" ./regression.sh depth.reg
 
 
 test/merge/test_bam_translate: test/merge/test_bam_translate.o test/test.o libst.a $(HTSLIB)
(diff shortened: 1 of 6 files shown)


daviesrob:
It builds! :smile:

On the curses front, building with plain `./configure` works, but with curses disabled in `tview`.  It finds the `ncursesw` library, but fails to find a matching header (it actually needs `ncurses/cursesw.h`).

`./configure --without-ncursesw` builds as noted, with a curses-enabled `tview`.

`./configure --without-curses` builds with curses disabled in `tview` (as you would expect).

`./configure --without-ncursesw --without-ncurses` fails to build:
```
"bam_tview_curses.c", line 270: undefined symbol: KEY_RESIZE
"bam_tview_curses.c", line 270: non-constant case expression
cc: acomp failed for bam_tview_curses.c
```

Apparently `KEY_RESIZE` is an `ncurses` feature.  It's a `#define`, so easy enough to check for.

Part of the problem is that `ax_with_curses.m4` chooses a library before checking that it has a suitable header file to go with it.  Autoconf archive has [an updated version](https://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob;f=m4/ax_with_curses.m4;hb=HEAD), but the main change seems to be adding `pkg-config` support (which drags in a load of extra dependencies).  It doesn't help much - the pkg-check test fails, and it falls back to probing with the same problem as the existing one.  I'm tempted to diverge and modify the one we've got so it only sets the outputs after it's found a working header.

jmarshall:
Yes, I lost interest in upstream _ax_with_curses.m4_ when one of the usual suspects rewrote it to be pkg-config-centric (the [antithesis of autoconf](https://twitter.com/jomarnz/status/826726790458597378), one might say) and renamed the output variables, and the maintainers accepted the patch without question. I see it still hasn't changed since. (On the up side, at least there is some attempt to probe for `tinfo` in that update but I haven't looked at the details.)

So I'd be in favour of your temptation to diverge and fix it locally.

jmarshall:
> Apparently KEY_RESIZE is an ncurses feature. It's a #define, so easy enough to check for.

It's interesting to compare the lists of `KEY_` #defines in <curses.h> and <ncurses/curses.h>: on Illumos, the ncurses one has `KEY_RESIZE` and `KEY_EVENT` added at the end that _curses.h_ doesn't have. OTOH the two files are the same on macOS and Linux…

daviesrob:
Thanks, will fix the `ax_with_curses.m4` presently.

PR #1176, opened 2020-01-24
[view, depad] Add support for reading remote reference files

valeriuo:
Use the new [HTSlib](https://github.com/samtools/htslib/pull/1017) method `fai_path`, which offers support for remote reference and index files.
Also, the reference file and the index file can now be provided as a single string argument, using the format **reference.fa##idx##index.fa.fai** (e.g. `samtools view -o align.cram -T ref.fa##idx##index.fa.fai align.bam`).

Fixes https://github.com/samtools/htslib/issues/933

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,11 @@
 Release a.b
 -----------
 
+ * samtools view now accepts remote URIs for FASTA and FAI files.
+   Furthermore, the reference and index file can be provided in a single
+   argument, such as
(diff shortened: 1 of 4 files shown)


jmarshall:
This PR has the byproduct of removing _sam.h_ and _sam.c_, i.e., it removes a large part of the legacy Samtools API. While it is now 2020 and removing that legacy is perhaps long overdue, it deserves to be considered explicitly — see #1184.

daviesrob:
I've restored _sam.c_ and _sam.h_ so their fate can be decided separately.

PR #1181, opened 2020-01-30
Spelling corrections in user-visible documentation

jmarshall:
Corrections in manual pages and other documentation read by end-users.


jkbonfield:
Thanks John.

PR #1185, opened 2020-02-03
Delete the "samtools import" subcommand.

### LINKED ISSUE #1182: samtools index: failed to create index:
Yes.
samtools 1.9
Using htslib 1.9

Linux 2.6.32-696.el6.x86_64

x86_64

gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-18)
Copyright (C) 2010 Free Software Foundation, Inc.

1)
I have run "samtools index -c -@ 20 ./output.sort.bam ./output.bam.csi",
the error information was "
[E::hts_idx_push] NO_COOR reads not in a single block at the end 19 -1
samtools index: failed to create index for "./output.bam": No such file or directory
"
2)
Before run index,I have also run "samtools sort -m 50G ./output.bam -o ./output.sort.bam" and it was completed,but the error still appear.

3)
It's worth noting that the bam file is about 75G and the length of chromosome is about 1GB~2GB,I'm not sure if the error is because of the file big size.

Thanks a lot,
Laura

jkbonfield:
This was deleted from the help / usage statements in 2009 (cc207d8)
between 0.1.5 and 0.1.6.  Thus it has never been a documented command
since the official launch of Samtools.

Import was just a wrapper around view, but the wrapper broke when we
added PG headers.  While a fix is trivial, this PR simply removes the
dead and untested code instead.

Fixes #1182 in as far as turning the segmentation fault into a handled
usage error instead.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,10 @@ Release a.b
    "chr" prefixes added to or removed from each sequence name as appropriate
    and listing both "M" and "MT" alternatives for mitochondria.
 
+ * The samtools import command, labelled as obsolete in May 2009 and
+   removed from all help and documentation later that year, has
+   finally been removed.  Use samtools view instead.
+
 Release 1.10 (6th December 2019)
 --------------------------------
 
--- a/bamtk.c
+++ b/bamtk.c
@@ -46,7 +46,6 @@ int bam_fillmd(int argc, char *argv[]);
 int bam_idxstats(int argc, char *argv[]);
 int bam_markdup(int argc, char *argv[]);
 int main_samview(int argc, char *argv[]);
-int main_import(int argc, char *argv[]);
 int main_reheader(int argc, char *argv[]);
 int main_cut_target(int argc, char *argv[]);
 int main_phase(int argc, char *argv[]);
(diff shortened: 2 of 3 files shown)


mr-c:
Could we be nice and show the new syntax to the user?

```
samtools import ex1.fa.fai ex1.sam.gz ex1.bam  
```
becomes
```
samtools view -bt ex1.fa.fai -o ex1.bam ex1.sam.gz
```

dpryan79:
I'd suggest not being nice here, since any user doing this is likely doing so many other things in antiquated ways that they should rethink everything.

jkbonfield:
Exactly my thought @dpryan79.  I wanted the user to think about what they're doing and look at the man page to see all the other features of view and to decide whether their usage is appropriate rather than suggest more cargo-cult cut paste non-thinking work.  Let's face it, since import was supported there have been new file formats, new SAM flags (SUPPLEMENTARY, etc) and so on.

Also IMO this is all a storm in a teacup.  The fact is there have been over 10,000 installs of samtools with the broken import command and not a single *user* reported it being broken.  We got acknowledgement an automatic test harness breaking, but it would appear in day to day usage this command has dropped out of common usage.

Issue #1184, opened 2020-02-03
Is it time to remove the legacy Samtools API?


jmarshall:
Since 2014, the samtools code itself no longer uses the legacy API declared in _bam.h_, _sam.h_ etc. (For the most part at least; it no longer uses the legacy accessor wrappers for things it uses directly from HTSlib — but it does use e.g. `bam_remove_B()` which comes from _bam.c_, a source file that mostly defines legacy wrappers.)

However when samtools was recoded in terms of HTSlib functions, the legacy API was left in place (‘temporarily’) for the benefit of existing third-party code using the now-legacy Samtools API. Since then, many but not all third-party programs have been recoded to use the HTSlib API directly.

The legacy API in samtools is mostly unmaintained and untested by the Samtools test suite, through it seems to have bitrotted less than one might expect.

Perhaps it is time to remove the legacy Samtools API entirely, with a prominent note in _NEWS_ directing people to the HTSlib API instead.

This would encourage straggling legacy-API-using projects either to make the move (yay) or to limit themselves to the final legacy-API-containing samtools (boo). Note that many such projects already bundle or otherwise encourage prehistoric versions of samtools (0.1.19, 1.2, …) anyway.

jkbonfield:
+1 to this.

Yes some people probably still use this old API in their embedded copies of samtools, but they're a lost cause anyway as whatever we do makes no difference to an embedded ancient copy.

There may be a few responsible people out there using the latest samtools, but I doubt many given that it doesn't install libbam.a.

If we're to do this, I'd advocate not only removal of the API functions but also removal of building a libbam.a at all.  Better that people get a "no such file or directory" than confusing over missing symbols.

jmarshall:
> If we're to do this, I'd advocate not only removal of the API functions but also removal of building a libbam.a at all

:+1: +1. Happily _libbam.a_ is only used in linking samtools, so whatever's left in `LOBJS` (already hardly anything) can easily be moved to `AOBJS` and then the _lib_ rules nuked.

There's indeed no helping people who have embedded ancient copies. OTOH removing it from current samtools would stop well-meaning but misguided packagers from packaging _libbam.a_ and the legacy headers — e.g. see bioconda/bioconda-recipes#19275 and bioconda/bioconda-recipes#19436.

PR #1192, opened 2020-02-12
Make more functions static.

jkbonfield:
Current policy is now all functions must be static, even in the executable.

The only ones left non-static are when used in other .c files or when
part of $(LOBJS) and hence part of libbam.a (although in time those
too will change).

(Note this has meant complete removal of some functions, along with
their warnings that they may be used by third party applications,
albeit presumably by dint of complete hackery and manually copying
over .c files into their own projects.)

### DIFF
--- a/bam2bcf.c
+++ b/bam2bcf.c
@@ -264,7 +264,7 @@ int bcf_call_glfgen(int _n, const bam_pileup1_t *pl, int ref_base, bcf_callaux_t
  *  reported values are slightly more favourable than those of a true random
  *  distribution.
  */
-double calc_vdb(int *pos, int npos)
+static double calc_vdb(int *pos, int npos)
 {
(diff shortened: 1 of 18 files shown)


mp15:
Looks good to me, if we get any complaints we could consider uplifting functions on request to htslib.

jkbonfield:
Absolutely that's the right thing to do where possible.  There are certainly a few useful utility functions in samtools which could be added to htslib.  Reverse complementing sequences,  some utility in argument processing, tmp file handling, etc.

However I think some people are driving samtools sub-commands as if they are functions in a library, eg sort, stats, etc.  IIRC this is how bioperl worked?  IMO that's wrong.  If you want to do that, use popen or similar and call the samtools command directly as a sub-process.

See also #1184

PR #1200, opened 2020-02-21
[depad] Fix the CIGAR op number arithmetic

valeriuo:
When `n < b->core.n_cigar`, `(n - b->core.n_cigar)` gets converted to a very large positive number, because `n_cigar` is unsigned. This could result in overflows or undesired results for subsequent operations.

### DIFF
--- a/padding.c
+++ b/padding.c
@@ -38,20 +38,38 @@ DEALINGS IN THE SOFTWARE.  */
 
 #define bam_reg2bin(b,e) hts_reg2bin((b),(e), 14, 5)
 
-static void replace_cigar(bam1_t *b, int n, uint32_t *cigar)
+static int replace_cigar(bam1_t *b, uint32_t n, uint32_t *cigar)
 {
+    int diff = 0;
     if (n != b->core.n_cigar) {
         int o = b->core.l_qname + b->core.n_cigar * 4;
-        if (b->l_data + (n - b->core.n_cigar) * 4 > b->m_data) {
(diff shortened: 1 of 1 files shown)


daviesrob:
While looking at this, it would be a good idea to add a check that the `realloc()` worked, and also add a check that the resulting BAM record won't be too big.  Something like:
```c
if ((INT_MAX - b->l_data)/4 < (n - b->core.n_cigar)) {
   fprintf(stderr, "BAM record too big\n");
   exit(1);
}
```

padding.c
```diff
@@ -38,20 +38,38 @@ DEALINGS IN THE SOFTWARE.  */
...
+    int diff = 0;
     if (n != b->core.n_cigar) {
         int o = b->core.l_qname + b->core.n_cigar * 4;
-        if (b->l_data + (n - b->core.n_cigar) * 4 > b->m_data) {
-            b->m_data = b->l_data + (n - b->core.n_cigar) * 4;
-            kroundup32(b->m_data);
-            b->data = (uint8_t*)realloc(b->data, b->m_data);
+        if (n > b->core.n_cigar) {
+            diff = (n - b->core.n_cigar) * 4;
+            if ((INT_MAX - b->l_data) < diff) {
```
daviesrob: `diff` could wrap around here for sufficiently large `n - b->core.n_cigar`, causing the test to fail.  That's why it's better to use `(INT_MAX - b->l_data)/4 < (n - b->core.n_cigar)` because both sides will fit in a `uint32_t`.

PR #1224, opened 2020-04-09
Minor documentation formatting fixes

jmarshall:
Use actual tab characters to delimit .TS/.TE table rows. (All the other tables in _samtools/doc/*.1_ are already using tabs.)

To improve the formatting on _htslib.org/doc_: man2fhtml converts to HTML one line at a time, so works best when sequences of `\fB...\fR` font escapes are kept within a single line. (It would be possible for man2fhtml to carry the font state from line to line and hopefully not introduce other problems, but Don't Do That is easier for now!)

### DIFF
--- a/doc/samtools-stats.1
+++ b/doc/samtools-stats.1
@@ -58,33 +58,33 @@ detailed descriptions.
 
 .TS
 lb l .
-CHK     Checksum
-SN      Summary numbers
-FFQ     First fragment qualities
-LFQ     Last fragment qualities
-GCF     GC content of first fragments
-GCL     GC content of last fragments
-GCC     ACGT content per cycle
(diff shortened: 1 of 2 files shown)


daviesrob:
Thanks for the update.  I have a patch for `man2fhtml` that makes it work with font escapes on different lines, and also works when `\fB` and `\fI` have been mixed without going through `\fP` or `\fR`.  I will try to remember to push it upstream...

jmarshall:
Thanks.

The latter should be already independently fixed upstream, and the former's a good idea too really. BTW hopefully you also noticed samtools/www.htslib.org#28 which fixes some garbledness in the _samtools-mpileup_ page and is what lead to this PR.

PR #1227, opened 2020-04-15
Ampliconstats

jkbonfield:
First stab at adding `samtools ampliconstats` and an associated `misc/plot-ampliconstats` script to generate more plots than you strictly needed.

It's a work in progress, but putting it out there for ideas and third party testing.
(Talking of which: it has none currently.)

### DIFF
--- a/amplicon_stats.c
+++ b/amplicon_stats.c
@@ -1,0 +1,942 @@
+/*  stats.c -- This is the former bamcheck integrated into samtools/htslib.
+
+    Copyright (C) 2020 Genome Research Ltd.
+
+    Author: James Bonfield <jkb@sanger.ac.uk>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.  */
+
+/*
+ * This tool is designed to give "samtools stats" style output, but dedicated
+ * to small amplicon sequencing projects.  It gathers stats on the
+ * distribution of reads across amplicons.
+ */
+
+/*
+ * TODO:
+ * - Cope with arbitrary sizes of reference and numbers of amplicons
+ * - Cope with multiple references.  What do we do here?  Just request one?
+ * - Permit regions rather than consuming whole file (maybe solves above).
+ * - Detect double-length products (or primer pool hopping).
+ * - eg 31_LEFT to 33_RIGHT, or even 31_LEFT to 32_RIGHT.
+ *   Binning via TLEN field will suffice.
+ */
+
+#include <config.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <inttypes.h>
+#include <getopt.h>
+#include <unistd.h>
+#include <math.h>
+
+#include <htslib/sam.h>
+#include <htslib/hfile.h>
+
+#include "samtools.h"
+#include "sam_opts.h"
+#include "bam_ampliconclip.h"
+
+#ifndef MIN
+#define MIN(a,b) ((a)<(b)?(a):(b))
+#endif
+
+#ifndef MAX
+#define MAX(a,b) ((a)>(b)?(a):(b))
+#endif
+
+#ifndef ABS
+#define ABS(a) ((a)>=0?(a):-(a))
+#endif
+
+#define MAX_AMP_LEN 1000   // Maximum length of any single amplicon
+#define MAX_PRIMER_PER_AMPLICON 4  // Max primers per LEFT/RIGHT
+#define MAX_DEPTH 5        // Number of different depths permitted
+#define MAX_TLEN 2000      // TLEN distribution size
+
+typedef struct {
+    sam_global_args ga;
+    uint32_t flag_require;
+    uint32_t flag_filter;
+    bed_pair_list_t sites;
+    int max_delta; // Used for matching read to amplicon primer loc
+    int min_depth[MAX_DEPTH]; // Used for coverage; must be >= min_depth deep
+    int use_sample_name;
+    int max_amp;
+    int64_t max_len;
+    int tlen_adj; // Adjust tlen by this amount, due to clip but no fixmate
+    FILE *out_fp;
+} astats_args_t;
+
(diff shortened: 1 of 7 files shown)


gkarthik:
[amplicon_stats.c#L174](https://github.com/jkbonfield/samtools/blob/6fc00ccb5e54d4e9bbe7b97d36eb396341ee2015/amplicon_stats.c#L174)

AFAIK, BED files are usually sorted by position, so in case of tiled amplicons, the assumption that primers are adjacent will not hold and this will be even trickier with overlapping primers. Struggled with this for a bit while writing [ivar](https://github.com/andersen-lab/ivar). Finally just added option for user to provide `.tsv` file with primer pairs to identify amplicons.

jkbonfield:
It's true they could possibly be sorted, which we don't handle at the moment, but there is no requirement for bed to be sorted by position.  Indeed the COG-UK primers for Covid-19 amplicon sequencing are sorted by primer, not by position, hence why I put the assumption in there (wanting something working asap).

Agreed it's something that could be relaxed.  However I think doing so requires parsing of the primer names, which gets far messier.  I'm wondering what's the lesser evil.  Documenting we need a specific type of parsable name so we can collate primers, or documenting that the need to be sorted by amplicon location with fwd and then rev primer together?

jmarshall:
> yes an alternative input format with the left/right on the same line would be so much easier

Possibly a job for [BEDPE](https://bedtools.readthedocs.io/en/latest/content/general-usage.html#bedpe-format), although the ‘or more’ in “one **or more** LEFT primers followed by one **or more** RIGHT primers” wouldn't fit very well…

jkbonfield:
An example of the input data we currently have (source: https://artic.network/ncov-2019)

```
MN908947.3      3771    3795    nCoV-2019_13_LEFT       nCoV-2019_1     +
MN908947.3      4142    4164    nCoV-2019_13_RIGHT      nCoV-2019_1     -
MN908947.3      4054    4077    nCoV-2019_14_LEFT       nCoV-2019_2     +
MN908947.3      4044    4068    nCoV-2019_14_LEFT_alt4  nCoV-2019_2     +
MN908947.3      4428    4450    nCoV-2019_14_RIGHT      nCoV-2019_2     -
MN908947.3      4402    4424    nCoV-2019_14_RIGHT_alt2 nCoV-2019_2     -
MN908947.3      4294    4321    nCoV-2019_15_LEFT       nCoV-2019_1     +
MN908947.3      4296    4322    nCoV-2019_15_LEFT_alt1  nCoV-2019_1     +
MN908947.3      4674    4696    nCoV-2019_15_RIGHT      nCoV-2019_1     -
MN908947.3      4666    4689    nCoV-2019_15_RIGHT_alt3 nCoV-2019_1     -
MN908947.3      4636    4658    nCoV-2019_16_LEFT       nCoV-2019_2     +
MN908947.3      4995    5017    nCoV-2019_16_RIGHT      nCoV-2019_2     -
```

I completely agree there are better formats, but I wrote the code to deal with the format as supplied, not our own alternative format, as that was simply the expedient approach.  This is what I mean by it being sorted by primer name, not position, and my desire to not have to parse those strings.  I just look for alternating +/- instead.  Not pleasant either, but rock and hard place springs to mind!

I think for now we just need to make sure this is better documented and in time add support for improved primer file formats, but that probably needs to be lead by the users rather than us second guessing them.  Right now they've decreed the above format is the one they're using.

jmarshall:
(Totally agree, just making sure BEDPE was on your radar in general.)

gkarthik:
Agree on all points. I think if you want to get something out asap, alternating strand should do the trick! Another point to note is that the fifth column in the standard BED format is actually a score but in V3 Bed (but not V2 and V1) that column is the amplicon name.  

Relevant issue and comment - 
https://github.com/artic-network/artic-ncov2019/issues/12#issuecomment-604726854

amplicon_stats.c
```diff
@@ -0,0 +1,924 @@
...
+        }
+
+        if (args->ga.nthreads > 0)
+            hts_set_threads(fp, args->ga.nthreads);
+
+        if (!(header = sam_hdr_read(fp)))
+            goto err;
+
+        if (args->use_sample_name)
+            sname = (char *)get_sample_name(header, NULL);
```
daviesrob, line 736: Looks like `sname` gets leaked it it's set here?

amplicon_stats.c
```diff
@@ -0,0 +1,924 @@
...
+
+char const *get_sample_name(sam_hdr_t *header, char *RG) {
+    kstring_t ks = {0};
+    sam_hdr_find_tag_id(header, "RG", RG?"ID":NULL, RG, "SM", &ks);
+    return ks.s;
+}
+
+int64_t get_ref_len(sam_hdr_t *header, char *SQ) {
+    kstring_t ks = {0};
+    sam_hdr_find_tag_id(header, "SQ", SQ?"ID":NULL, SQ, "LN", &ks);
```
daviesrob, line 690: If this fails, `ks.s` will be NULL, causing a blow-up on the next line.
You could also instead try:
```
int tid = SQ ? sam_hdr_name2tid(header, SQ) : 0;
return tid >= 0 ? sam_hdr_tid2len(header, tid) : -1;
```
which removes the need for the kstring.

jkbonfield:
> We could really do with a better input format, which is more explicit about where the amplicons are...

Agreed I was thinking this yesterday.  A summary of maximal (or minimal) start/end per amplicon would permit more useful plots I think.

I have plans for plots based on genomic coords as well as purely amplicon number, and those would benefit from an alternate background shading for the amplicon locations.

amplicon_stats.c
```diff
@@ -0,0 +1,943 @@
...
+    int64_t end = bam_endpos(b);
+
+    int anum = (b->core.flag & BAM_FREVERSE)
+        ? (end-1 >= 0 && end-1 < args->max_len ? pos2end[end-1] : -1)
+        : (start >= 0 && start < args->max_len ? pos2start[start] : -1);
+
+
+    // ivar sometimes soft-clips 100% of the bases.
+    // This is essentially unmapped
+    if (end == start && (args->flag_filter & BAM_FUNMAP)) {
```
daviesrob, line 365: Did you mean to use `||` here?

doc/samtools-ampliconstats.1
```diff
@@ -0,0 +1,280 @@
...
+
+The alignment files should have previously been clipped of primer
+sequence, for example by "samtools ampliconclip" and the sites of
+these primers should be specified as a bed file in the arguments.
+Each amplicon must be present in the bed file with one or more LEFT
+primers (direction "+") followed by one or more RIGHT primers.  For
+example:
+
+.EX 2
+MN908947.3  1875  1897  nCoV-2019_7_LEFT        nCoV-2019_1  +
```
daviesrob, line 63: Strictly this is an invalid BED file.  Column 5 should be a score, i.e. a number.  The V2 `nCoV-2019.bed` got it right, the V3 one was wrong.

misc/plot-ampliconstats
```diff
@@ -0,0 +1,884 @@
...
+# Command line argument handling & help
+
+my %opts;
+
+my $res = GetOptions("size=s"      => \$opts{size},
+                     "size2=s"     => \$opts{size2},
+                     "size3=s"     => \$opts{size3},
+                     "help"        => \$opts{help},
+                     "page=i"      => \$opts{page},
+		     "amp_add=i"   => \$opts{amp_add},
```
daviesrob, line 42: It looks like there are some tabs here.  It looks like the anti-tab system missed them due to no `.pl` on the file name.

amplicon_stats.c
daviesrob: Actually, probably not.

jkbonfield:
Rebased and squashed.

amplicon_stats.c
```diff
@@ -0,0 +1,943 @@
...
+                                     amplicon_t *amp, int namp,
+                                     int64_t max_len) {
+    int64_t i, j;
+
+    if (!pos_lookup_len) {
+        pos_lookup_len = max_len;
+        if (!(pos2start = calloc(max_len+1, sizeof(*pos2start))))
+            return -1;
+        if (!(pos2end   = calloc(max_len+1, sizeof(*pos2start)))) {
+            free(pos2start);
```
daviesrob, line 118: Just noticed a small bug.  If `pos2end` fails to allocate, `pos2start` will be double freed by `free_amp_pos_lookup()` if it's also done here.

jkbonfield: Good spot.  Fixed.

PR #1230, opened 2020-04-24
Handle records with no sequence in calmd + add error checks

### LINKED ISSUE #1229: calmd halts in an input record has SEQ=*, with incomprehensible error messages
####
samtools 1.10

####
- OS: Linux 4.4.0-141-generic
- compiler: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609

####
Small example is available at https://psu.app.box.com/s/kzewixahnr8p9b290u8zy83mt05xbg18/file/656631838103

Run this command:
```
samtools calmd -e \
      --output-fmt SAM \
      reads_vs_assembly.bam \
      temp/${assembly}.fa \
  > reads_vs_assembly.calmd.sam \
  2> reads_vs_assembly.calmd.log
```

Get this failure:
```
[bam_fillmd1] different NM for read 'm74581_185104_510317/7374047/5_37840': 86 -> 697
[E::bam_aux_get] Corrupted aux data for read m74581_185104_510317/7374047/5_37840
[E::sam_format1_append] Corrupted aux data for read m74581_185104_510317/7374047/5_37840
samtools calmd: failed to write to output file: Invalid argument
```

I had originally reported this as issue lh3/minimap2#599 for minimap2 (since the bam file was created by minimap2). @jmarshall figured out the problem is that the BAM file contains records with SEQ=*. Removing these records from the BAM file sidesteps the problem.

Probably not relevant, but the BAM file was created by minimap2, like this:
```
minimap2 -t 3 -x map-pb -a --cs \
    ${assemblyPath}/${assembly}.idx \
    reads.fastq.gz \
  | tee reads_vs_assembly.sam \
  | samtools view -b -o reads_vs_assembly.bam
```

daviesrob:
* Replace some one-letter variable names in `bam_fillmd1_core()` to make it easier to understand
* Move a kstring to the stack to get rid of a `calloc()`
* Add lots of error checking to `bam_fillmd1_core()` and ensure it doesn't read or write beyond the end of the sequence in the alignment record.
* Make `bam_fillmd1_core()` skip over alignment records where the sequence has not been stored and make it report when this happens.

Fixes #1229

### DIFF
--- a/bam_md.c
+++ b/bam_md.c
@@ -30,6 +30,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <string.h>
 #include <ctype.h>
 #include <limits.h>
+#include <errno.h>
 #include "htslib/faidx.h"
 #include "htslib/sam.h"
 #include "htslib/kstring.h"
@@ -46,7 +47,7 @@ DEALINGS IN THE SOFTWARE.  */
 
 int bam_aux_drop_other(bam1_t *b, uint8_t *s);
 
-void bam_fillmd1_core(bam1_t *b, char *ref, hts_pos_t ref_len, int flag, int max_nm, int quiet_mode)
+int bam_fillmd1_core(bam1_t *b, char *ref, hts_pos_t ref_len, int flag, int max_nm, int quiet_mode)
 {
     uint8_t *seq = bam_get_seq(b);
     uint32_t *cigar = bam_get_cigar(b);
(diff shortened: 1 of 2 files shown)


bam_md.c
```diff
@@ -46,110 +47,144 @@ DEALINGS IN THE SOFTWARE.  */
...
+                    goto aux_fail;
             }
         }
     }
 
     // drop all tags but RG
     if (flag&DROP_TAG) {
         uint8_t *q = bam_aux_get(b, "RG");
-        bam_aux_drop_other(b, q);
+        if (q) bam_aux_drop_other(b, q);
```
valeriuo, line 187: `bam_aux_drop_other` already dealt internally with a NULL `q`. What is the reasoning behind this extra-check?

daviesrob: Yes, I think I mistook what it did with `q == NULL`.  I'll remove this.

daviesrob: The drop-all-tags "feature" is restored.  Although I'm not entirely sure what it's for - it's not documented and it seems to get rid of the `MD` and `NM` tags that were just calculated.

PR #1231, opened 2020-04-24
Add documentation and both ends clipping to ampliconclip

whitwham:
Put in a basic version of both ends clipping.  Add tests and man pages.


whitwham:
Not quite ready yet.  Still need to do more tests on clipping both ends, which means going through more PacBio data.  Also need to add a test for that.

doc/samtools-ampliconclip.1
```diff
@@ -0,0 +1,116 @@
+'\" t
+.TH samtools-ampliconclip 1 "6 December 2019" "samtools-1.10" "Bioinformatics tools"
+.SH NAME
+samtools ampliconclip \- clip reads using a BED file
+.\"
+.\" Copyright (C) 2008-2011, 2013-2020 Genome Research Ltd.
+.\" Portions copyright (C) 2010, 2011 Broad Institute.
+.\"
+.\" Author: Heng Li <lh3@sanger.ac.uk>
+.\" Author: Joshua C. Randall <jcrandall@alum.mit.edu>
```
mcshane, line 10: You should give yourself some credit here, @whitwham

whitwham: Ah yes, I just copied over the boilerplate from another file.

daviesrob:
Looks good.  Exactly which nugs did you fix in the first commit?

whitwham:
> Looks good. Exactly which bugs did you fix in the first commit?

The length of l_data wasn't being set properly when the read was hard clipped.  It was leading to aux tag strangeness.

daviesrob:
Thanks.  I'll add them to the squashed commit message.

doc/samtools-ampliconclip.1
tseemann: You used variable numbers > 1 character so it's definitenly not Heng's :-)

PR #1233, opened 2020-04-30
Fix Makefile dependencies for newly added source files

jmarshall:
I've got my script for this mostly converted for use in _.travis.yml_ and `make maintainer-check` but really must get around to finishing that up…


jkbonfield:
Oops, you can see which line I cut and pasted from there. :-)

Thanks John.  Yes a maintainer-check to spot these would be useful.

PR #1236, opened 2020-05-04
More ampliconstats improvements.

### LINKED ISSUE #1234: ampliconstats suggestions
1. Add the command line and version used generate the ampliconstats file to the header so users can, for instance, check what BED file was used to generate the file. Alternatively or in addition to the command line, could duplicate the BED file within the ampliconstats file which could then be used for additional plots within plot-ampliconstats (or custom plots from the stats file)

2. Increase MAX_AMP_LEN or add option to configure this - we are testing 1kb amplicons and had to adjust this.

3. Related to above would be good to add some detail about the coordinates being checked to the error messages when checking bounds on the BED file. We had some custom BED files that had mistakes and adding to the error messages helped track that down.

jkbonfield:
This fixes #1234 by making MAX_AMP_LEN configurable, adding command line and program version to the output, and improving some error messages.

The TLEN output has also been completely rewritten and is now TCOORD instead.  TLEN wasn't used by any plots and it's not that helpful either.  The actual start and end coords for the template are more informative than length as it permits us to count the frequency of each primer being used and to see which alts are working.  Note to be useful requires fixmates to have been run after primer clipping.

TODO:  write a plot for TCOORD (but I already know what I want this to look like as I've done manual constructions of it)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,7 +207,7 @@ sam_view.o: sam_view.c config.h $(htslib_sam_h) $(htslib_faidx_h) $(htslib_khash
 sample.o: sample.c config.h $(sample_h) $(htslib_khash_h)
 stats_isize.o: stats_isize.c config.h $(stats_isize_h) $(htslib_khash_h)
 stats.o: stats.c config.h $(htslib_faidx_h) $(htslib_sam_h) $(htslib_hts_h) $(htslib_hts_defs_h) $(htslib_khash_str2int_h) $(samtools_h) $(htslib_khash_h) $(htslib_kstring_h) $(stats_isize_h) $(sam_opts_h) $(bedidx_h)
-amplicon_stats.o: amplicon_stats.c config.h $(htslib_sam_h) $(samtools_h) $(sam_opts_h) bam_ampliconclip.h
+amplicon_stats.o: amplicon_stats.c config.h $(htslib_sam_h) $(htslib_khash_h) $(samtools_h) $(sam_opts_h) bam_ampliconclip.h
 bam_markdup.o: bam_markdup.c config.h $(htslib_thread_pool_h) $(htslib_sam_h) $(sam_opts_h) $(samtools_h) $(htslib_khash_h) $(htslib_klist_h) $(htslib_kstring_h) $(tmp_file_h)
 tmp_file.o: tmp_file.c config.h $(tmp_file_h) $(htslib_sam_h)
 bam_ampliconclip.o: bam_ampliconclip.c config.h $(htslib_thread_pool_h) $(sam_opts_h) $(htslib_hts_h) $(htslib_hfile_h) $(htslib_kstring_h) $(htslib_sam_h) $(samtools_h) bam_ampliconclip.h
--- a/amplicon_stats.c
+++ b/amplicon_stats.c
@@ -30,12 +30,8 @@ DEALINGS IN THE SOFTWARE.  */
 
 /*
  * TODO:
- * - Cope with arbitrary sizes of reference and numbers of amplicons
  * - Cope with multiple references.  What do we do here?  Just request one?
  * - Permit regions rather than consuming whole file (maybe solves above).
- * - Detect double-length products (or primer pool hopping).
- * - eg 31_LEFT to 33_RIGHT, or even 31_LEFT to 32_RIGHT.
- *   Binning via TLEN field will suffice.
  */
 
 #include <config.h>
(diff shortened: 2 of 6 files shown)


jkbonfield:
One possible improvement is to change the AMPLICON lines to also include usage count.  Ie START-END/FREQ(,START-END/FREQ)*.  We may then want to have FAMPLICON and CAMPLICON so we can view per file amplicon primer usage and combined primer usage.

The plan here is, where possible to assign each read not only to an amplicon, but also to a specific primer (this may be hard in some cases).  When we have a choice of alt primers we can then report back which ones are being utilised.  This may permit better optimisation / normalisation of primer usage.

Of use?

jkbonfield:
TCOORD plots now written.  An example output showing a slightly problematic run with some primer-skipping (red) and an unknown end coordinate (black) caused by primer-dimer mediated mispriming.

tcoord

mcshane:
Some suggested edits for the help message:
```diff --git a/amplicon_stats.c b/amplicon_stats.c
--- a/amplicon_stats.c
+++ b/amplicon_stats.c
@@ -953,18 +953,18 @@ static int usage(astats_args_t *args, FILE *fp, int exit_status) {
 "\n"
 "Options:\n");
     fprintf(fp, "  -f, --required-flag STR|INT\n"
-            "               Only include reads with all  of the FLAGs in present [0x%X]\n",args->flag_require);
+            "               Only include reads with all of the FLAGs present [0x%X]\n",args->flag_require);
     fprintf(fp, "  -F, --filter-flag STR|INT\n"
-            "               Only include reads with none of the FLAGs in present [0x%X]\n",args->flag_filter & 0xffff);
+            "               Only include reads with none of the FLAGs present [0x%X]\n",args->flag_filter & 0xffff);
     fprintf(fp, "  -a, --max-amplicons INT\n"
             "               Change the maximum number of amplicons permitted [%d]\n", MAX_AMP);
     fprintf(fp, "  -l, --max-amplicon-length INT\n"
-            "               Change the maximum length of an individual ampicon [%d]\n", MAX_AMP_LEN);
+            "               Change the maximum length of an individual amplicon [%d]\n", MAX_AMP_LEN);
     fprintf(fp, "  -d, --min-depth INT[,INT]...\n"
             "               Minimum base depth(s) to consider position covered [%d]\n", args->min_depth[0]);
     fprintf(fp, "  -m, --pos-margin INT\n"
-            "               Margin of error for matching primer positions [%d]n", args->max_delta);
-    fprintf(fp, "  -o, FILE\n"
+            "               Margin of error for matching primer positions [%d]\n", args->max_delta);
+    fprintf(fp, "  -o, --output FILE\n"
             "               Specify output file [stdout if unset]\n");
     fprintf(fp, "  -s, --use-sample-name\n"
             "               Use the sample name from the first @RG header line\n");
@@ -1000,6 +1000,7 @@ int main_ampliconstats(int argc, char **argv) {
         {"flag-require", required_argument, NULL, 'f'},
         {"flag-filter", required_argument, NULL, 'F'},
         {"min-depth", required_argument, NULL, 'd'},
+        {"output", required_argument, NULL, 'o'},
         {"pos-margin", required_argument, NULL, 'm'},
         {"use-sample-name", no_argument, NULL, 's'},
         {"max-amplicons", required_argument, NULL, 'a'},```

**Comment by jkbonfield** [comment]
Thanks Shane.  I took the liberty of assigning you as author of the commit (your patch above) using your Sanger address as per previous samtools commits.  If you wish me to pick a new email I can edit the commit.

**Comment by daviesrob** [comment]
While not strictly part of this PR, what is the `amp_add` fudge-factor for in `plot-ampliconstats`?  It can lead to confusing results when the number of reads present is small.

In the case I was looking at, the `combined-amp` plot had much higher bars than the individual plots on some amplicons that had (mostly) failed to sequence.

**Comment by daviesrob** [comment]
I ran into a problem when attempting to merge this.  The tests fail because they run over `$$opts{path}/ampliconclip/*.expected.sam` and 88ccfa2 added two more `expected.sam` files.

Would it be best to limit the input to the files that currently exist on this branch:
```
test/ampliconclip/1_hard_clipped.expected.sam
test/ampliconclip/1_soft_clipped.expected.sam
test/ampliconclip/1_soft_clipped_strand.expected.sam
test/ampliconclip/2_both_clipped.expected.sam
```

jkbonfield:
I guess it depends whether we think the new files add anything to testing of ampliconstats.  I don't have specific tests purposely for ampliconstats, although maybe I should craft some.  My strategy was any test is better than none so to start with existing data and work from there.

I'd be happy with either including the new ones or limiting it.

PR #1238, opened 2020-05-06
 Fix no argument ampliconstats run

valeriuo:
`isatty` doesn't work on MinGW consoles resulting in a **segfault**, when `samtools ampliconstats` is run with no arguments.

### DIFF
--- a/amplicon_stats.c
+++ b/amplicon_stats.c
@@ -917,18 +917,20 @@ int main_ampliconstats(int argc, char **argv) {
         }
     }
 
-    if (argc <= optind+1 && isatty(STDIN_FILENO))
+    if (argc <= optind)
         return usage(&oargs, stdout, EXIT_SUCCESS);
+    if (argc <= optind+1 && isatty(STDIN_FILENO))
+        return usage(&oargs, stderr, EXIT_FAILURE);
 
     int64_t longest;
-    if (load_bed_file_pairs(argv[optind++], 1, 0, &args.sites, &longest)) {
+    if (load_bed_file_pairs(argv[optind], 1, 0, &args.sites, &longest)) {
         print_error_errno("ampliconstats",
-                          "Could not read file \"%s\"", optarg);
+                          "Could not read file \"%s\"", argv[optind]);
         return 1;
     }
 
     int ret;
-    if (argc == optind) {
+    if (argc == ++optind) {
         char *av = "-";
         ret = amplicon_stats(&args, &av, 1);
     } else {


jmarshall:
The crash occurs on other platforms too. `isatty`/MinGW is a red herring; the real problem is not checking the required `primers.bed` argument is present.

This could be just

```c
if (argc <= optind)
         return usage(&oargs, stdout, EXIT_SUCCESS);
```

as this suffices for `samtools ampliconstats` to give the usage.

daviesrob:
It should probably be:
```c
    if (argc <= optind+1) {
        if (argc == optind || isatty(STDIN_FILENO)) {
            return usage(&oargs, stdout, EXIT_FAILURE);
        }
    }
```

Also I see some brokenness in the [error reporting](https://github.com/samtools/samtools/blob/89760c1/amplicon_stats.c#L925) from `load_bed_file_pairs()`.  It uses `optarg`, which is wrong as it isn't in the `getopt()` loop.  It should instead use `argv[optind]`, and optind should be incremented after the if statement instead of as a side-effect in the condition.

jmarshall:
What it _should_ be is

```c
if (argc <= optind)
    return usage(&oargs, stdout, EXIT_SUCCESS);
if (argc <= optind+1 && isatty(STDIN_FILENO))
    return usage(&oargs, stderr, EXIT_FAILURE);
```

to retain the no-arguments `samtools ampiconstats` usage request going to stdout, but having the interactive "need explicit `-` to read from stdin" go to stderr. However due to the required BED file argument, you don't need this `isatty` trickery to distinguish this case from the no-arguments-usage-request case, so really the checking is sufficient without the second `if` statement… :smile:

daviesrob:
Don't you need the second one to catch the waiting endlessly for input problem if you supply a bed file but no bams? (Apart from on Windows, due to `isatty()` not working.)

jmarshall:
Define “need” :smile: The (pretty minor) point that I'm not communicating very well is:

* You **need** to catch the waiting endlessly problem when people type `samtools view` or `samtools ampliconstats`, because we have the convention that people can type subcommands without any arguments to get a usage display — equivalently to other programs' `--help`.

* But maybe you don't need to catch the waiting endlessly problem for `samtools ampliconstats foo.bed` as people who have typed the BED file argument can perhaps be assumed to know that they're giving it a BAM file on stdin, and because samtools does not have a different conventional meaning for this that we're trying to separate out behaviour for.

However now that we've spent so much time discussing it, there's nothing to gain by simplifying and both `if` statements should just go in. (But there are good reasons to have the `stdout`+`SUCCESS` versus `stderr`+`FAILURE` behaviour different for the two cases.)

PR #1241, opened 2020-05-08
update bam2depth.c.

TuBieJun:
Add bam_mplp_init_overlaps  func to insure an overlap pair reads shoud be counted as 1 depth unit not 2 units. The depth result will be same with picard CollectWgsMetrics and samtools mpileup.


TuBieJun:
> This change should be accompanied by a command line option (probably a long one, e.g. `--nodup`). Also, you should verify the return value of `bam_mplp_init_overlaps` and exit with an error message, in case of error.

I have add the option.But I still think an overlap pair read that cover a pos must be count once. This is recognized. GATK, picard even samtools mpileup follow this rule.

jkbonfield:
> I have add the option.But I still think an overlap pair read that cover a pos must be count once. This is recognized. GATK, picard even samtools mpileup follow this rule.

It would also cause pipelines to no longer be able to reproduce the results from an earlier run.  Changes like this need to be optional therefore, possibly with the default behaviour being no change.

As for what is best to do, agreed template depth is usually the right thing, but sometimes people also want actual read depth.

valeriuo:
Merged into develop as aa944e9faba856fea1fcc1d6fcaab7c5809298de to 3156beac615754e6f8811b0aec49539aae5875a3

PR #1247, opened 2020-05-13
Add filter and fail marking options.

whitwham:
Filter or fail by minimum match length option.
Write filtered reads to a file.
Filter out QCFAIL or unmapped reads.


whitwham:
Not convinced I've done the right thing in terms of determining useful length of query.  I've used bam_cigar2qlen method but ignored any soft clips.

jkbonfield:
How is `fail_len` meant to be defined - ie what would you document it to mean in the man page?  

The length of the query minus primers?  If so then yes your approach of qlen - SOFT_CLIP seems correct.

The number of query bases that have a corresponding base in the reference?  If so then we're excluding insertion too and leaving just the "match" ops (M, =, X).

whitwham:
> The length of the query minus primers? If so then yes your approach of qlen - SOFT_CLIP seems correct.

Pretty much.  More exactly it would be length of query after any clipping has been done.

PR #1248, opened 2020-05-14
[manpage, minor] Mention BAI chromosome length limit

jmarshall:
On a [bfx.SE answer](https://bioinformatics.stackexchange.com/a/13283/134) noting that BAI indexes cannot handle chromosomes longer than 2<sup>29</sup>, someone pseudonymously asks

> Do you have a reference for this limitation? Is it mentioned in the docs somewhere?

This is common knowledge in the field of course and is spelt out in e.g. SAMv1.pdf. But I didn't see an explicit mention in the samtools man pages, so this adds it to `man samtools index`.

### DIFF
--- a/doc/samtools-index.1
+++ b/doc/samtools-index.1
@@ -82,6 +82,11 @@ or
 .IB aln.sam.gz .csi
 will be created, depending on the index format selected.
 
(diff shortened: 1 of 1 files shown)


daviesrob:
Thanks, at least this gives us something to point people to.  It's only taken about six years :smile:

PR #1250, opened 2020-05-19
[ampliconstats] Prevent buffer overflow

valeriuo:
Prevent a couple of buffer overflows that might occur in `dump_stats` when the amplicon length is larger than expected.


jkbonfield:
Did you manage to get some data to trigger a bug here?

I did test by deliberately tweaking the maximum amplicon length lower and validating that the code reported an error and didn't buffer overrun.  This is managed by the check in `bed2amplicon` when parsing the bed file, so in theory it shouldn't be able to get to the code you changed.

That said, it certainly doesn't harm to have extra checking.

valeriuo:
Yes, I did. By mistake, actually. I used the command on non-amplicon data.

PR #1252, opened 2020-05-20
[depth] Display 0 coverage on files with no reads

### LINKED ISSUE #1249: Samtools depth -aa doesn't report anything when BAM file only has header
Hi there, 

Unless I am wrong, samtools depth doesn't generate an output file when it is given a BAM that only contains a header (i.e. no reads were successfully mapped). Bedtools genomecov does I believe, and although I could use bedtools in the circumstances when I fail to map anything, it would be much neater for me if samtools depth -aa could do the same. 

Is there a reason for this behaviour?

valeriuo:
Display a coverage column of zeros for the entire reference length, when `samtools depth -aa` is run on a file with no alignments.

Fixes #1249


bam2depth.c
```diff
@@ -365,6 +365,21 @@ int main_depth(int argc, char *argv[])
         }
     }
 
+    if (!ret && pos < 0 && all > 1 && !reg) {
```
daviesrob, line 368: While this works fine, could it be merged with the "Handle terminating region" part above to avoid some duplication?  It should be possible to get it to produce the desired output with something like this just before the while loop:
```c
if (pos < 0 && last_tid < 0 && !reg) {
   last_tid = 0;
}
```

test/mpileup/depth.reg
```diff
@@ -60,6 +60,10 @@ P d5_b3aa.out  $samtools depth -aa -b xx.bed3 xx#depth3.bam
 # Fail as -a shouldn't output xn:16-20 (it doesn't with -r)
 F d5_b3a.out   $samtools depth -a  -b xx.bed3 xx#depth3.bam
 
+# All reference positions in a file with no alignments
+INIT x $samtools view -H -b -o noreads.bam xx#depth1.sam
+P d5_all4.out $samtools depth -aa noreads.bam
```
daviesrob, line 65: noreads.bam should be added to `.gitignore`.

jmarshall: Even better, if you call it e.g. `xx#depth-empty.bam` (or anything matching either `xx#depth*.bam*` or `*.tmp`) it'll already be ignored and also removed by `make testclean`.

bam2depth.c
valeriuo: Yes, it's a more elegant solution.

PR #1265, opened 2020-06-08
Add -u option to various samtools commands.

jkbonfield:
This changes `fixmate`, `addreplacerg`, `markdup`, `ampliconclip`, and `sort`.  Unchanged is `split` as it already uses -u for something else.

### DIFF
--- a/bam_addrprg.c
+++ b/bam_addrprg.c
@@ -51,6 +51,7 @@ struct parsed_opts {
     rg_mode mode;
     sam_global_args ga;
     htsThreadPool p;
+    int uncompressed;
 };
 
 struct state;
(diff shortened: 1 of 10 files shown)


jmarshall:
These commands all already have a `-O`/`--output-fmt` option. What about instead parsing `-Oubam` etc in `hts_parse_format()` to get the same effect without wasting an option letter?

jkbonfield:
It already supports `-O bam,level=0`, but I was wanting something brief that is already in use by many tools.  So far `calmd`, `collate`, `merge`, `view` and `depad` already use this option.

Adding ubam as a format (and suffix perhaps) is a reasonable middle ground though, and irrespective of whether we wish to keep -u I think it would be a useful addition.  (It's another repository though so not this PR.)

tseemann:
For those like me reading this and not knowing what `-u` is for, it means "write uncompressed output".

PR #1268, opened 2020-06-15
Ampliconclip Tag Handling Options

whitwham:
Two new options for samtools ampliconclip:

`--original` adds an OA tag with the unclipped values.
`--delete-tag` removes the (potentially) no longer valid NM and MD tags.


whitwham:
On review, I am changing `--delete-tag` to `--keep-tag` and removing the NM and MD tags by default.

bam_ampliconclip.c
```diff
@@ -526,6 +528,49 @@ static inline void swap_bams(bam1_t **a, bam1_t **b) {
...
+    char strand;
+    uint8_t *nm_tag, *old_oa_tag;
+    uint32_t *cigar;
+    int64_t nm = 0;
+    int i;
+
+    ks_clear(oa_tag);
+
+    if ((old_oa_tag = bam_aux_get(orig, "OA"))) {
+        ksprintf(oa_tag, "%s", bam_aux2Z(old_oa_tag));
```
daviesrob, line 542: Strictly, there should be error checking on these kstring functions.

bam_ampliconclip.c
```diff
@@ -526,6 +528,49 @@ static inline void swap_bams(bam1_t **a, bam1_t **b) {
...
+    }
+
+    if (nm_tag) {
+        ksprintf(oa_tag, ",%d,%"PRId64";", orig->core.qual, nm);
+    } else {
+        ksprintf(oa_tag, "%d,;", orig->core.qual);
+    }
+
+    if (old_oa_tag)
+        bam_aux_del(orig, old_oa_tag);
```
daviesrob, line 568: I think it's too early to call `bam_aux_del()` here.  The new OA tag will only be added if some clipping happens, so removong the old one at this point makes it disappear in the case where nothing was clipped.  You can see this in action by running ampliconclip --original on the `test/ampliconclip/1_original_tag.expected.sam` file.

bam_ampliconclip.c
```diff
@@ -609,15 +658,36 @@ static int bam_clip(samFile *in, samFile *out, samFile *reject, char *bedfile,
...
                         f_count++;
                     }
 
+                    if (param->oa_tag) {
+                        uint8_t *tag;
+
+                        if ((tag = bam_aux_get(b, "OA")))
+                            bam_aux_del(b, tag);
+
+                        bam_aux_append(b, "OA", 'Z', strlen(oat.s) + 1, (uint8_t *)oat.s);
```
daviesrob, line 678: This could be done more easily by calling `bam_aux_update_str(b, "OA", oat.l + 1, oat.s)`.  It can handle both replacing an existing tag and adding a new one.

There should also be a check to ensure it worked.

bam_ampliconclip.c
whitwham: It is not something we normally do, not in samtools, not in htslib.

bam_ampliconclip.c
whitwham: Yes, good catch.  That was leftover from an earlier iteration.

bam_ampliconclip.c
whitwham: There is a `bam_aux_update_str`?  Excellent, I will use that instead.

bam_ampliconclip.c
daviesrob: I'm trying to encourage it, e.g. b2d1b9ff.  Ignoring kstring function return values can lead to silent failures so it's best to check them.

daviesrob: You can of course do:
```
int e = 0;
e |= ksprintf(...) < 0;
e |= kputw(...) < 0;
// ... etc
if (e) return -1;
```
to reduce the number of branches needed.

daviesrob:
Squashed and rebased...

PR #1274, opened 2020-06-23
[stats] Add new GCT section

### LINKED ISSUE #1270: samtools stats ACGT content bug
There seems to be a bug in the **ACGT content** reported by `samtools stats` which is apparently due to faulty processing of reverse-strand alignments. One way to reveal the problem is to align a single sequence to multiple places in the mouse genome using bowtie and run samtools stats on the output:

```
bowtie -a -v 0 -S mm10 -c AGTTCTTTGGGAACC | samtools stats | grep ^GCC | cut -f 2-
# reads processed: 1
# reads with at least one reported alignment: 1 (100.00%)
# reads that failed to align: 0 (0.00%)
Reported 16 alignments
1	68.75	0.00	0.00	31.25	0.00	0.00
2	0.00	31.25	68.75	0.00	0.00	0.00
3	31.25	0.00	0.00	68.75	0.00	0.00
4	31.25	0.00	0.00	68.75	0.00	0.00
5	0.00	68.75	31.25	0.00	0.00	0.00
6	31.25	0.00	0.00	68.75	0.00	0.00
7	31.25	0.00	0.00	68.75	0.00	0.00
8	31.25	0.00	0.00	68.75	0.00	0.00
9	0.00	31.25	68.75	0.00	0.00	0.00
10	0.00	31.25	68.75	0.00	0.00	0.00
11	0.00	31.25	68.75	0.00	0.00	0.00
12	68.75	0.00	0.00	31.25	0.00	0.00
13	68.75	0.00	0.00	31.25	0.00	0.00
14	0.00	68.75	31.25	0.00	0.00	0.00
15	0.00	68.75	31.25	0.00	0.00	0.00
```

Tested with SAMtools 1.10 and 1.9 and 1.5.

valeriuo:
The GCC section of `samtools stats` shows the distribution of nucleotides along read cycles. However, the initial decision was to count the bases as they appear in the SAM output, even if they reside on reverse complemented reads.
The new GCT section gives the true base percentages, by complementing bases on reverse complemented reads.

Fixes #1270


stats.c
```diff
@@ -1612,7 +1621,18 @@ void output_stats(FILE *to, stats_t *stats, int sparse)
                 100.*(acgtno_count_1st->t + acgtno_count_2nd->t)/acgt_sum,
                 100.*(acgtno_count_1st->n + acgtno_count_2nd->n)/acgt_sum,
                 100.*(acgtno_count_1st->other + acgtno_count_2nd->other)/acgt_sum);
-
+    }
+    fprintf(to, "# ACGT reverse complemented content per cycle. Use `grep ^GCT | cut -f 2-` to extract this part. The columns are: cycle; A,C,G,T base counts as a percentage of all A/C/G/T bases [%%]\n");
```
daviesrob, line 1625: `ACGT reverse complemented content per cycle` implies the data is always reverse complemented.  Maybe something like `ACGT content per cycle, read oriented` or `Read-oriented per-cycle ACGT content` would be better?

daviesrob:
I've added a minor change to the documentation.  Does it look OK?

valeriuo:
Yes, clearer in a few aspects.

PR #1281, opened 2020-07-03
Restore long-standing mpileup optional column ordering (broken in 1.10 release)

jmarshall:
`samtools mpileup -s -O` / `--output-MQ --output-BP` produces a different column ordering in samtools 1.10 than in previous samtools releases, when these two optional columns are both selected. (And similarly when the combination `--output-MQ --output-QNAME` is selected.)

PR #1073 (which landed in 1.10) moved the MAPQ column to its natural position amongst the other new --output-extra SAM columns, and hence inadvertently changed the ordering of these two previously-existing optional columns from that of previous samtools releases. See [this report of the ensuing incompatibility](https://twitter.com/CFlensburg/status/1278857379400966145); hat tip @ChristofferFlensburg.

This PR restores the documented expected output format by making `--output-MQ` and
`--output-extra MAPQ` control two separate mapping quality columns, each appearing in its expected position in the column ordering.

For user flexibility, it also changes the format produced by the new-in-1.10 `--output-extra MAPQ` to show comma-separated numeric MAPQs, like the other --output-extra columns, instead of a string of ASCII-encoded characters. (This is a change from the behaviour in 1.10 for this new-in-1.10 option, so this part of the PR could be omitted if that was preferred. OTOH PR #1279 is a recent precedent for changing infelicitous new-in-1.10 options.)

Also clarify the mpileup format documentation's description of all these `--output-extra` columns and their ordering, and restore the previous _a17.out_ expected `mpileup -sO` output.


jmarshall:
@daviesrob et al: Any thoughts on this one? It could be argued that this deserves a samtools-1.10.1 release…

valeriuo:
It will most likely go in the next release, due to come out in a few weeks. I'll get on with reviewing it.

valeriuo:
Merged as 15c857747ed71abec69ab9a0d04ba911ba9a4ffd. Thanks for the improvements.

ChristofferFlensburg:
Thanks for looking into this.

So if I need to parse samtools mpileup output as part of my package with unknown user version of samtools, I need to check `samtools --version` and if 1.10 the columns will be reversed, but 1.10.1 or any other version will have the 1.9 column order. Is that correct?

valeriuo:
Yes. 1.10 will be the only version with the flipped order.

jmarshall:
Your options would be:

1. Simply declare samtools 1.10 dead to you in your documentation. (This will be a more tenable position once a corrected update has been released.)

2. Check `samtools --version-only` (which will be easier to parse than `--version`). This will pick up users using the 1.10 release, but not those who built in-development versions between 1.9-90-g3247bfc and 1.10-84-g8796aa5, which is basically impossible to check for. This might not be a major concern.

3. At positions covered by _n_ ≥ 2 reads, the `-s` column has _n_ characters while the `-O` column has at least 2 _n_ - 1 characters. So you could detect when the 7th column is longer than the 8th column (or the 8th column is not parseable because it's not just commas and digits) and abort with a message to the effect of _Bad mpileup format detected — don't use samtools 1.10_ (on the basis that you may have already incorrectly processed some positions covered by only 1 read).

4. Detect as per (3), and swap the columns — on the basis that single-read-covered positions don't affect your output (much) for some reason.

5. Construct a simple one or two line SAM file with known mpileup output by hand, run `samtools mpileup foo.sam` instead of `samtools --version-only`, and use the output to detect which flavour — correct, or broken 1.10-style — of mpileup output the user's samtools has.

Personally I'd do 1 & 3.

ChristofferFlensburg:
Sounds good!

I am going with 2. I'll just check for the exact output of `samtools --version` of 1.10 and swap columns if exactly that. I'm not sure `--version-only` exists in the older 0.X versions, and this works. Any other output and I use standard order.

Thanks for quick reactions and a great tool!

jmarshall:
The ancient 0.1.x versions did not support `samtools --version` either. The two options were added together in c4d398b9a49e1c86b327e7f2e0e28738e5fbfe89 during development of 1.0.

Whichever of the two options you use, 0.1.x's “unrecognized command” output (on standard error though) does not match the 1.10 output you're looking for — so both are equally usable for this task.

PR #1283, opened 2020-07-09
Add options to specifiy index location.

### LINKED ISSUE #983: Provide `--output-idx` and `--input-idx` for index output location

The current `--output` option specifies the location of an output fasta file. There is no corresponding option for specifying where the index file should be placed. That means that it's always placed in the same directory as the input fasta. 

See https://github.com/samtools/samtools/issues/970#issuecomment-439950989

It would be helpful to have `--output-idx` and corresponding `--input-idx` option to specify where the to either store or load the index file, respectively.

whitwham:
Options added ---fai-idx and --gzi-idx.

These are companion options to --output, specifying the locations of the .fai and (if needed) the .gzi index files.

As a convenience in indexing mode (so when no region file is specified) --output will be treated the same as --fai-idx.

This closes #983.

### DIFF
--- a/doc/samtools-faidx.1
+++ b/doc/samtools-faidx.1
@@ -3,7 +3,7 @@
 .SH NAME
 samtools faidx \- indexes or queries regions from a fasta file
 .\"
-.\" Copyright (C) 2008-2011, 2013-2018 Genome Research Ltd.
+.\" Copyright (C) 2008-2011, 2013-2018, 2020 Genome Research Ltd.
 .\" Portions copyright (C) 2010, 2011 Broad Institute.
 .\"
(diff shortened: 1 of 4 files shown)


daviesrob:
`compressed index` might be a bit ambiguous, as it could be a compressed version for the `.fai` file.  Would it be better to say `compressed file index` , or maybe `BGZF index`?

whitwham:
> 
> 
> `compressed index` might be a bit ambiguous, as it could be a compressed version for the `.fai` file. Would it be better to say `compressed file index` , or maybe `BGZF index`?

Maybe.  I like 'compressed file index' the better of the two.

PR #1296, opened 2020-07-31
Improve samtools view error reporting.

jkbonfield:
"[main_samview] truncated file" is a staggeringly unhelpful and
misleading catch-all for something went wrong.

Knowing precisely what isn't particularly trivial, but the next best
thing is reporting the system errno as it may at least point the
finger of blame somewhere sensible.


daviesrob:
It would be better to change the error printing code to use `print_error_errno()`, which is already used elsewhere.  At the same time we could change the misleading "truncated file" message to something better, e.g.:

```
print_error_errno("view", "error reading read \"%s\"", fn_in);
```

daviesrob:
Merged as 90a8ca43

PR #1298, opened 2020-08-17
updating plot-ampliconstats to create index.html

FredDodd6:
plot-ampliconstats now creates a index.html file to display plots.

### DIFF
--- a/misc/plot-ampliconstats
+++ b/misc/plot-ampliconstats
@@ -94,7 +94,6 @@ if ($1 < 5) {
 #-----------------------------------------------------------------------------
 # Load plot meta-data
 # This allows us to auto-scale plots.
-
 my $namp = 0;
 my $nfile = 0;
 my $ref_len = 0;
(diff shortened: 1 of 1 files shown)


misc/plot-ampliconstats
```diff
@@ -1250,3 +1249,97 @@ END
...
+    padding: 5px;
+    border: 1px solid #000;
+    background-color: #e5e5e5;
+}
+.thumbnail:hover span  /*CSS for enlarged image on hover*/
+{
+    visibility: visible;
+    left: 550px;
+    top: 10px;
+}
```
daviesrob, line 1281: This bit of CSS looks redundant at the moment.  It looks like it got copied over from `plot-bamstats`.  Did you want to be able to see a bigger version when hovering over the thumbnail?

To get the feature to work, you would need to add a `<span>` containing the full-size version of each image next to the thumbnail, something like this:
```
print $fh qq[<td><a class="thumbnail" href="$imgs[$i]"><img src="$imgs[$i]" width="200px"><span><img src="$imgs[$i]"></span></a></td>\n];
```
You might also want to look into where the bigger image pops up.  At the moment it would obscure some of the thumbnails, making them difficult to select.

If you don't want the bigger image on hover (and don't want to add it later) it would be good to remove these lines to avoid confusion.

misc/plot-ampliconstats
```diff
@@ -1250,3 +1249,97 @@ END
...
+    
+	if ( $i>0 ) { print $fh "</tr>\n"; }
+	if ($new_section) {
+	    print $fh "<tr><td>$fn_prefix</td></tr>\n";
+	    $new_section = 0;
+	}
+	print $fh "<tr>";
+	print $fh qq[<td><a class="thumbnail" href="$imgs[$i]"><img src="$imgs[$i]" width="200px"></a></td>\n];
+    } else {
+	print $fh qq[<td><a class="thumbnail" href="$imgs[$i]"><img src="$imgs[$i]" width="200px"></a></td>\n];
```
daviesrob, line 1335: And while I'm looking at them, I noticed these lines are identical so they could be replaced with a single copy outside the if statement.

FredDodd6: These seem to have been resolved in #1299 .

daviesrob: Ah yes, I see they have been - although the replacement still has two copies, it's just that they're very subtly different.

misc/plot-ampliconstats
daviesrob: On discussing with @jkbonfield we decided that the behaviour of popping up full-sized images wasn't going to work well for `plot-ampliconstats` output, so I've added a commit to remove this bit of CSS.  I've also got rid of the `.pad` and `.right` sections which weren't referenced anywhere.

I'll squash the commits together on merge and rebase #1299 on top.

daviesrob:
To get your local copy of develop back into sync, you should be able to do something like this:
```
git fetch origin develop
git checkout -B develop origin/develop
```

Note that this assumes you called remote `https://github.com/samtools/samtools.git` `origin`.  If you used a different name you'll have to substitute that in the commands above.

You can then sort out your fork on github by deleting the `FredDodd6/develop` branch and pushing it back from your now-correct local checkout:

```
git push FredDodd6 develop
```

PR #1299, opened 2020-08-17
Ampliconstats index

jkbonfield:
Builds on top of @FredDodd6's PR #1298 with a few additional tweaks:

- Renamed fn_prefix to fn_base as we use prefix already and it's confusing.
    
- Added the ability to create .thumb.png images via -thumbnails option, although this needs ImageMagick convert too.
    
- Thumbnail (visible) size in index.html is now configurable via --thumb_size.  Defaults at 200 pixels.
    
- Hidden option of -index-only; used mainly for debugging so we can keep regenerating index.html without running the rest of the script.
    
- Index.html is generated in the same directory as the images themselves, via the global $prefix.
    
          plot-ampliconstats run1234 ampstats.txt
    
will create run1234-*.png and index.html in current dir, while
    
          plot-ampliconstats /tmp/output/run1234 ampstats.txt
    
will create /tmp/output/run1234*.png and /tmp/output/index.html.
    
- Filenames referred to in index.html will always be relative to that path, so we can move the directory without breaking the links.

- Add -long-opt aliases to -long_opt for consistency with Unix norms.  -long-opt are the ones listed in the help now.

- Removed tabs.

### DIFF
--- a/misc/plot-ampliconstats
+++ b/misc/plot-ampliconstats
@@ -42,11 +42,14 @@ my $res = GetOptions("size=s"      => \$opts{size},
                      "help"        => \$opts{help},
                      "page=i"      => \$opts{page},
                      "amp_add=i"   => \$opts{amp_add},
+                     "amp-add=i"   => \$opts{amp_add},
                      "orient=s"    => \$opts{orient},
                      "depth_max=s" => \$opts{depth_max},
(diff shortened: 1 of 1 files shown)


daviesrob:
Rebased.  Sorry, not quite as cleanly as I'd hoped due to the commit in common with #1298

PR #1300, opened 2020-08-21
Explicitly document that sort -n is incompatible with indexing

jmarshall:
Neophytes getting mysterious errors from `samtools index` because they prepared their BAM files with `samtools sort -n` is somehow in vogue at the moment: see [Q1](https://bioinformatics.stackexchange.com/questions/13997/samtools-index-chromosome-blocks-not-continuous), [Q2](https://bioinformatics.stackexchange.com/questions/14179/no-coor-reads-not-in-a-single-block-at-the-end-0-1), and also [Q3](https://bioinformatics.stackexchange.com/questions/5352/error-given-while-trying-to-index-a-bam-file-with-samtools-index-no-coor).

This spells out in the documentation that for use with `samtools index`, you just want the default sort.

Now that there is a header API, it would be possible to emit a warning (at least) if the input has an `@HD SO` value other than `coordinate` or `unknown`, but that all happens under the covers in `sam_index_build3()` so would be a separate HTSlib PR.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -2599,7 +2599,7 @@ static void sort_usage(FILE *fp)
 "  -m INT     Set maximum memory per thread; suffix K/M/G recognized [768M]\n"
 "  -M         Use minimiser for clustering unaligned/unplaced reads\n"
 "  -K INT     Kmer size to use for minimiser [20]\n"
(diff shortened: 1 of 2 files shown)


daviesrob:
Thanks.  Documentation updates added.

PR #1302, opened 2020-08-27
Merge mem bug

jkbonfield:
Fix a memory overflow in samtools merge.
    
The htslib ks_resize change exosed a bug in samtools merge where copying over a new transformed ID failed to add room for the nul string termination byte, but this was included however in the subsequent memmove.


daviesrob:
Removed stray commit.

PR #1308, opened 2020-09-04
Stop calling unneeded code in markdup.

whitwham:
The code that corrects tags when multiple duplicates of single read occur should only be called when tagging is needed.


whitwham:
The code that does the correcting is very slow when faced with large numbers of duplicates.  I aim to replace it entirely but this is a quick fix for the person who reported the bug.

PR #1313, opened 2020-09-16
Seq cache populate help

jkbonfield:
Two commits:

- One to make seq_cache_populate.pl report the associated REF_CACHE setting to utilise its output.

- Plus another to explain how REF_CACHE is interpreted in the absence of any explicit %s rules.

### DIFF
--- a/doc/samtools.1
+++ b/doc/samtools.1
@@ -944,26 +944,28 @@ temporary directory if no home directory is found) will be used similarly.
 
 .TP
 .B REF_CACHE
-This can be defined to a single directory housing a local cache of
+This can be defined to a single location housing a local cache of
 references.  Upon downloading a reference it will be stored in the
-location pointed to by REF_CACHE.  When reading a reference it will be
-looked for in this directory before searching REF_PATH.  To avoid many
-files being stored in the same directory, a pathname may be
-constructed using %\fInum\fRs and %s notation, consuming \fInum\fR
-characters of the MD5sum.  For example
-\fB/local/ref_cache/%2s/%2s/%s\fR will create 2 nested subdirectories
-with the filenames in the deepest directory being the last 28
-characters of the md5sum.
+location pointed to by REF_CACHE.  REF_CACHE will be searched
+before attempting to load via the REF_PATH search list.  If no
+REF_PATH is defined, both REF_PATH and REF_CACHE will be automatically
+set (see above), but if REF_PATH is defined and REF_CACHE not then no
+local cache is used.
 
-The REF_CACHE directory will be searched for before attempting to load
-via the REF_PATH search list.  If no REF_PATH is defined, both
-REF_PATH and REF_CACHE will be automatically set (see above), but if
-REF_PATH is defined and REF_CACHE not then no local cache is used.
+To avoid many files being stored in the same directory, REF_CACHE may
+be defined as a pattern using \fB%\fR\fInum\fR\fBs\fR to consume \fInum\fR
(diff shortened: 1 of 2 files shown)


doc/samtools.1
```diff
@@ -953,7 +953,9 @@ constructed using %\fInum\fRs and %s notation, consuming \fInum\fR
 characters of the MD5sum.  For example
 \fB/local/ref_cache/%2s/%2s/%s\fR will create 2 nested subdirectories
 with the filenames in the deepest directory being the last 28
-characters of the md5sum.
+characters of the md5sum.  If no \fB%s\fR rules are present in the
+REF_CACHE value it will it assumed to end with an implicit \fB%s\fR, ie
```
jmarshall, line 957:
```suggestion
REF_CACHE value it will be assumed to end with an implicit \fB%s\fR, i.e.
```

Also this is not quite true in two ways: there's a bunch of code about adding a `/` if necessary, so it usually gets an implicit `/%s` appended; this happens if there's some `%2s` rules but no all-remaining `%s` rule. So maybe write as “If no without-digits `%s` rules” (or something…) and “using no additional subdirectories” (also no hyphen in subdirectories).

jkbonfield: Thanks, how about this?

jmarshall: The (other) scenario I was thinking would be confusing is when I set `REF_CACHE=foo/bar` then “using no subdirectories” is confusing because it seems to say it'll undo my `bar`. The new addition sadly makes the “ie…” harder to parse…

It's pretty hard for users to know what we're trying to communicate to them with this :smile:

In the context of the whole paragraph, what about a different approach (probably as its own paragraph right after this one)? 

> If you have set up a nested tree structure (e.g. with _misc/seq_cache_populate.pl_), you will need to use …/%2s/… correspondingly in `REF_PATH` and/or `REF_CACHE` so that tools will be able to use your tree structure correctly.

jkbonfield: Well the "using no subdirectories" does immediately follow the sentence with the %2s/%2s/%s and "2 nested subdirectories", so I was assuming the reader could put 2 and 2 together. :-)

However I'll take another bash at it.

jkbonfield: Third try.  This needed a much bigger restructuring to avoid using seq_cache_populate in an example before it's described.  I also tried to clarify a few other minor bits.  I kept the bit about an implicit trailing /%s as it does need documenting somewhere.

doc/samtools.1
```diff
@@ -944,26 +944,29 @@ temporary directory if no home directory is found) will be used similarly.
...
-\fB/local/ref_cache/%2s/%2s/%s\fR will create 2 nested subdirectories
-with the filenames in the deepest directory being the last 28
-characters of the md5sum.
-
-The REF_CACHE directory will be searched for before attempting to load
-via the REF_PATH search list.  If no REF_PATH is defined, both
-REF_PATH and REF_CACHE will be automatically set (see above), but if
-REF_PATH is defined and REF_CACHE not then no local cache is used.
+location pointed to by REF_CACHE.  REF_CACHE will be searched in
+befo
```
daviesrob, line 950: Small typo here.  Also I think removing the "in" reads better.

jkbonfield: Not sure how I ended up with that.  Probably stray window cursor at some point. :)  Thanks.

I was umming and ahhing over the "in" bit when I wrote it.  Deleted it now.

doc/samtools.1
jmarshall: :+1:

PR #1314, opened 2020-09-21
A fasta/fastq sanitizer that redacts illegal chars.

jkbonfield:
The SAM and VCF specs ban certain characters in reference names.  Trying to work around this after alignment is problematic for numerous reasons, so it is preferable to rename your reference sequences before alignment.

This tool does this.

I'm not sure on the -t bit to stop at the tab instead of first white-space.  With hindsight I think this is a misunderstanding from the test data we were given, which had already been modified to remove the spaces so they didn't need to be processed afterall. (Thank goodness)

I can remove that if needed, or just leave it as a hidden option.  The script is pretty naive, but seems to work and it's totally standalone.

### DIFF
--- a/misc/fasta-sanitize.pl
+++ b/misc/fasta-sanitize.pl
@@ -23,11 +23,8 @@
 # DEALINGS IN THE SOFTWARE.
 
 
-# Usage: fasta-sanitize.pl [-t] file.in > file.out
-# Or via stdin, e.g. zcat file.in.gz | fasta-sanitize.pl [-t] > file.out
-#
-# Option -t ends names on a tab or newline, which permits spaces in names.
-# Otherwise names are considered to end on any whitespace.
+# Usage: fasta-sanitize.pl file.in > file.out
+# Or via stdin, e.g. zcat file.in.gz | fasta-sanitize.pl > file.out
 #
 # Also supports and autodetects fastq.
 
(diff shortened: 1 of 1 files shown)


misc/fasta-sanitize.pl
```diff
@@ -0,0 +1,105 @@
...
+my $fastq = 0;
+my $in_qual = 0;
+my $seq_len = 0;
+
+my $name_whitespace = qr/^([>@])\s*(\S*)(.*)/;
+my $name_tab = qr/^([>@])\s*([^\t]*)(.*)/;
+
+my $name_re = $name_whitespace;
+if ($ARGV[0] eq "-t") {
+    pop(@ARGV);
```
daviesrob, line 64: This doesn't work:
```
$ ./misc/fasta-sanitize.pl -t /tmp/test.fa
Can't open -t: No such file or directory at ./misc/fasta-sanitize.pl line 68.
```

I think `pop` should be `shift`.

misc/fasta-sanitize.pl
```diff
@@ -0,0 +1,105 @@
...
+        # FASTQ separator between seq and qual
+        if ($fastq && /^\+/) {
+            print;
+            $in_qual = 1;
+            next;
+        }
+
+        # Seq
+        print;
+        $seq_len += length($_);
```
daviesrob, line 99: `$_` will include any new-lines, so end-of-qual detection could be defeated in the event that the seq and qual parts of a fastq file are wrapped at different line lengths.  It would be safer to call `chomp` to remove the line ending from the length calculation.

misc/fasta-sanitize.pl
jkbonfield: Oops, one of those last minute configuration tweaks that I clearly didn't test!

I just ditched it entirely as it was added due to user test data having spaces in names, but it later transpired it broke everything else so they'd already ran a sed script to remove all of those anyway.  An ill conceived attempt to work around something that's not an issue.

misc/fasta-sanitize.pl
jkbonfield: That's a sneaky attack method!

Fixed. :-)

daviesrob:
Thanks.  Fixes have been squashed and merged in.  The random appveyor failure was unrelated, and went away after rebasing.

PR #1324, opened 2020-10-13
[view] Added -N to filter by file containing read names

d-cameron:
Also refactored the string khash sets to remove code duplication.

This feature has come up multiple times on biostars and I personally have had to use (the much slower) `picard.FilterSamReads` multiple times purely because this option is missing from samtools.

More generally, what is the threshold for inclusion into samtools? I'm slowly building up a collection of useful but not-quite-as-universally-useful-as-samtools utilities. Should I make PRs for them, or keep them in a specialised utility? For example, my utility that converts just the unmapped bases (ie, unmapped reads, soft clipped bases, and the bases that don't map to any chimeric alignment) to fastq could be a PR to `samtools fastq/fasta`.

### DIFF
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -87,6 +87,7 @@ option is required whenever writing CRAM output.
 The
 .BR -L ,
 .BR -M ,
+.BR -N ,
 .BR -r ,
 .BR -R ,
 .BR -d ,
@@ -253,6 +254,10 @@ regions specified on the command line will be reported multiple times.
 The usage of a BED file is optional and its path has to be preceded by
 .BR -L
 option.
+.TP
+.BI "-N " FILE
+Output only alignments with read names listed in
+.I FILE.
 .TP
 .BI "-r " STR
 Output alignments in read group
--- a/sam_view.c
+++ b/sam_view.c
@@ -41,16 +41,15 @@ DEALINGS IN THE SOFTWARE.  */
 #include "sam_opts.h"
 #include "bedidx.h"
 
-KHASH_SET_INIT_STR(rg)
-KHASH_SET_INIT_STR(tv)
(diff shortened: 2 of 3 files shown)


daviesrob:
Looks good thanks, especially removing the code duplication.  I've taken the liberty of pushing a commit to add the option into the man page, hopefully that looks alright.  If it is, I'll squash everything together and merge it.

Do let us know if you think you have anything else that might be useful.  There's no hard-and-fast rules about what gets included - I would guess the criteria is how much work there is in porting it into the samtools framework plus ongoing maintenance compared to how many people are likely to want to use it.

d-cameron:
Thanks for the prompted response. It's now rebased into a single commit. Let me know if there's anything else needed for this PR.

daviesrob:
Nothing else needed, and now merged.

Thanks.

mcshane:
Could this be extended to exclude readnames too? I use both sets of logic in different contexts. Perhaps adding a `^` to the front of the `FILE` could indicate exclusion as used in bcftools.

daviesrob:
You can invert the filter at the moment, albeit not completely efficiently, using the `-U` option ("output reads not selected by filters to FILE").  So just use `-U` to output the data you want and `-o /dev/null` to throw away the rest.  E.g.:
```
samtools view -N names -b -o /dev/null -U keep.bam in.bam
```

You can also use `-U -` to send the data to stdout.

mcshane:
Ha! Never would have though to try that. Thanks.

fengcong3:
Hi, I would like to ask how is the speed of this function? Is it random access? or traverse the input file

d-cameron:
In-memory hash table for read name filtering then standard samtools view
for the reads. If you filter to a region it'll use the index

PR #1335, opened 2020-11-05
Add Cirrus-CI integration

jkbonfield:


.cirrus.yml
```diff
@@ -0,0 +1,171 @@
+# Note we have a maximum of 16 CPUs available, so adjust our
+# builds so we can start all concurrently without needing to schedule.
+
+# Sadly though there is still a finite limit to MacOS X of one instance.
```
jmarshall, line 4: “macOS” is spelt thus nowadays (small m, no X, no space).

.cirrus.yml
```diff
@@ -0,0 +1,171 @@
...
+
+timeout_in: 10m
+
+#--------------------------------------------------
+# Template: htslib clone & build
+#
+# We try to clone htslib using the same branch name and owner as this
+# samtools branch.  If it exists, it's likely the user is making a
+# joint samtools+htslib PR and wants both checked in unison.
+# Failing that we use samtools/htslib:devel
```
jmarshall, line 15:
```suggestion
# Failing that we use samtools/htslib:develop
```

jkbonfield: Thanks.

Also I was hoping it'd work more like AppVeyor than Travis, but maybe I need a .cirrus dir with an updated script.  Will explore.  This PR is using samtools/htslib develop branch, not the cirrus-ci branch.  For me in my repos it "does the right thing" here.  Will explore more so expect a flurry of commits ;)

jkbonfield: Seems not possible.  Oh well, we had the same problem with Travis.  Env set or this PR:

```
CIRRUS_BASE_BRANCH=develop
CIRRUS_BRANCH=pull/1335
CIRRUS_REPO_FULL_NAME=samtools/samtools
CIRRUS_REPO_NAME=samtools
CIRRUS_REPO_OWNER=samtools
```

Lacking jkbonfield/samtools or my branch of cirrus-ci branch means it has no means to find the matching htslib branch.  In this case it's irrelevant, but often not (eg the two fastq branches).  Shame.  Only choice I could think of for fixing that, if we care, would be grokking the commit messages.

PR #1337, opened 2020-11-09
[cat] Fix the short option for the number of threads

### LINKED ISSUE #1336: Samtools cat option for threads (-@) is invalid
#### Are you using the latest version of [samtools](https://github.com/samtools/samtools/releases/latest) and [HTSlib](https://github.com/samtools/htslib/releases/latest)? 

Using latest release

Linux with gcc 8.2

Running `samtools cat -@ 4 file1.bam file2.bam` fails with 
```
cat: invalid option -- '@'
Segmentation fault
```

Running `samtools cat --threads 4 file1.bam file2.bam` is valid, and behaves as expected with multithreading.

Both options (@/threads) are specified in the documentation for `samtools cat`, and follows the notation for other tools, so hopefully a simple fix.

valeriuo:
`samtools cat` was missing the short option for the number of threads from the option string passed to `getopt_long`. This caused `optarg` to be 0.

Fixes #1336

### DIFF
--- a/bam_cat.c
+++ b/bam_cat.c
@@ -467,7 +467,7 @@ int main_cat(int argc, char *argv[])
     char *outfn = 0;
     char **infns = NULL; // files to concatenate
     int infns_size = 0;
-    int c, ret = 0, no_pg = 0;
+    int c, ret = 0, no_pg = 0, usage = 0;
     samFile *in;
     sam_global_args ga;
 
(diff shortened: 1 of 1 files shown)


jmarshall:
Try `samtools cat -x` (or any other unused option letter), which also crashes.

Also needs an explicit `case '?'` with a fall-through from `default` as done in e.g. _cut_target.c_.

jmarshall:
:+1:

PR #1342, opened 2020-11-18
List --write-index in samtools sort usage [trivial]

jmarshall:
This option was implemented for sort in PR #1148, but in a minor oversight, it was not added to the usage display shown by `samtools sort` without arguments.

(Tests will currently fail because samtools's _test/mpileup/mp_N2.sam_ suffers (on two lines) from the same CIGAR syntax error as one of HTSlib's test files as highlighted in samtools/htslib#1169.)


jmarshall:
(I've verified that all the other subcommands listed in _NEWS_ as supporting this global option do list it in their usages.)

PR #1346, opened 2020-11-25
samtools view filtering and documentation

jkbonfield:
Add filter expressions to samtools.1 man page

Also add -e FILTER to samtools view.

This is a validation of the new `sam_passes_filter` function, meaning it can be used in conjunction with -U to output the reads that do not match the filter.

Infact both can be used together.  For example

    samtools view \
        --input-fmt-option "filter=rlen-qlen >= 10" \
        -e 'ncigar <= 10' in.bam -o out.sam -U outU.sam

will pre-filter to only reads where the length on the reference is at least 10 bases longer than the length of the sequence.  Those that pass are then written to either out.sam or outU.sam depending on whether they have <= 10 cigar operations.

### DIFF
--- a/doc/samtools.1
+++ b/doc/samtools.1
@@ -773,6 +773,11 @@ compression, otherwise SAM defaults to uncompressed.
 Specifies the number of threads to use during encoding and/or
 decoding.  For BAM this will be encoding only.  In CRAM the threads
 are dynamically shared between encoder and decoder.
+.TP
+.BI filter= STRING
+Apply filter STRING to all incoming records, rejecting any that do not
+satisfy the expression.  See the FILTER EXPRESSIONS section below for
+specifics.
 .\" CRAM specific
 .TP
 .BI reference= fasta_file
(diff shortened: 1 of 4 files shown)


sam_view.c
```diff
@@ -465,6 +471,12 @@ int main_samview(int argc, char *argv[])
             break;
         case 'M': settings.multi_region = 1; break;
         case 1: no_pg = 1; break;
+        case 'z':
+            if (!(settings.filter = hts_filter_init(optarg))) {
+                fprintf(stderr, "main_samview: Failed to parse filter\n");
```
jmarshall, line 476: `print_error()`

jmarshall:
What's the reason for choosing `-z`? e.g. `samtools view -e` is available (though `-e` is used in `calmd`, is a no-longer-documented legacy option for `mpileup`, and an undocumented option for `phase` — so perhaps it's that you were aiming for something universally available in samtools?).

sam_view.c
jkbonfield: Heh, I just copied the local style from about 10 lines above (-x), but can change it.

jkbonfield:
To be honest no real reason!  I initially started with -z because so many others were in use that I just went through the alphabet (backwards), but it kind of stuck.  I'm not planning on adding this option elsewhere.  View is different as it's the most commonly used tool and as you pointed out it needs different logic to cope with the -U option anyway.

Indeed -e is a better choice.

sam_view.c
jmarshall: 16 `/fprintf.*main_samview/` vs 29 `/print_error/` in _sam_view.c_ — gotta keep that ratio improving one line at a time :smile:

doc/samtools.1
```diff
@@ -756,6 +756,11 @@ compression, otherwise SAM defaults to uncompressed.
 Specifies the number of threads to use during encoding and/or
 decoding.  For BAM this will be encoding only.  In CRAM the threads
 are dynamically shared between encoder and decoder.
+.TP
+.BI filter STRING
```
valeriuo, line 760: `.BI filter= STRING`

doc/samtools.1
```diff
@@ -913,6 +918,133 @@ Look for MD5 via REF_CACHE environment variable.
...
+
+.TS
+l lb l .
+Grouping	(, )	E.g. "(1+2)*3"
+Values:	literals, vars	Numbers, strings and variables
+Unary ops:	+, -, !, ~ 	E.g. -10 +10, !10 (not), ~5 (bit not)
+Math ops:	*, /, %	Multiply, division and (integer) modulo
+Math ops:	+, -	Addition / subtractin
+Bit-wise:	&	Integer AND
+Bit-wise	|	Integer OR
```
valeriuo, line 943: XOR (`^`) has higher precedence than OR (`|`).

valeriuo (changes requested):

sam_view.c
```diff
@@ -465,6 +471,12 @@ int main_samview(int argc, char *argv[])
             break;
         case 'M': settings.multi_region = 1; break;
         case 1: no_pg = 1; break;
+        case 'e':
+            if (!(settings.filter = hts_filter_init(optarg))) {
+                print_error("main_samview", "Failed to parse filter");
```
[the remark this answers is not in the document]
jkbonfield: Actually it's wrong anyway as it doesn't parse it here.  I'll reword.

valeriuo:
Rebased and merged as 1e3f85dc81f301107465c5f036cf090d2852f20f and 9abe06bbff5fa0e4b0cc40aab3ba0f6ef47587c9

PR #1351, opened 2020-12-09
Add ampliconstats to the main samtools man page.

whitwham:
Added the missing command.

### DIFF
--- a/doc/samtools.1
+++ b/doc/samtools.1
@@ -67,6 +67,8 @@ samtools bedcov aln.sorted.bam
 .PP
 samtools depth aln.sorted.bam
 .PP
+samtools ampliconstats primers.bed in.bam
+.PP
 samtools mpileup -C50 -f ref.fasta -r chr3:1,000-2,000 in1.bam in2.bam
 .PP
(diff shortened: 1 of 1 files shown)


whitwham:
Note, the man page on htslib.org needs updating as it is still on 1.10.  After this is merged though.

jkbonfield:
Thanks for this.

I'm in two minds as to where this belongs in the ordering.

It could either be adjacent to ampliconclip, as the two form their own group of amplicon-based analysis, or immediately after stats as part of the overall statistics group (flagstats, idxstats, stats).  I guess  bedcov, depth, mpileup, coverage all form genome coordinate depth style plotting, which ampliconstats may also be said to share a common grouping.  I guess this is why you put it there.

How about stats and bedcov which feels like a natural transition between the two groups? (Which rather fortuitously just happen to be adjacent.)

No strong feelings though so feel free to argue your case. :-)

whitwham:
Yes, I was in two minds about where it should go.  It seemed more stat like so I put it there.  It could well have just been put at the end near ampliconclip.  I have no really strong feelings about where it should go either.

PR #1355, opened 2021-01-06
make the meaning of -s options in samtools depth subcmd more clear

wulj2:
in fact, the ```-s``` option will automatically raise baseQ threshold to 1 internally.
see https://github.com/samtools/samtools/blob/develop/bam2depth.c#L172-L175
so the document of ```-s``` option is somewhat misleading?


valeriuo:
Merged as fe8efa74efdb388b2656909d3cba9bb2fced1960
Thanks for the fix!

PR #1367, opened 2021-01-28
[view] Add support for int value tags to options -d and -D

### LINKED ISSUE #1357: support int tags with samtools view options -d and -D
the -d and -D options of samtools view currently handle only tags of type string, as evidenced by the following code:

https://github.com/samtools/samtools/blob/d58fc8a16729f25407da6729c440a51140396f4c/sam_view.c#L104-L112

It would be nice if those options could also handle tags of type integer.
For example, the following mods allowed me to use option -D with a file containing 1651 integers:
```C
    if (settings->tvhash && settings->tag) {                                     
        uint8_t *s = bam_aux_get(b, settings->tag);                              
        if (s) {                                                                 
            uint8_t t[24];                                                         
            khint_t k;                                                             
            if (*s=='i') { sprintf((char*)t,"%ld",bam_aux2i(s)); s=t; } else ++s;  
            k = kh_get(tv, settings->tvhash, (char*)s);                          
            if (k == kh_end(settings->tvhash)) return 1;                         
        } else {                                                                 
            return 1;                                                            
        }                                                                        
    }                                                                            
``` 
Otherwise, the [man page](http://www.htslib.org/doc/samtools-view.html) should state that the -d and -D options can only handle tags of type string.

valeriuo:
`samtools view` can now filter alignments by tags with integer values, using options **-d** or **-D**.

Fixes #1357

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -106,7 +106,15 @@ static int process_aln(const sam_hdr_t *h, bam1_t *b, samview_settings_t* settin
         uint8_t *s = bam_aux_get(b, settings->tag);
         if (s) {
             if (settings->tvhash) {
-                khint_t k = kh_get(str, settings->tvhash, (char*)(s + 1));
+                char t[32], *val;
+                if (*s == 'i' || *s == 'I' || *s == 'c' || *s == 'C') {
+                    int ret = snprintf(t, 32, "%"PRId64, bam_aux2i(s));
+                    if (ret > 0) val = t;
+                    else return 1;
+                } else {
+                    val = (char *)(s+1);
+                }
+                khint_t k = kh_get(str, settings->tvhash, val);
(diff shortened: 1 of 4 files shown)


sam_view.c
```diff
@@ -106,7 +106,15 @@ static int process_aln(const sam_hdr_t *h, bam1_t *b, samview_settings_t* settin
         uint8_t *s = bam_aux_get(b, settings->tag);
         if (s) {
             if (settings->tvhash) {
-                khint_t k = kh_get(str, settings->tvhash, (char*)(s + 1));
+                char t[32], *val;
+                if (*s == 'i' || *s == 'I' || *s == 'c' || *s == 'C') {
```
jmarshall, line 110: What about `s` and `S`?

valeriuo: Good point. Thanks!

PR #1368, opened 2021-02-03
[release] News update for the next release (1.12)

valeriuo:
NEWS update

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,14 +1,76 @@
 Release a.b
 -----------
 
- * Samtools stats now displays the number of supplementary reads in the
+ * The legacy samtools API (libbam.a, bam.h, sam.h, etc) has not been actively
+   maintained since 2015. It is deprecated and will be removed entirely in a
+   future SAMtools release. We recommend coding against the HTSlib API directly.
+
+ * I/O errors and record parsing errors during the reading of SAM/BAM/CRAM
+   files are now always detected. Thanks to John Marshall (#1379; fixed #101)
+
+ * New make targets have been added: check-all, test-all, distclean-all,
+   mostlyclean-all, testclean-all, which allow SAMtools installations to
+   call corresponding Makefile targets from embedded HTSlib installations.
+
+ * samtools --version now displays a summary of the compilation details and
+   available features, including flags, used libraries and enabled plugins
+   from HTSlib. As an alias, `samtools version` can also be used. (#1371)
+
+ * samtools stats now displays the number of supplementary reads in the
    SN section. Also, supplementary reads are no longer considered when
    splitting read pairs by orientation (inward, outward, other).
    
- * Samtools stats now counts only the filtered alignments that overlap
+ * samtools stats now counts only the filtered alignments that overlap
    target regions, if any are specified.
 
- * Samtools view now supports int value tags for options -d and -D.
+ * samtools view now accepts option -N, which takes a file containing
+   read names of interest. This allows the output of only the reads with
+   names contained in the given file. Thanks to Daniel Cameron. (#1324)
+
+ * samtools view -d option now works without a tag associated value, which
+   allows it to output all the reads with the given tag. (#1339; fixed #1317)
+
+ * samtools view -d and -D options now accept integer values associated with
+   tags, not just strings. Thanks to `@dariome` for the suggestion. (#1357)
+
+ * samtools view now works with the filtering expressions introduced by HTSlib.
+   The filtering expression is passed to the program using the specific option
+   -e or the global long option --input-fmt-option. E.g.
+   `samtools view -e 'qname =~ "#49$" && mrefid != refid && refid != -1 && 
+    mrefid != -1' align.bam` looks for records with query-name ending in `#49`
+   that have their mate aligned in a different chromosome. More details can be
+   found in the FILTER EXPRESSIONS section of the main man page. (#1346)
+
+ * samtools markdup now benefits from an increase in performance in the
+   situation when a single read has tens or hundreds of thousands of duplicates.
+   Thanks to `@denriquez` for reporting the issue. (#1345; fixed #1325)
+
+ * The documentation for samtools ampliconstats has been added to the samtools
+   man page. (#1351)
+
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -1,13 +1,60 @@
...
+
+ * samtools markdup now benefits from an increase in performance in the
+   situation when a single read has tens or hundreds of thousands of duplicates.
+   Thanks to `@denriquez` for reporting the issue. (#1345; fixed #1325)
+
+ * The documentation for samtools ampliconstats has been added to the samtools
+   man page. (#1351)
+
+ * A new FASTA/FASTQ sanitizer script (`fasta-sanitize.pl`) was added, which
+   corrects the illegal characters in the reference names. (#1314) Thanks to
```
jmarshall, line 34: Suggest `invalid` rather than `illegal` (see [the 4th-last paragraph here](https://www.gnu.org/prep/standards/html_node/GNU-Manuals.html)), and the installation fix is too small to mention IMHO :smile:

NEWS
```diff
@@ -1,13 +1,60 @@
...
+ * samtools view -d option now works without a tag associated value, which
+   allows it to output all the reads with the given tag. (#1339; fixed #1317)
+
+ * samtools view -d and -D options now accept integer values associated with
+   tags, not just strings. Thanks to `@dariome` for the suggestion. (#1357)
+
+ * samtools view now works with the filtering expressions introduced by HTSlib.
+   The filtering expression is passed to the program using option -e, e.g.
+   `samtools view -e '!(flag&2048)' align.bam` filters out all the supplementary
+   reads. (#1346)
```
jmarshall, line 24: Possibly mention that other subcommands can do similar with `-Ofilter=expr` or so.

NEWS
```diff
@@ -1,13 +1,60 @@
...
+   out. (#1335; #1365)
+
+ * Updated the Windows image used by Appveyor to 'Visual Studio 2019'.
+  (#1333; fixed #1332)
+
+ * Fixed a bug in samtools cat, which prevented the command from running in
+   multi-threaded mode. Thanks to Alex Leonard for reporting the issue.
+   (#1337; fixed #1336)
+
+ * A couple of invalid CIGAR strings have been removed from the test data.
```
jmarshall, line 51:
```suggestion
 * A couple of invalid CIGAR strings have been corrected in the test data.
```

NEWS
```diff
@@ -1,13 +1,60 @@
...
+
+ * A new FASTA/FASTQ sanitizer script (`fasta-sanitize.pl`) was added, which
+   corrects the illegal characters in the reference names. (#1314) Thanks to
+   John Marshall for the installation fix. (#1353)
+
+ * The CI scripts have been updated to recurse the HTSlib submodules when
+   cloning HTSlib, to accommodate for the CRAM codecs, which now reside in the
+   htscodecs submodule. (#1359)
+
+ * Added Cirrus-CI integration as a replacement for Travis, which was phased
```
jmarshall, line 41: Maybe avoid (any appearance of) editorialising about Travis CI GmbH by just saying something like “The CI integrations now include Cirrus-CI rather than Travis”. Or just delete “which was phased out”…

NEWS
```diff
@@ -1,13 +1,60 @@
-Release a.b
------------
+Release 1.12 (xx February 2021)
```
daviesrob, line 1: Ths line needs to stay as `Release a.b` as the release script looks for it.

NEWS
jkbonfield: This correct syntax would be e.g. `--input-fmt-option 'filter=!(flag&2048)'`.

Also I think this example isn't useful, unless we deliberately with to make the point it's identical to `-F 2048`.  I'd argue it's also better to write it as `!flag.supplementary` anyway as it's self documenting.

IMO a better example is something you cannot do with the existing filter options.  Eg looking for records with query-name ending in `#49` that have a mate aligned in a different chromosome. Ie:

```
samtools view -e 'qname =~ "#49$" && mrefid != refid'
```

jmarshall: Oops, I knew I should have written out the long option (and then might have noticed the thinko). Thanks.

jmarshall:
Apropos #1184 (Is it time to remove the legacy Samtools API?): I have a [draft](https://github.com/jmarshall/samtools/commits/no-API) of removing it, but wasn't planning to make a distracting PR from that branch until after the release. :smile:

I thought we had a note deprecating it in _NEWS_, but the closest I can find is this item from the samtools 1.3 era: “Added legacy API sam.h functions sam_index_load() and samfetch() providing bam_fetch()-style iteration over either BAM or CRAM files.  (In general we recommend recoding against the htslib API directly, but this addition may help existing libbam-using programs to be CRAM-enabled easily.)”

If the maintainers agree with removing it somewhere in the 1.13 to 1.13+n timeframe, perhaps the 1.12 section should add a clear notice near the top of the section, e.g.,

> The legacy samtools API (libbam.a, bam.h, sam.h, etc) has not been actively maintained since 2015. It is deprecated and will be removed entirely in a future SAMtools release. We recommend coding against the HTSlib API directly.

jmarshall:
Thanks for applying the API deprecation note.

Sorry to be endlessly wordsmithing… :smile:

> The error reporting on read failures has been improved for most of the samtools commands.

…but this makes previous existing SAMtools sound worse than it is. Only 9 out of ~30 commands needed fixes, and several of them were obscure commands or minor fixes. Conversely I only audited the reading of the main SAM/BAM/CRAM file(s); there might still be infelicities reading BED files or other minor input files.

If _NEWS_ wants to list exactly which commands users might consider checking their use of, something like this would be a precise description:

```
 * I/O errors and record parsing errors during the reading of SAM/BAM/CRAM
   files are now always detected. Thanks to John Marshall (#1379; fixed #101)
 
   (The bedcov, flagstat, phase, targetcut, and tview commands previously
   treated these errors as EOF; the coverage, depth, fixmate, and mpileup 
   commands previously exited with a failed status but now also print an 
   error message; all other commands already detect these errors and fail 
   with error messages, so needed no changes.)
```
…while admittedly being long and boring…

PR #1370, opened 2021-02-04
[view] Support extra int type encodings for auxiliary tags

valeriuo:
Integer value types of auxiliary tags supported by BAM are: **c** (`int8_t`), **C** (`uint8_t`), **s** (`int16_t`), **S** (`uint16_t`), **i** (`int32_t`), **I** (`uint32_t`).

Related to #1367

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -1,6 +1,6 @@
 /*  sam_view.c -- SAM<->BAM<->CRAM conversion.
 
-    Copyright (C) 2009-2020 Genome Research Ltd.
+    Copyright (C) 2009-2021 Genome Research Ltd.
     Portions copyright (C) 2009, 2011, 2012 Broad Institute.
 
     Author: Heng Li <lh3@sanger.ac.uk>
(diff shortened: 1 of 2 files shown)


jmarshall:
The previous PR also made a `Z` → `[ZiIcC]` change to _test/test.pl_ that might need adjusting too.

This sort of invites the question of what other types can be handled easily enough in this piece of code (`A`? `H`? `f` is a bit harder), but happy to leave that for next time…

valeriuo:
> This sort of invites the question of what other types can be handled easily enough in this piece of code (A? H? f is a bit harder), but happy to leave that for next time…

We'll implement them as requests. I assume `f` would probably be the next in line, in order of usage frequency.

jmarshall:
Sounds good. `f` will be harder to implement, as comparing floating point values by comparing them as formatted strings probably isn't going to fly. (e.g. I write `-d 1e3` but the tag values get formatted as `1000`.)

daviesrob:
Looks OK.  I've added a copyright line update that I'll squash in on merge.

jmarshall:
Looks like you just got a request for `A` :smile: — #1392

valeriuo:
Another one in the basket — #1397

PR #1379, opened 2021-02-17
Ensure all sam_read1() and sam_itr_next() failures are reported

### LINKED ISSUE #101: samtools treats many errors as EOF, silently hiding problems
In the development of Antonie, I generated BAM files that were subtly malformed. This is of course my own bug, and of course samtools should not compensate for my bugs. However, samtools silently accepted my BAM files, and appeared to process them quite well!

Further investigation found that 'samread()' returns negative values for both errors and EOFS, and that many loops within samtools treat all negative values equally. In other words, they turn an error into a normal EOF, which generates no error or warning message.

Any loop like this is problematic: 
    while (samread(sam,bam_line) >= 0) {... }

As an example, http://ds9a.nl/tmp/blah.bam has an invalid sequence id in there, but 'samtools stats blah.bam' processes it without apparent error, but also without producing any statistics beyond the problematic read.

While I of course appreciate the samtools software, I would suggest screaming bloody murder on any kind of unexpected error, lest our users end up with invalid results because part of their data was silently skipped!

Thanks for your attention.

jmarshall:
Ensure that error return codes from all invocations of these functions are captured and result in a diagnostic message and failing exit status. An audit of the 32 instances of `sam_read1()` and the 10 instances of `sam_itr_next()` in code contributing to the `samtools` executable showed that the following were missing and are fixed by this PR:

For coverage, depth, fixmate, mpileup: the command already exited with failure but this adds an error message. (For mpileup, this changes the exit status in this case from 255 (i.e., -1 🤮) to 1.)

For flagstat: convert a "Continue anyway" message into an error and failure.

For bedcov, phase, targetcut: add the missing `bam_[m]plp_auto()` check and error message and failure.

For tview: add the missing `sam_itr_next()` check, and exit immediately (as tview already does for other "can't happen" errors.)

Fixes #101. Fixes a number of instances of #51.

### DIFF
--- a/bedcov.c
+++ b/bedcov.c
@@ -72,7 +72,7 @@ int main_bedcov(int argc, char *argv[])
     kstream_t *ks;
     hts_idx_t **idx;
     aux_t **aux;
-    int *n_plp, dret, i, j, m, n, c, min_mapQ = 0, skip_DN = 0;
+    int *n_plp, dret, i, j, m, n, c, ret, status = 0, min_mapQ = 0, skip_DN = 0;
     int64_t *cnt, *pcov = NULL;;
     const bam_pileup1_t **plp;
     int usage = 0, has_index_file = 0;
@@ -92,7 +92,7 @@ int main_bedcov(int argc, char *argv[])
         case 'g':
             tflags = bam_str2flag(optarg);
             if (tflags < 0 || tflags > ((BAM_FSUPPLEMENTARY << 1) - 1)) {
-                print_error_errno("depth", "Flag value \"%s\" is not supported", optarg);
+                print_error("bedcov", "Flag value \"%s\" is not supported", optarg);
                 return 1;
             }
             flags &= ~tflags;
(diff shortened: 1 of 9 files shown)


bedcov.c
```diff
@@ -226,6 +226,14 @@ int main_bedcov(int argc, char *argv[])
                     if (min_depth >= 0 && pd >= min_depth) pcov[i]++;
                 }
             }
+
+        if (ret < 0) {
+            print_error("bedcov", "error reading from input file");
+            status = 2;
+            bam_mplp_destroy(mplp);
+            continue;
```
daviesrob, line 234: Maybe bail out completely at this point?

daviesrob (commented):
Thanks for the updates, these are certainly long overdue.

Some of them seem to keep the "continue anyway" behaviour, albeit with returning an error code now.  I think it might be time to make them bail out as soon as the error is found, as the output is unlikely to be useful anyway.

jmarshall:
> Some of them seem to keep the "continue anyway" behaviour, albeit with returning an error code now.

IMHO the defining quality of the “continue anyway” messages is that they are reassuring, which these aren't.

However I hadn't tested `mpileup` and `depth` with `-a` and you are quite right that in this case the terminating region code prints masses of incorrect output. Code lifted up as suggested, and the error message still appears **after** all the output hence is noticeable.

For `bedcov`, I was imagining later regions in the BED file might avoid the corrupted parts of the file hence be salvageable. But how would the user know which ones to trust? So I've also adjusted this code as suggested.

Thanks for the review.

daviesrob:
Thanks for the updates, will squash and merge.

PR #1381, opened 2021-02-19
Added a reference to the SAM specification.

### LINKED ISSUE #1376: Documentation on SAM flags could use some fleshing out
FYI, "man samtools-flagstat" says only this about secondary alignments:

```
                         secondary
                                0x100 bit set

```
And "man samtools-flags" says this:

`       0x100   SECONDARY       secondary alignment
`
SAMv1.pdf doesn't really add anything.  Would be good if the PDF and one of the man pages had a clear definition of a secondary alignment as flagged by samtools, as well as features indicated by other flags.  There are definitions to be found online, of course, but someone new to the field can never be sure what to trust.

Cheers,

    Jason

whitwham:
Add a reference in a couple of man pages to the SAM specification.  This is to provide further information on flags that the man pages do not go into.  This should close #1376.

### DIFF
--- a/doc/samtools-flagstat.1
+++ b/doc/samtools-flagstat.1
@@ -52,8 +52,10 @@ Does a full pass through the input file to calculate and print statistics
 to stdout.
 
 Provides counts for each of 13 categories based primarily on bit flags in
-the FLAG field. Information on the meaning of the flags is given in the SAM specification
-document (https://samtools.github.io/hts-specs/SAMv1.pdf).
+the FLAG field.
+Information on the meaning of the flags is given in the SAM specification
+document <https://samtools.github.io/hts-specs/SAMv1.pdf>.
+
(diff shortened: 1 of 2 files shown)


doc/samtools-flagstat.1
```diff
@@ -52,7 +52,8 @@ Does a full pass through the input file to calculate and print statistics
 to stdout.
 
 Provides counts for each of 13 categories based primarily on bit flags in
-the FLAG field.
+the FLAG field. Information on the meaning of the flags is given in the SAM specification
+document (https://samtools.github.io/hts-specs/SAMv1.pdf).
```
daviesrob, line 56: The paragraph break has gone missing here.  I'll fix on merge.

I'll also change the brackets to angle ones `< >` for consistency with other URLs referenced in SAMtools man pages.

Issue #1393, opened 2021-03-13
segfault merge operation


kemin711:
#### Are you using the latest version of [samtools](https://github.com/samtools/samtools/releases/latest) and [HTSlib](https://github.com/samtools/htslib/releases/latest)? If not, please specify.
(run `samtools --version`)
1.10

 #### Please describe your environment.
- OS Linux Centos and Ubuntu tested both OS
- machine architecture x86_64
- compiler gcc 9.2

#### Please specify the steps taken to generate the issue, the command you are running and the relevant output.
samtools merge out.bam in1.bam in2.bam
Segmentation fault

This happens regardless of other options -O BAM --thread 6

If using bamtools merge no problem.  If I convert the two input bam files into sam file, then samtools merge has no problems.
There must be something strange happened before this step.
I used bwamem2 mem did the alignment. That could be the reason. I have used samtools merge for a long time. only recently I see this. I will keep updating this bug. May compile the latest version to see, the bug is still there.

kemin711:
Recompile the latest version (1.11), still got segmentation fault.

kemin711:
Program received signal SIGSEGV, Segmentation fault.
0x0000555555573272 in finish_rg_pg ()
(gdb) bt
#0  0x0000555555573272 in finish_rg_pg ()
#1  0x0000555555576202 in trans_tbl_init ()
#2  0x000055555557bfdb in bam_merge_core2 ()
#3  0x000055555557dd9d in bam_merge ()
#4  0x00007ffff763b0b3 in __libc_start_main (main=0x55555555beb0 <main>, argc=7, argv=0x7fffffffdcd8, init=<optimized out>, 
    fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdcc8) at ../csu/libc-start.c:308
#5  0x000055555555c6ce in _start ()

kemin711:
(gdb) r merge -O BAM H009994_merged.bam H009994_pair.bam H009994_single.bam
Program received signal SIGSEGV, Segmentation fault.
0x0000555555573ce3 in kl_shift_hdrln (kl=0x0, d=0x7fffffffd7c0) at bam_sort.c:98
98	KLIST_INIT(hdrln, char*, hdrln_free_char)
(gdb) bt
#0  0x0000555555573ce3 in kl_shift_hdrln (kl=0x0, d=0x7fffffffd7c0) at bam_sort.c:98
[...]

PR #1394, opened 2021-03-14
Add missing samtools merge return code check

jmarshall:
This prevents a segfault in cases like #1393, though I don't know how the headers in that issue's input files have managed to get `NULL` from `trans_rg_pg()`.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -71,6 +71,9 @@ Release a.b
  * The documentation for `samtools depth -s` has been improved.
    Thanks to `@wulj2`. (#1355)
 
+ * Fixed a `samtools merge` segmentation fault when it failed to merge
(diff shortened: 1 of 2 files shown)


daviesrob:
Thanks.  I've taken the liberty of adding a NEWS item.

jmarshall:
This prevents the observed segfault, but there still appear to be failure routes through `trans_rg_pg()` that don't produce an error message. So I don't think this has fixed the underlying problem.

daviesrob:
Ok, I will remove the "fixed ..." bit.

jmarshall:
Cheers. Probably ought to credit Kemin Zhou for discovering the problem too.

daviesrob:
Adjusted.  Apologies for the force push.

jmarshall:
No apology needed :smile: :+1:

Issue #1396, opened 2021-03-15
samtools ampliconstats - support for multi reference sequence genomes


ces:
samtools ampliconstats doesn't support multi reference sequence/contig genomes (known issue) but it would be really nice if could or if the lack of support was clearly documented and handled cleanly by the code (e.g. didn't seg fault).

[sanger only example data at /lustre/scratch117/sciops/team117/npg/ces/data/ampliconstats_multiseq/ ]

jkbonfield:
What does it mean to have multiple references for amplicon sequencing?  Our bed format doesn't have a contig name, so what do we want to do?

Do we need a new format with a ref+pos instead of just pos?  This isn't something I can change as it's not my format.

If not, do we want to have one bed file per reference and run the command multiple times specifying the appropriate bed and the chromosome we want to process in the BAM/CRAM?

jkbonfield:
My initial thought here was that the ampliconstats output would add an additional column per reference, eg instead of:

```
# AMPLICON	NUMBER	LEFT	RIGHT
AMPLICON	1	11-35,21-40	301-330
AMPLICON	2	261-275	441-460
AMPLICON	3	401-420	851-880,856-882
..
FREADS	file1	6	0	4
FREADS	file2	6	0	4
```

I'd have:

```
# AMPLICON	REF	NUMBER	LEFT	RIGHT
AMPLICON	r1	1	31-54	386-410
AMPLICON	r1	2	321-342	705-726
AMPLICON	r2	1	11-35,21-40	301-330
AMPLICON	r2	2	261-275	441-460
AMPLICON	r2	3	401-420	851-880,856-882
...
FREADS	file1	r1	5	1
FREADS	file1	r2	6	0	4
FREADS	file2	r1	5	1
FREADS	file2	r2	6	0	4
```

(And so on for all the types of data output.)

It occurs to me that maybe I'd want a single amplicon numbering across all refs.  Eg:

```
# AMPLICON	REF	NUMBER	LEFT	RIGHT
AMPLICON	r1	1	31-54	386-410
AMPLICON	r1	2	321-342	705-726
AMPLICON	r2	3	11-35,21-40	301-330
AMPLICON	r2	4	261-275	441-460
AMPLICON	r2	5	401-420	851-880,856-882
...
FREADS	file1	5	1	6	0	4
FREADS	file2	5	1	6	0	4
```

[...]

PR #1398, opened 2021-03-16
Add a check for multi-ref BED files in ampliconclip & ampliconstats.

jkbonfield:
(See #1396.  **This isn't a fix**, but it detects the problem and responds more gracefully than a crash.)

Long term we wish to make these cope, but for now we take a more
defensive approach by recognising the case where we would give
incorrect results.

This extends bed_pair_list_t with a fixed length buffer, which is a
temporary fudge.  It'll need something like the pooled string
allocator and a ref name pointer within bed_pair_t to properly deal
with multi-ref BEDs.

Ampliconstats now also has code to filter out any reads that are for
another chromosome, so if we have a multi-ref BAM and a single-ref BED
then it'll still work correctly.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -76,6 +76,12 @@ Release a.b
    header `@PG` records. Thanks to John Marshall.  (#1394; reported by
    Kemin Zhou in #1393)
 
+ * Ampliconclip and ampliconstats now guard against the BED file
+   containing more than one reference (chromosome) and fail when
+   found.  Adding proper support for multiple references will appear
+   later.  (#1398)
+
+
 Release 1.11 (22nd September 2020)
 ----------------------------------
 
--- a/bam_ampliconclip.c
+++ b/bam_ampliconclip.c
@@ -1,7 +1,7 @@
 /*  bam_ampliconclip.c -- loads amplicons from a BED file and cuts reads
                           from the 5' end.
 
-    Copyright (C) 2020 Genome Research Ltd.
+    Copyright (C) 2020-2021 Genome Research Ltd.
 
     Authors: Andrew Whitwham <aw7@sanger.ac.uk>
              Rob Davies <rmd+git@sanger.ac.uk>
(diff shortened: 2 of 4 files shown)


bam_ampliconclip.c
```diff
@@ -102,12 +104,23 @@ int load_bed_file_pairs(char *infile, int get_strand, int sort_by_pos,
         if (get_strand) {
             char strand;
 
-            if (sscanf(line.s, "%*s %"SCNd64" %"SCNd64" %*s %*s %c", &left, &right, &strand) != 3) {
+            if (sscanf(line.s, "%255s %"SCNd64" %"SCNd64" %*s %*s %c",
```
whitwham, line 107: So, there are two sscanf statements depending on whether the strand is being read or not.  Should both be changed to handle the ref name.

whitwham: Also, in my copyright police role, the years need changing.

jkbonfield: Probably yes - I didn't notice the second sscanf.  I'll do both.

daviesrob:
As well as the copyright date, could you add something to the NEWS file for this as well, please?

NEWS
```diff
@@ -76,6 +76,10 @@ Release a.b
    header `@PG` records. Thanks to John Marshall.  (#1394; reported by
    Kemin Zhou in #1393)
 
+ * Ampliconclip and ampliconstats now guard against the BED file
```
jmarshall, line 79: I am no ampli-expert, but what do you mean by “reference” here? GRCh37 vs 38? Or reference chromosomes, i.e., more than one different value in the CHROM column?

jkbonfield: Reference as it RNAME.    Terminology is such a mess!  SAM calls them references so that's what I used.  VCF calls them contigs.  I'm assuming from your comment that BED calls them chromosomes? :/

daviesrob:
Hmm, feeding `ampliconclip` a BED file with multiple references triggers a double free - [in load_bed_file_pairs()](https://github.com/samtools/samtools/blob/ec2c3da5/bam_ampliconclip.c#L190) then in [bam_clip()](https://github.com/samtools/samtools/blob/ec2c3da5/bam_ampliconclip.c#L860).  I think this is actually a pre-existing bug in ampliconclip, but now easier to trigger.  If the free in `load_bed_file_pairs()` is needed to avoid a memory leak in `ampliconstats`, setting it to NULL after would be the best fix.

jkbonfield:
Ok I've done that.

What it really needs is a create and destroy function so that the memory is handled by function calls rather than the callers doing it all manually.  However it was done in a bit of a rush at the time of writing, and we're in a rush now too as we don't want to hold up the release.

When we fix it to handle multiple refs we can refactor it more, but I'm hesitant to rush out a bunch of complex changes at the last minute.

daviesrob:
I think it's fine for now, thanks.

NEWS
jmarshall: Messy indeed :smile: Yep, the documents that have historically served as specs say stuff like “`chrom` — The name of the chromosome or scaffold”. (I reckon SAM calls them “reference sequences”, sometimes shortened to “references”.) Thanks for the clarification applied.

PR #1400, opened 2021-03-18
Fix negative depth values in ampliconstats output.

jkbonfield:
These come about due to primer-dimers where the sequence start coordinate matches a known primer, but the sequence doesn't extend beyond the primer itself.  If the primer in question is one of a pair (ie having an ALT primer), then the primer-dimer sequences can finish multiple bases before the inner start coordinate for the amplicon.

This case wasn't handled, leading to a negative increment to the number of amplicon bases and consequential negative CDEPTH.

Also applied a minor rounding error correction to CDEPTH STDDEV where a value of zero is sometimes a tiny negative value, giving rise to -nan when passed into sqrt.

### DIFF
--- a/amplicon_stats.c
+++ b/amplicon_stats.c
@@ -523,20 +523,24 @@ static int accumulate_stats(astats_t *stats,
         stats->nfailprimer++;
 
     if (anum >= 0) {
-        stats->nreads[anum]++;
-        // NB: ref bases rather than read bases
-        stats->nbases[anum] +=
-            MIN(end,amp[anum].min_right+1) - MAX(start,amp[anum].max_left);
+        int64_t c = MIN(end,amp[anum].min_right+1) - MAX(start,amp[anum].max_left);
+        if (c > 0) {
+            stats->nreads[anum]++;
+            // NB: ref bases rather than read bases
+            stats->nbases[anum] += c;
 
-        int64_t i;
-        if (start < 0) start = 0;
-        if (end > args->max_len) end = args->max_len;
+            int64_t i;
+            if (start < 0) start = 0;
+            if (end > args->max_len) end = args->max_len;
 
-        int64_t ostart = MAX(start, amp[anum].min_left-1);
-        int64_t oend = MIN(end, amp[anum].max_right);
-        int64_t offset = amp[anum].min_left-1;
-        for (i = ostart; i < oend; i++)
-            stats->coverage[anum*stats->max_amp_len + i-offset]++;
+            int64_t ostart = MAX(start, amp[anum].min_left-1);
+            int64_t oend = MIN(end, amp[anum].max_right);
+            int64_t offset = amp[anum].min_left-1;
+            for (i = ostart; i < oend; i++)
+                stats->coverage[anum*stats->max_amp_len + i-offset]++;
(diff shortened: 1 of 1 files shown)


amplicon_stats.c
```diff
@@ -523,20 +523,24 @@ static int accumulate_stats(astats_t *stats,
...
-        int64_t ostart = MAX(start, amp[anum].min_left-1);
-        int64_t oend = MIN(end, amp[anum].max_right);
-        int64_t offset = amp[anum].min_left-1;
-        for (i = ostart; i < oend; i++)
-            stats->coverage[anum*stats->max_amp_len + i-offset]++;
+        int64_t c = MIN(end,amp[anum].min_right+1) - MAX(start,amp[anum].max_left);
+        if (c > 0) {
+            stats->nreads[anum]++;
+            // NB: ref bases rather than read bases
+            stats->nbases[anum] += MAX(c, 0);
```
daviesrob, line 530: Isn't this `MAX` redundant due to the condition `if (c > 0)` above?

amplicon_stats.c
```diff
@@ -928,7 +932,7 @@ int dump_stats(char type, char *name, astats_t *stats, astats_args_t *args,
             double n1 = stats->nbases[i] / alen;
             double v = stats->nbases2[i] / (alen*alen) /nfile
                 - (n1/nfile)*(n1/nfile);
-            fprintf(ofp, "\t%.1f", sqrt(v));
+            fprintf(ofp, "\t%.1f", sqrt(v+1e-10));
```
daviesrob, line 935: This didn't work for my test file [MILK-10285F1.210210_A01250_0009_AH32GCDRXY.2t316.cram](ftp://ftp.sra.ebi.ac.uk/vol1/run/ERR531/ERR5313570/MILK-10285F1.210210_A01250_0009_AH32GCDRXY.2t316.cram) as I still got some `nan`s.  Possibly better might be to multiply through by `nfile` and pull the denominator (now all squared) out of the `sqrt()` resulting in:
```
   double alen = amp[i].min_right - amp[i].max_left+1;
   fprintf(ofp, "\t%.1f", sqrt(stats->nbases2[i] * nfile - stats->nbases[i] * stats->nbases[i]) / alen / nfile);
```
However, some of the values in `stats->nbases2` can get a bit big, for example in the test file above:
```
p stats->nbases2[6]
$1 = 226819353033529
```
This needs 48 bits, which doesn't leave that much head-room before it overflows.  It might be a good idea to look into [Welford's algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm) to see if it can calculate the variance without the accumulated numbers getting too big.

jkbonfield: Maybe 1e-10 is too small then, but we could just add a `nfile>1?sqrt(v):0` instead as that's realistically the only time this matters.

I'm not so worried about sizes yet.  For nbases^2 to break 64-bit we'd need 32-bits worth of combined depth across the samples.  I don't think we're going to be more than 1 billion fold coverage yet!  We don't need extra complication here until it really matters, and I don't think it does.  Note we're printing this up with `%.1f` so accuracy isn't vital.

PR #1401, opened 2021-03-23
samtools flags: allow multiple flag arguments to be converted

### LINKED ISSUE #749: samtools flags fails with multiple flag parameters, despite help msg
```

# claims to take multiple things with [,...]
About: Convert between textual and numeric flag representation
Usage: samtools flags INT|STR[,...]

# works
$ samtools flags 60
0x3c    60      UNMAP,MUNMAP,REVERSE,MREVERSE

# help message and no answer
$ samtools flags 60 43

About: Convert between textual and numeric flag representation
Usage: samtools flags INT|STR[,...]

Flags:
        0x1     PAIRED        .. paired-end (or multiple-segment) sequencing technology
        0x2     PROPER_PAIR   .. each segment properly aligned according to the aligner
        0x4     UNMAP         .. segment unmapped
        0x8     MUNMAP        .. next segment in the template unmapped
        0x10    REVERSE       .. SEQ is reverse complemented
        0x20    MREVERSE      .. SEQ of the next segment in the template is reversed
        0x40    READ1         .. the first segment in the template
        0x80    READ2         .. the last segment in the template
        0x100   SECONDARY     .. secondary alignment
        0x200   QCFAIL        .. not passing quality controls
        0x400   DUP           .. PCR or optical duplicate
        0x800   SUPPLEMENTARY .. supplementary alignment

# lets try the comma it suggested, but it ignores the second one
$ samtools flags 60,43
0x3c    60      UNMAP,MUNMAP,REVERSE,MREVERSE
```

jmarshall:
The `samtools flags` command has long annoyed me every time I use it by refusing to convert more than one flag value at once. There's no reason for this, so this PR relaxes the restriction that only one argument be supplied.

Clarify the documentation: each argument is either one INT or one string that is itself a comma-separated list of STRs; note in the man page and usage that multiple arguments are allowed, and reformat the usage to fit in 80 columns.

Fix memory leak: the string returned by `bam_flag2str()` should be `free()`ed by the caller.

Fixes #749 by implementing the desired functionality rather than tweaking the ambiguous documentation. :smile:

### DIFF
--- a/bam_flags.c
+++ b/bam_flags.c
@@ -32,38 +32,54 @@ DEALINGS IN THE SOFTWARE.  */
 #include <unistd.h>
 #include <stdarg.h>
 #include <htslib/sam.h>
+#include "samtools.h"
 
-static void usage(void)
+static void usage(FILE *fp)
 {
-    fprintf(stderr, "\n");
-    fprintf(stderr, "About: Convert between textual and numeric flag representation\n");
-    fprintf(stderr, "Usage: samtools flags INT|STR[,...]\n");
-    fprintf(stderr, "\n");
-    fprintf(stderr, "Flags:\n");
-    fprintf(stderr, "\t0x%x\tPAIRED        .. paired-end (or multiple-segment) sequencing technology\n", BAM_FPAIRED);
-    fprintf(stderr, "\t0x%x\tPROPER_PAIR   .. each segment properly aligned according to the aligner\n", BAM_FPROPER_PAIR);
-    fprintf(stderr, "\t0x%x\tUNMAP         .. segment unmapped\n", BAM_FUNMAP);
-    fprintf(stderr, "\t0x%x\tMUNMAP        .. next segment in the template unmapped\n", BAM_FMUNMAP);
-    fprintf(stderr, "\t0x%x\tREVERSE       .. SEQ is reverse complemented\n", BAM_FREVERSE);
-    fprintf(stderr, "\t0x%x\tMREVERSE      .. SEQ of the next segment in the template is reversed\n", BAM_FMREVERSE);
-    fprintf(stderr, "\t0x%x\tREAD1         .. the first segment in the template\n", BAM_FREAD1);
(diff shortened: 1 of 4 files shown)


bam_flags.c
```diff
@@ -32,38 +32,54 @@ DEALINGS IN THE SOFTWARE.  */
...
-    else
+    if ( argc < 2 ) { usage(stdout); return 0; }
+
+    int i;
+    for (i = 1; i < argc; i++)
     {
-        int mask = bam_str2flag(argv[1]);
-        if ( mask<0 ) { fprintf(stderr,"Error: Could not parse \"%s\"\n", argv[1]); usage(); return 1; }
-        printf("0x%x\t%d\t%s\n", mask, mask, bam_flag2str(mask));
+        int mask = bam_str2flag(argv[i]);
```
valeriuo, line 78: A check of the `mask` value against the maximum flag numerical combination (4095) would be welcome.

valeriuo (commented):
This change is also NEWS worthy.

bam_flags.c
jmarshall: IMHO what to do for unassigned bits should be left to `bam_str2flag()` and `bam_flag2str()`.

jmarshall:
I'm happy to write a NEWS entry if that's how the _NEWS_ file is being handled at this early stage of this development cycle.

jkbonfield:
> I'm happy to write a NEWS entry if that's how the _NEWS_ file is being handled at this early stage of this development cycle.

I don't mind either way, but I think it's something we should probably discuss as a team.  (Unfortunately our weekly meeting was taken up with other topics.)

Doing it as we go makes the release easier, however doing it as we go in PR is a guaranteed merge failure as most PRs end up modifying the same lines in NEWS.

IIRC we used to have [NEWS: text] style annotations in the commit messages so we could git log and parse them out to produce news.   It feels like a more manageable strategy that doesn't make needless work for ourselves.

valeriuo:
Merged as d515e1b0359fa0a6f45894c9dc30fbf64eabddc4

PR #1402, opened 2021-03-26
Typo patrol

jmarshall:
I noticed the “_in which ase_” one just now, and here is the result of running a spelling checker over the man pages. A pity this didn't all happen a couple of weeks ago… :smile:

Also modernise the `samtools view` examples: Demonstrate using `-o` rather than shell redirection, which is more error-prone particularly in conjunction with `nohup`.

In particular, purge the very legacy `samtools view -bS in.sam > out.bam` usage.


daviesrob:
All looks good, thanks.

jmarshall:
Thanks for merging.

I just noticed there are very similar examples at the bottom of _samtools.1_, but at least they don't use `-S`…

PR #1404, opened 2021-03-30
add support for addreplacerg when RG ID exist in origin bam

ycsos:
When someone create a Bam with wrong RG line or forget to add some tag, but want to keep RG ID.
addreplacerg can only replace a different RG line, the commit allow replace the same RG line.


bam_addrprg.c
```diff
@@ -360,9 +360,8 @@ static bool init(const parsed_opts_t* opts, state_t** state_out) {
         // Check does not already exist
         kstring_t hdr_line = { 0, 0, NULL };
         if (sam_hdr_find_line_id(retval->output_header, "RG", "ID", opts->rg_id, &hdr_line) == 0) {
-            fprintf(stderr, "[init] ID of new RG line specified conflicts with that of an existing header RG line. Overwrite not yet implemented.\n");
-            free(hdr_line.s);
-            return false;
+            sam_hdr_remove_line_id(retval->output_header, "RG", "ID", opts->rg_id);
+            fprintf(stderr, "[init] ID of new RG line specified found, we will replace the old ID line with new one.\n");
```
valeriuo, line 364: Since it is no longer considered an erroneous path, printing to `stderr` should be removed. Maybe add an error message in case `sam_hdr_remove_line_id` fails.

valeriuo (commented):
Elegant solution. Could do with a test case.

bam_addrprg.c
```diff
@@ -360,9 +360,8 @@ static bool init(const parsed_opts_t* opts, state_t** state_out) {
         // Check does not already exist
         kstring_t hdr_line = { 0, 0, NULL };
         if (sam_hdr_find_line_id(retval->output_header, "RG", "ID", opts->rg_id, &hdr_line) == 0) {
-            fprintf(stderr, "[init] ID of new RG line specified conflicts with that of an existing header RG line. Overwrite not yet implemented.\n");
-            free(hdr_line.s);
-            return false;
+            sam_hdr_remove_line_id(retval->output_header, "RG", "ID", opts->rg_id);
```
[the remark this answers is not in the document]
ycsos: I will add a check for sam_hdr_remove_line_id, and when return -1 print error and free hdr_line.s

bam_addrprg.c
ycsos: I totally agree with that, but I want to print some information to remind user you replace the old RG line with the new one, how should I do to achieve the goal without print to stderr

jkbonfield:
Thanks both.  Looks good and tests well.  I approve of using `ks_free` too :-)

PR #1410, opened 2021-04-01
Read Multiple Refs/Chroms from a BED file

whitwham:
Change samtools ampliconclip so that it will read and use multiple references (or chroms in BED terms).

The other program to use this BED reading code, samtools ampliconstats has been left for later.  The only changes made are to accommodate new structure names and contents.


bam_ampliconclip.c
```diff
@@ -64,15 +64,170 @@ typedef struct {
...
+                memcpy(list->ref, ref, 256); // FIXME keep internal ref?
+                list->longest = 0;
+                list->size = 256;
+                list->length = 0;
+
+                if ((list->bp = malloc(list->size * sizeof(bed_entry_t))) == NULL) {
+                    fprintf(stderr, "[ampliconclip] error: unable to allocate memory for bed data.\n");
+                    ret = 1;
+                    goto error;
+                }
```
jkbonfield, line 142: For brevity, this could be simplified by setting size 0 and bp to NULL and letting the realloc below do the initial allocation so we merge the error handling and calls.  Ie here:

```
                list->longest = 0;
                list->size = 0;
                list->length = 0;
                list->bp = NULL
```

and below in the realloc stage, something similar to (instead of `list->size *= 2`)

```
    list->size += list->size/2 + 256;
```

bam_ampliconclip.c
```diff
@@ -64,15 +64,170 @@ typedef struct {
...
+        if (bed_itr == kh_end(bed_lists)) { // new ref entry
+            char *ref_name = strdup(ref); // need a copy for the hash key
+
+            if (!ref_name) {
+                fprintf(stderr, "[amplicon] error: unable to allocate memory for ref name.\n");
+                ret = 1;
+                goto error;
+            }
+
+            bed_itr = kh_put(bed_list_hash, bed_lists, ref_name, &hret); // FIXME ref needs to be a strdup somewhere
```
jkbonfield, line 127: I think this FIXME has already been fixed

bam_ampliconclip.c
```diff
@@ -64,15 +64,170 @@ typedef struct {
...
+
+            if (hret >= 0) {
+                list = &kh_val(bed_lists, bed_itr);
+
+                // initialise the new hash entry
+                memcpy(list->ref, ref, 256); // FIXME keep internal ref?
```
jkbonfield, line 133: This FIXME comment seems like it was written before the strdup code was added.

Now it can presumably just be a pointer to the same strdup memory used as the hash key.  Or alternatively, if we have a literal char[256] buffer, then we could use that as the hash key and avoid strduping it.  Either/or but not both are needed.

bam_ampliconclip.c
```diff
@@ -64,15 +64,170 @@ typedef struct {
...
+
+            if (!ref_name) {
+                fprintf(stderr, "[amplicon] error: unable to allocate memory for ref name.\n");
+                ret = 1;
+                goto error;
+            }
+
+            bed_itr = kh_put(bed_list_hash, bed_lists, ref_name, &hret); // FIXME ref needs to be a strdup somewhere
+
+            if (hret >= 0) {
```
jkbonfield, line 129: We have `hret >= 0` in and if-else clause, but kh_put is tri-state (well quad-state but 2 are the same for our purposes); <0 = fail, 0 = already-present, >0 = new/deleted.

We probably want the `<0` case covered too, although admittedly we often fail elsewhere to consider hash table growth failures.

bam_ampliconclip.c
```diff
@@ -111,21 +266,21 @@ int load_bed_file_pairs(char *infile, int get_strand, int sort_by_pos,
                 ret = 1;
                 goto error;
             }
-            if (*pairs->ref) {
-                if (strncmp(ref, pairs->ref, 256)) {
+            if (*entries->ref) {
+                if (strncmp(ref, entries->ref, 256)) {
                     fprintf(stderr, "[ampliconclip] error: "
```
jkbonfield, line 271: I'm unsure why this error message and check are still here.  It's meant to handle this case now afterall.

Is it a rebasing issue?  I think I added that check relatively recently, perhaps before you forked off your branch.

bam_ampliconclip.c
```diff
@@ -140,44 +295,44 @@ int load_bed_file_pairs(char *infile, int get_strand, int sort_by_pos,
...
```
jkbonfield, line 300: As above

jkbonfield (commented):
First round of comments.  I still need to produce some test data and play with it on the command line.

At some point we'll need to revist the bed reader to make it more generic (eg it shouldn't choke on bed files with more columns), but for our purpose it's a restricted tight definition so that's somethjing we can deal with later.

bam_ampliconclip.c
whitwham: I can do that.  It really only removes one if statement but it does reduce the number of moving parts.

bam_ampliconclip.c
whitwham: It has.

bam_ampliconclip.c
whitwham: We have talked about this in other places, but for the record I wasn't sure if keeping the reference name inside the structure was useful or not.

bam_ampliconclip.c
whitwham: The `<0` is handled in the `else` part of the `if`.

The `==0` is handled elsewhere (for memory allocation reasons) so `hret >= 0` should really be only `hret > 0`.

bam_ampliconclip.c
whitwham: For the record (see above) this is the old code needed to make ampliconstats build.  Once ampliconstats is converted we can remove this entire section of code.

bam_ampliconclip.c
whitwham: As above.

bam_ampliconclip.c
jkbonfield: Thanks.  I don't mind hugely either way, but in general I prefer simplicity if possible.  Less to go wrong and easier to understand (mostly).

bam_ampliconclip.c
jkbonfield: I do still use `sites->ref` in ampliconstats, but that's historic and it could trivially be replaced by `kh_key(bed_hash, k)` which I think should always be the same thing.  I'll make the change so it's one less thing to worry about.

bam_ampliconclip.c
jkbonfield: My mistake.  On re-reading it I see I managed to muddle my indentation levels and mistook the `// existing ref` else clause as associated with that if.  I see it already has the "ref hashing failure" error handled.  Ignore my comment.

PR #1414, opened 2021-04-12
Add citation section and link to relevant paper

valeriuo:


daviesrob:
Should it be "cite this paper" rather than "cite this repository"?

PR #1415, opened 2021-04-13
[coverage] Add option for setting the maximum coverage depth.

### LINKED ISSUE #1395: samtools coverage unbounded memory usage (lacks -d/--depth parameter)
Version: 1.11-3-g7028dd4 (using htslib 1.10.2-145-g8016127)

Running `samtools coverage` has unbounded memory usage and will crash if there exists a region of very high coverage.

Potential solutions are:
- Added a `-d` parameter similar to `samtools depth`
- Redesign algorithm to only load 1 read at a time.

The former is simpler, the later is much more memory efficient and can correctly report depth for high-coverage samples. The only output that requires look-ahead is the coverage field and that only requires 1 bit per position with the max length of the positional look-ahead bounded by (reference) length of the longest read alignment.

The following test case causes `samtools coverage` to use of over 10GB memory:
```
#!/bin/bash
two_to_the_n=25
compressable_data=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
echo "@SQ	SN:$compressable_data	LN:1000" > test_depth_header.sam
echo "$compressable_data	0	$compressable_data	1	60	200M	*	0	0	$compressable_data	$compressable_data" > test_depth_1.sam
for i in $(seq 2 $two_to_the_n) ; do
	last_file=test_depth_$(($i-1)).sam
	cat $last_file $last_file > test_depth_${i}.sam
done
cat test_depth_header.sam test_depth_${two_to_the_n}.sam | samtools -@ $(nproc) view -b - > test_depth_${two_to_the_n}.bam && rm test_depth_*.sam
samtools coverage test_depth_${two_to_the_n}.bam
```

valeriuo:
`samtools coverage` now accepts option `-d` (long form `--depth`) for setting the maximum depth at which alignments are considered when computing the coverage figures. Also, the default value of the maximum depth has been reset from `INT_MAX` to 1000000.

Fixes #1395

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,9 @@
 Release a.b
 -----------
 
+ * samtools coverage now allows setting the maximum depth, using the -d/--depth
+   option. Also, the default maximum depth has been set to 1000000.
+
  * samtools flags now accepts any number of command line arguments,
    allowing multiple SAM flag combinations to be converted at once. (PR #1401)
 
--- a/coverage.c
+++ b/coverage.c
@@ -24,7 +24,7 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.  */
 
 /* This program calculates coverage from multiple BAMs
- * simutaneously, to achieve random access and to use the BED interface.
+ * simultaneously, to achieve random access and to use the BED interface.
  * To compile this program separately, you may:
  *
  *   gcc -g -O2 -Wall -o bamcov -D_MAIN_BAMCOV coverage.c -lhts -lz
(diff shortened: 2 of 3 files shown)


daviesrob:
Merging as in interim solution to the coverage memory consumption problem.  If the methods used for depth in #1428 can be extended to coverage we may be able to remove this later.

PR #1422, opened 2021-04-26
Stop samtools cat from outputting multiple CRAM EOF markers.

jkbonfield:
Samtools cat naively copies all incoming containers, and then writes an empty EOF container when closing the file.  This leads to multiple internal EOF markers as they're implemented in CRAM as an empty container.

Oddly there was code explicitly dealing with this, which seems redundant.  Perhaps it served a purpose in early versions.  Regardless internal EOF markers is permitted and are ignored, so this change is just for tidyness sake and to improve robustness incase bad implementations stop on the first EOF.

### DIFF
--- a/bam_cat.c
+++ b/bam_cat.c
@@ -271,6 +271,11 @@ int cram_cat(int nfn, char * const *fn, const sam_hdr_t *h, const char* outcram,
         // Copy contains and blocks within them
         while ((c = cram_read_container(in_c))) {
             if (cram_container_is_empty(in_c)) {
+                cram_block *blk;
+                // Container compression header
+                if (!(blk = cram_read_block(in_c)))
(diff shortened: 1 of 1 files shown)


valeriuo:
I am investigating some large memory allocations ~1 GB. It seems that a `cram_read_container(in_c)` call after a true `cram_container_is_empty(in_c)` might lead to some uninitialized container fields, especially `num_landmarks`, which results in a large [allocation](https://github.com/samtools/htslib/blob/c74a63ef8c1519b2592b4b718398917c1d4139d3/cram/cram_io.c#L3783).

valeriuo:
Merged into develop as 1ea60adbf492d0596a8fb01fd44bafe8fcee5fc0

Issue #1419, opened 2021-04-22
Samtools coverage histogram not viewed properly


deminatanja:
Hi,

I am using samtools 1.10, Linux 3.10.0-1062.45.1.el7.x86_64, gcc (GCC) 9.1.0.

Samtools coverage command used as below gives a histogram file that can't be viewed properly.

samtools coverage -m -A -w 32 /path/x.sorted.bam -o /path/outputfile

Viewed with head command:
NODE_1_length_307510_cov_3.543238 (307.5Kbp)
>  90.00% │████████████████████████████████│ Number of reads: 50206
>  80.00% │████████████████████████████████│     (1068 filtered)
>  70.00% │████████████████████████████████│ Covered bases:   307.5Kbp
>  60.00% │████████████████████████████████│ Percent covered: 100%
>  50.00% │████████████████████████████████│ Mean coverage:   24.5x
>  40.00% │████████████████████████████████│ Mean baseQ:      35.2
>  30.00% │████████████████████████████████│ Mean mapQ:       40.1
>  20.00% │████████████████████████████████│ 
>  10.00% │████████████████████████████████│ Histo bin width: 9.6Kbp

Viewed in nano:
NODE_1_length_307510_cov_3.543238 (307.5Kbp)
>  90.00% ?^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^$
>  80.00% ?^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^$
>  70.00% ?^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^$
>  60.00% ?^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^$
>  50.00% ?^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^$
>  40.00% ?^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^$
>  30.00% ?^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^??^?^$
[...]

PR #1423, opened 2021-04-26
Make coverage ASCII mode true ASCII.

jkbonfield:
The example in the man page used "#" characters, but in source code it
was using a more limited UTF8 encoding of only full and half-height
blocks instead of 8 height levels.

It's a misnomer to claim this is ASCII.  We could still have a
half-height UTF8 mode, but I'm not sure I see the point in catering to
the middle case much.  ASCII however should be purely printable
without having to worry about text editors or pagers escaping
non-printable characters.  This switches to : and . chars.

Fixes https://github.com/samtools/samtools/issues/1419

### DIFF
--- a/coverage.c
+++ b/coverage.c
@@ -1,7 +1,7 @@
 /* coverage.c -- samtools coverage subcommand
 
     Copyright (C) 2018,2019 Florian Breitwieser
-    Portions copyright (C) 2019-2020 Genome Research Ltd.
+    Portions copyright (C) 2019-2021 Genome Research Ltd.
 
     Author: Florian P Breitwieser <florian.bw@gmail.com>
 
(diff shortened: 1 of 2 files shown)


jmarshall:
> We could still have a half-height UTF8 mode, but I'm not sure I see the point in catering to the middle case much.

Presumably because you're not a Windows user connecting to your remote samtools machine using putty. Perhaps @fbreitwieser can comment on whether this is still a problem.

jkbonfield:
Indeed I'm not.  I much prefer a proper tabbed terminal (either [Microsoft's own](https://www.microsoft.com/en-gb/p/windows-terminal/9n0dx20hk701) which is pretty feature rich, or what I normally use is xfce4-terminal running under WSL as I want far more than just putty to a remote server).

There's really no difference in functionality with this PR, just style.  You still get two levels of depth (full and half height), only it's now ascii-fied instead of filled in UTF8 blobs.   If that's really a cause for concern I can fix it obviously, but erring on side of simplicity for now.

daviesrob:
Rebased so tests should work...

PR #1427, opened 2021-04-29
Support an HTSlib source tree that has a separate build directory

jmarshall:
Companion PR to samtools/htslib#1277 which enables samtools's _configure_ to recognise an HTSlib separate build directory, which is treated the same as a source directory — as for external projects' purposes this is identical apart from `-I` needing to point to the associated source directory (only).

Pick `HTSSRCDIR` out of the HTSlib build directory's _htslib_vars.mk_ (generated from _builddirs_vars.mk.in_). `@HTSLIB_CPPFLAGS@` and `@HTSLIB_LDFLAGS@` need to be fully expanded: `$(HTSDIR)` would not be available for external projects that don't use `@HTSDIR@` or _htslib.mk_.

In due course _m4/ax_with_htslib.m4_ should be copied to bcftools too.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -123,6 +123,9 @@ specify further optional external requirements:
     installation directory (i.e., it has 'include' and 'lib' subdirectories
     containing HTSlib headers and libraries).
 
+    (If the desired HTSlib source tree has been configured to build in a
+    separate build directory, DIR should refer to the build directory.)
+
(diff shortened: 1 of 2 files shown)


daviesrob:
Looks good, thanks.  I've tested against both in- and out-of-tree htslibs, and got successful builds.

Issue #1411, opened 2021-04-03
How to speed up samtools depth


ShenTTT:
It seems that multithreading is not supported for samtools depth, and it runs super slow...Is there a way I can speed it up?

valeriuo:
You can run multiple `samtools depth` processes in parallel, one mapped to each region. Take a look at [this script](https://github.com/samtools/samtools/issues/1242#issuecomment-627386086).

Issue #889, opened 2018-07-10
samtool depth reports unexpected values


bernt-matthias:
#### Are you using the latest version of [samtools](https://github.com/samtools/samtools/releases/latest) and [HTSlib](https://github.com/samtools/htslib/releases/latest)? If not, please specify.
(run `samtools --version`)

1.8

 #### Please describe your environment.
- OS (run `uname -sr` on Linux/Mac OS or ` wmic os get Caption, Version` on Windows)
- machine architecture (run `uname -m` on Linux/Mac OS or ` wmic os get OSArchitecture` on Windows)
- compiler (run `gcc --version` or `clang --version`)

Linux 4.4.0-130-generic
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609

#### Please specify the steps taken to generate the issue, the command you are running and the relevant output.

`samtools depth -d 3 eboVir3.bam`

the file can be downloaded here: https://github.com/bernt-matthias/tools-iuc/blob/f686ef690b2e932e40acb3fd38bb9a00dd0ec911/tool_collections/samtools/samtools_depth/test-data/eboVir3.bam?raw=true

According to the documentation I do not expect depth values > 3, but there are. Could you please clarify this?

peterjc:
I noticed a certain fuzziness while writing this tool which was tested on ultra-high coverage viral data:

https://github.com/peterjc/pico_galaxy/blob/master/tools/coverage_stats/coverage_stats.py

I concluded the limit was applied at read level, not mapped base level, but rather than solving exactly what samtools was doing and trying to fix it, I took a pragmatic way out - I told samtools to use the desired max-depth plus a margin, and then applied the desired max-depth in post-processing.

jkbonfield:
The depth limit is there to prevent excessive memory usage in extremely deep locations.  It is not a cast iron guarantee of precise maximum depth.  I haven't checked the code in detail, but my expectation is this is likely caused by a naive depth filtering that doesn't take into account the full nuances of how CIGAR strings can change read sizes slightly.

bernt-matthias:
Then I guess the following questions should to be answered in the documentation: 

- What is actually counted (ie in which cases is read counted for a base)
- Are the results of the program exact with `-d 0`? 
- How does -d !=0 affect the counts?

peterjc:
[...]

Issue #1036, opened 2019-04-15
mpileup read-pair overlap detection not detecting all overlaps?


jeanlain:
Checked on both samtools 1.9 on macOS and samtools 1.5 on Redhat Linux

Steps:
Decompress and index the attached bam file with `samtools index AV-WXa1591_m_scaffold1.bam`
then 
`samtools mpileup -A -Q1 -x -r scaffold_1:5064-5064 WXa1591_m_scaffold1.bam`

The program detects 40G and 13T at that position, which appears correct.

Now run

`samtools mpileup -A -Q1 -r scaffold_1:5064-5064 WXa1591_m_scaffold1.bam`

The program returns 38G and 12T.

Expected results : **36**G and 12T.

My analysis of this file indicates that both mates in **5** read pairs overlap on position 5064, hence ignoring overlaps should result in coverage dropping from 53 to 48, not 53 to 50.
Position 5064 is just a random one. Other positions show this problem. 
Note: this apparent issue may in fact be my misunderstanding of what the "read-pair overlap detection" is supposed to do.

[AV-WXa1591_m_scaffold1.bam.zip](https://github.com/samtools/samtools/files/3080155/AV-WXa1591_m_scaffold1.bam.zip)

tseemann:
Some relevant discussion here?
https://bioinformatics.stackexchange.com/questions/5427/double-counting-coverage-of-overlapped-read-pairs

jeanlain:
No this doesn't seem relevant. It does not discuss issues with overlap detection in mpileup.

valeriuo:
I haven't looked at your data yet, but it looks similar to #986.

valeriuo:
After looking at the data, it appears not be related to #986. In fact, it works as expected.
Two of the 5 reads pairs overlapping position 5064 are:
```
HWI-ST218:508:C77YEANXX:4:1213:7943:80820       161     scaffold_1      5013    60      100M2I23M       =       5022
HWI-ST218:508:C77YEANXX:4:1213:7943:80820       81      scaffold_1      5022    60      91M2I32M        =       5013

HWI-ST218:508:C77YEANXX:6:1209:11011:36200      145     scaffold_1      4969    60      3S122M  =       4970
HWI-ST218:508:C77YEANXX:6:1209:11011:36200      97      scaffold_1      4970    60      121M4S  =       4969
```
[...]

PR #1428, opened 2021-05-05
Complete rewrite of samtools depth

jkbonfield:
This uses a circular buffer to accumulate and report depth, instead of
relying on mpileup.  This means it's considerably faster (particularly
on deep data and on CRAM files) and doesn't have the need for a depth
limit to avoid high memory usage.  It still silently accepts the -d/-m
(max depth) parameter, but ignores it.

All output should be identical to the old command except potentially
for the overlap removal (-s) option.  The old code here is complex,
requiring all in-flight records to be held in memory by design.
This is explicitly what I wanted to remove as it is remembering these
records-in-flight that gives the high memory usage and the requirement
of a maximum depth.

Therefore the new depth algorithm may give different results:

- It doesn't force a -q1 or higher.  This is needed with mpileup as
  mpileup doesn't actually removal overlaps, it merely marks them.
  (Try old "depth -q0 -s" and "depth -s -q0" to see this.)

- Even if the user manually adds the -q1 option, the overlap removal
  is purely coordinate based.  The mpileup "removal" is coordinate,
  sequence and quality based.  If overlapping bases agree, the quality
  used is the summation of both qualities, else it's 80% of the
  highest quality.

  This gives differences where read 1 has zero qual and read 2
  overlapping has non-zero qual.  With -q1 we'd ignore that first base
  while mpileup would still count that depth.

- Overlap removal applies to all paired reads, not only those with the
  proper-pair flag set.  This is related to #1036.  Strongly
  overlapping reads are sometimes not labelled as a proper pair as
  they fall outside the expected insert size distribution, but those
  are the reads we really ought to be deduplicating.  (Note this is
  still an issue for mpileup though so it doesn't fix that issue.)

In the process of assessing this, two bugs were discovered in the old
depth code.

- Specifying multiple BAM files where the first file consists solely
  of unmapped reads produces no output.  It should report the data for
  the 2nd file onwards still.

- Due to a mistake in the CRAM_OPT_REQUIRED_FIELDS initialisation, the
  overlap removal (-s) option didn't work correctly on CRAMs, wrongly
  removing every other read regardless of read name.


jkbonfield:
NB:  I also took the liberty of prepending this with a trivial gcc 11.1 commit.  It's not sufficient to get it fully working, but will be once we fix kseq.h in htslib.

jkbonfield:
An example comparing against 1.11 (I didn't have a convenient pre-made 1.12 binary kicking around, but we haven't changed it) on multiple **deep** data (5-10k):

```
@ deskpro[samtools.../samtools]; time samtools depth -d0 /var/tmp/34002/34
002_1#{1..9}.cram | md5sum 
506530a840e940db54b093d22ad87d27  -

real	0m1.070s
user	0m0.913s
sys	0m0.095s

@ deskpro[samtools.../samtools]; time samtools111 depth -d0 /var/tmp/34002/34002_1#{1..9}.cram | md5sum
506530a840e940db54b093d22ad87d27  -

real	0m35.448s
user	0m35.351s
sys	0m0.105s
```

On shallower data (30-50x), in BAM, with more complex options enabled:

```
@ deskpro[samtools.../samtools]; time samtools depth -q20 -Q25 -l100 -J -d0 /var/tmp/novaseq-10m.bam|md5sum
bc38ced84324558965a47adf9acd892b  -

real	0m6.049s
user	0m6.479s
sys	0m0.313s

@ deskpro[samtools.../samtools]; time samtools111 depth -q20 -Q25 -l100 -J -d0 /var/tmp/novaseq-10m.bam|md5sum
bc38ced84324558965a47adf9acd892b  -

real	0m19.755s
user	0m20.194s
sys	0m0.387s
```

So between 3 and 30x speed up in those examples.

jkbonfield:
Mostly fixes #1411 as it's both considerably faster, and has some benefit from threads (maybe 2-3x speed up before the main thread becomes the bottleneck).

Fixes #889

bam2depth.c
```diff
@@ -41,377 +43,883 @@ DEALINGS IN THE SOFTWARE.  */
...
-            return 1;
-        }
-        n = (argc - optind) / 2;
+
+    // Accumulate depth, based on CIGAR
+    uint32_t *cig = bam_get_cigar(b);
+    int ncig = b->core.n_cigar, j, k, spos = 0;
+
+    // Zero new (previously unseen) coordinates so increment works later.
+    // FIXME: do this for all files so output is easy,
```
whitwham, line 287: Since you have decided on which case to use, do you want to remove the FIXME?

bam2depth.c
```diff
@@ -41,377 +43,883 @@ DEALINGS IN THE SOFTWARE.  */
...
+            }
+
+            if (b[i]->core.tid < 0)
+                continue;
+            if (b[i]->core.flag & opt->flag)
+                continue;
+            if (b[i]->core.qual < opt->min_mqual)
+                continue;
+
+            // FIXME?  Should this be sequence length minus soft-clips?
```
whitwham, line 532: Since the read start is from the soft clipped position there is a good case to be made for using the clipped length.  I guess the original depth did not do this though.

whitwham (commented):
Looks good.  There are a few FIXMEs in there.  I don't know whether you want to address them, remove them or leave them till later.  I'm happy to check it in if you are.

bam2depth.c
jkbonfield: Good idea.

bam2depth.c
jkbonfield: I'd forgotten about that.  Yes it probably needs improving with an aligned-length function.  I did later realise why it calls the function to compute the sequence length from CIGAR instead of just using `b->core.l_qseq`, which totally baffled me originally.  It's because sequence `*` has length 0, but the min length parameter still wishes to accept it if CIGAR indicates it spans a large region.

Yes my intention was to have this in two stages.  The first is to get as bug-compatible as possible with the former samtools depth so it's easy to validate the impact of the rewrite, followed up with fixing what I view as bugs in the original implementation.

I did originally have it split up, but I think I got overly tidy and squashed it together, losing this separation.  Hence the bug fix for PROPER_PAIR vs PAIRED flag is now in there already IIRC.  Or maybe I figured it's only for overlap removal which is already different so not possible to validate anyway.  Not sure now.

PR #1434, opened 2021-05-18
Add `samtools merge -o FILE` option as an optional alternative to the current oddball merge syntax

jmarshall:
The transition of `samtools sort [-o] in.bam out.prefix` to the typical Unix-style
```
samtools sort [-o out.bam] in.bam
```
is long since complete. As I recall, it was intended to transition `samtools merge out.bam in1.bam in2.bam…` similarly, but this was comparatively easy (no conflicting meaning for `-o` to deal with!) hence it never quite actually happened.

This PR adds support for
```
samtools merge [options…] -o out.bam [options…] in1.bam in2.bam…
```
which is easy to support unambiguously alongside the existing syntax. This means that `merge` can be used with the same familiar Unix-style `-o` output file control as `view`, `sort`, etc, avoiding [reasonable complaints like this one](https://twitter.com/David_McGaughey/status/1394331493820817408).

Add an `fnout` variable and increment `optind` past it for old-style command lines. As `optind` has been incremented, replace `optind+1` with `optind` in all the filename-list calculations.

The difficult additional transition here would be to default to writing to standard output, and this PR does not attempt to do that. As the existing syntax is quite entrenched, it would not be feasible to move to that more typical default without a long transition period, so for now (and for the foreseeable) `-o -` remains required in pipelines.

### DIFF
--- a/doc/samtools-merge.1
+++ b/doc/samtools-merge.1
@@ -1,7 +1,7 @@
 '\" t
 .TH samtools-merge 1 "17 March 2021" "samtools-1.12" "Bioinformatics tools"
 .SH NAME
-samtools merge \- merges multiple sorted input files into a single output.
+samtools merge \- merges multiple sorted files into a single file
 .\"
 .\" Copyright (C) 2008-2011, 2013-2019 Genome Research Ltd.
 .\" Portions copyright (C) 2010, 2011 Broad Institute.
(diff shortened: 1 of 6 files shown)


doc/samtools-merge.1
```diff
@@ -106,6 +106,19 @@ are ignored.)
 The input alignments are sorted by read names rather than by chromosomal
 coordinates
 .TP
+.BI -o \ FILE
```
valeriuo, line 109: Perhaps the synopsis should include the new option, as well.

valeriuo (approved):

doc/samtools-merge.1
jmarshall: I thought about it, both here and in the Usage display. It needs to be in both if this is to become the new “standard”, but conversely dual synopses kind of increases the confusion… Happy to add it if consensus develops…

It probably will need a NEWS entry too, so people can look back and see when this syntax became available.

daviesrob: Maybe:
```
Usage: samtools merge -o <out.bam> [-nurlf] [-h inh.sam] [-b <bamlist.fofn>] <in1.bam> [<in2.bam> ... <inN.bam>]
or:    samtools merge [-nurlf] [-h inh.sam] [-b <bamlist.fofn>] <out.bam> <in1.bam> [<in2.bam> ... <inN.bam>]
```

Deliberately moving the output file to emphasise that it doesn't need to be with the list of inputs.

valeriuo: A NEWS entry is definitely required.

jmarshall:
Synopses updated, basically using Rob's suggestion but just writing `[options]` rather than writing out (some of) the actual options as previously, for brevity and clarity. NEWS entry added.

There was previously an extra word thinkoed into the synopsis (“[**merge** samtools merge […]](https://www.htslib.org/doc/samtools-merge.html)”). The same problem existed in the view and tview man pages, which are fixed in an extra commit.

valeriuo:
Merged as 677491c411c0b73e2aebe2a9f3aac7284ba9672d and 4eae367b24dc2426b79e3acfc7f00cceba5c144f

PR #1436, opened 2021-05-19
Fix rf filter error

### LINKED ISSUE #1435: Filter flag do not work as expect.
For example, if I would like to filter R1 of proper pair reads, while the read is in the some orientation of the reference sequence. I can use flag 99 (1+2+32+64) in `samtools view` to subset such reads.
But in `sample mpileup`, the `--rf` argument can not deal with this situation correctly. Because flag `99` contains the `1` flag, any paired reads will be selected by `mpileup`, ignoring other flags (2, 32, 64).

image

- I think this line should be fixed

https://github.com/samtools/samtools/blob/5fcc7a4fa822bb216030b6d6803e846f99b7ad41/bam_plcmd.c#L332

yech1990:
close #1435

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -329,7 +329,7 @@ static int mplp_func(void *data, bam1_t *b)
             skip = 1;
             continue;
         }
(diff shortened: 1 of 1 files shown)


jkbonfield:
Closing as this changes behaviour.  We will fix documentation instead.

PR #1438, opened 2021-05-27
Replace fopen with stat for file exists check.

### LINKED ISSUE #1437: samtools hangs when using a named pipe
Yes, samtools 1.12

Ubuntu 20.04.2 LTS
Linux 5.8.0-53-generic, x86_64
samtools installed via conda

Using these [example](https://github.com/samtools/samtools/files/6550353/example.zip) files (please unzip first).

I can run:

```
samtools merge intermediary *.sorted.bam
samtools mpileup intermediary | gzip > all.mpileup.gz
```

but when trying to use a [named pipe](https://en.wikipedia.org/wiki/Named_pipe), the second command just hangs and does nothing:

```
mkfifo intermediary
samtools merge intermediary *.sorted.bam &
samtools mpileup intermediary | gzip > all.mpileup.gz
# rm intermediary
```

I have also [reported this to Snakemake](https://github.com/snakemake/snakemake/issues/1018), where this issue first came up when trying to bind different samtool steps without having to write out all intermediate files.

jkbonfield:
Access is already used in sam.c, so hopefully this won't have any impact on compatibility (access is POSIX rather than C).

This fixes an issue where the check for file existence when file is a named pipe causes a hang.  Fixes #1437

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -33,6 +33,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include <stdio.h>
 #include <string.h>
 #include <time.h>
+#include <sys/types.h>
 #include <sys/stat.h>
 #include <unistd.h>
 #include <getopt.h>
@@ -1506,9 +1507,8 @@ int bam_merge(int argc, char *argv[])
 
     hts_srand48(random_seed);
     if (!(flag & MERGE_FORCE) && strcmp(fnout, "-") != 0) {
-        FILE *fp = fopen(fnout, "rb");
-        if (fp != NULL) {
-            fclose(fp);
+        struct stat sbuf;
+        if (stat(fnout, &sbuf) == 0 && S_ISREG(sbuf.st_mode)) {
             fprintf(stderr, "[%s] File '%s' exists. Please apply '-f' to overwrite. Abort.\n", __func__, fnout);
             ret = 1;
             goto end;


jmarshall:
You could instead stat the filename and skip the test for FiFOs and sockets:

```c
struct stat sbuf;
if (stat(fnout, &sbuf) == 0 && !(S_ISFIFO(sbuf.st_mode) || S_ISSOCK(sbuf.st_mode))) { // …
```

But really this just highlights how unwise this sort of pre-flighting is. Ideally this could be refactored to use `sam_open_format(…, "wx", …)` instead, but I see `O_EXCL` fails for FIFOs unfortunately.

bam_sort.c
```diff
@@ -1506,9 +1506,7 @@ int bam_merge(int argc, char *argv[])
 
     hts_srand48(random_seed);
     if (!(flag & MERGE_FORCE) && strcmp(fnout, "-") != 0) {
-        FILE *fp = fopen(fnout, "rb");
-        if (fp != NULL) {
-            fclose(fp);
+        if (access(argv[optind], F_OK) == 0) {
```
jmarshall, line 1509: `fnout`!!!!

jkbonfield: Ugh!  That was dumb.   I think this is probably a cut/paste error when I redid the change.  It wouldn't let me pull from develop to update it due to a conflict, so I reset it pulled, and then hacked in the one/two lines of code again.   Wrongly ;)

jkbonfield:
I thought about delaying to the actual open, but yes O_EXCL breaks it and it's a much bigger change to make.  I also thought about stat, but have vague recollections of portability issues with the parameters.  Having said that we do appear to use it too, so it may be a better change.  I'll give it a whirl and AppVeyor will soon educate me either way. :-)

jmarshall:
Yeah, “ideally” was doing some heavy lifting there :smile:… Contra-indicated by the size of the change, and the fact that it wouldn't circumvent the problem for FIFOs anyway, alas.

jkbonfield:
I rewrote it to use stat, and have a check explicit for `IS_REG` rather than not fifo/socket.  I felt it's more appropriate that way around as it makes the error handling easier and we don't have to have separate logic for all the dumb cases like it's a directory.

The old code previously reported:

```
$ samtools111 merge dir *.sorted.bam
[bam_merge] File 'dir' exists. Please apply '-f' to overwrite. Abort.
```

Now it falls through to the file open code which gives a more meaningful "Is a directory" error.

jmarshall:
Good point about checking regular rather than checking not-FIFO/socket: e.g. it should probably suppress the message for character specials like `/dev/stdout` too, and this does that.

Curious why you dropped the `S_IFLNK` part, as that one seems like it does want the message — was this the portability concern?

This very source file already uses `S_ISDIR()`, so you might as well write this using `S_ISREG()`. And the commit message now needs rejigging.

jkbonfield:
In testing I realised it's redundant so removed it incase it confuses people.

The `S_IFLNK` type can only be returned using `lstat` and not `stat`, as the latter will follow the symlink and report the type of file it points to.  That's clearly what we want.

Good point on `S_ISREG()`.  I'd forgotten about those, but it's curious that this isn't documented on Linux's `stat` man page, only in `inode.7` strangely.

jmarshall:
Ah — very good point about `lstat` vs `stat`.

I tend to go straight to the [POSIX spec](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html) so didn't notice that the BSD-derived macOS man page doesn't mention them either. Annoying.

jkbonfield:
Think we got there in the end.  Thanks for your review John.

It's particularly pleasing to see it fixes bogus errors with /dev/stdout too, which I didn't even think about.  I wonder how many other tools have this issue.  I couldn't spot any at first grep.

valeriuo:
Merged as e77c7c69183e64f568ce6f0ad9f36d202c7d7434

Issue #1358, opened 2021-01-08
Clearing previous duplicate settings and tags with samtools markdup shouldn't require that fixmates was run with the mote scoring option.


ghuls:
#### Is your feature request related to a problem? Please specify.

Clearing previous duplicate settings and tags with samtools markdup shouldn't require that fixmates was run with the mate scoring option.

I just want to remove the duplicate flag so I can run a barcode aware mark duplicate tool.
It would be great if a barcode aware mark duplicate tool would be added to `samtools markdup` 

```bash
java -Dpicard.useLegacyParser=false -jar $EBROOTPICARD/picard.jar MarkDuplicates \
        -I "${input_bam_file}" \
        -O "${output_bam_file}" \
        -M "${output_bam_file%.bam}.mark_duplicates_metrics.txt" \
        -BARCODE_TAG CB \
        -REMOVE_DUPLICATES false \
        -OPTICAL_DUPLICATE_PIXEL_DISTANCE 100
```

https://github.com/broadinstitute/picard/issues/1579

whitwham:
Would you want markdup to do it or an entirely separate program?

ghuls:
Clearing the DUP flag without that the mate scoring options are (necessarily) present would be good if it is supported by `samtools markdup`.

The posibility to clean any flag might make sense too to have in another tool (at the moment I don't know which other flags would make sense to be able to clear/set besides the DUP flag).

barcode aware markdup option would make sense in samtools markdup itself.

whitwham:
Clearing DUP flags and tags when not about to do more duplicate marking is not really a job for **markdup**.  It seems like **view** would be the best place to add that.

Barcode awareness I can add.  I assume that means that if a read looks like a duplicate but has a different barcode then it is not a duplicate?

ghuls:
> Clearing DUP flags and tags when not about to do more duplicate marking is not really a job for **markdup**. It seems like **view** would be the best place to add that.

I wrongly assumed that `-c` would only clear the flags, but not run the rest of the markdup code itself. `view` would indeed be a better place for it (as it supports clearing AUX tags already).

> Barcode awareness I can add. I assume that means that if a read looks like a duplicate but has a different barcode then it is not a duplicate?

Yes. It is also important that the BAM tag which should be used for getting the UMI/BARCODE_TAG can be chosen.

The Picard implementation should be in this dir:
[...]

PR #1441, opened 2021-06-02
Add flag set and unset options to view.

whitwham:
Add options to set and unset flags.  Along with the existing option to remove aux tags this will provide the duplicate mark clearing requested in #1358.


whitwham:
Leaving this as draft as I am not convinced this is the best approach and I am putting this up for discussion.

This PR also needs documentation and test cases writing.

whitwham:
Could do.  If we are going to have long options then making them consistent would be good.

valeriuo:
Merged as 12520cf5e001371d2ce5507e713523b0bbac7dd5 to a7449a0472f977e7a71695245c8ec284e87fa6ee

jmarshall:
Noticing slightly late… making them consistent with mpileup's existing options and #1442's proposed long options would mean that they should be `--add-flags`/`--remove-flags` i.e. plural. 🤔

valeriuo:
Perhaps it can be corrected in #1442 upon rebasing.

jmarshall:
Yep, I'll be rebasing that one this morning now that this is merged. Good idea.

Issue #217, opened 2014-05-27
overloading of `samtools view -s` considered harmful


jrandall:
I've spent the past 8 hours working on the `samtools view` documentation and am nearly done.  However, I can't get my head around `samtools view -s`. In my opinion it is unnecessary and somewhat absurd to overload the `-s` option of `samtools view` such that the integer part seeds the random number generator and the fractional part sets the fraction of templates to subsample. 

This is not clever - it is awkward and confusing. I suggest a new long option `--seed` to replace the seed part of `-s`.

PR #1442, opened 2021-06-08
Add long options to `samtools view`

jmarshall:
[Another PR that's been gestating for rather too long…]

This adds long option equivalents for all of `view`'s command-line options. The added expressiveness of the long names I think clarifies some of the usage explanations (while staying within 80 characters!).

It also adds `‑‑region-file` and `‑‑regions-file` as synonyms for the `-M -L` combination. Along with the `‑‑target-file`/`‑‑targets-file` long forms for plain `-L`, this means `samtools view` has the same long options for jump-to-region and stream-filtering-by-region as bcftools. And hopefully highlighting `‑‑region-file` as an option will help raise the visibility of `-M`.

Also `-s` subsampling is split into two long options, thus finally fixing #217.

Proposed usage display:

```
Usage: samtools view [options] <in.bam>|<in.sam>|<in.cram> [region ...]

Output options:
  -b, --bam                  Output BAM
  -C, --cram                 Output CRAM (requires -T)
  -1, --fast                 Use fast BAM compression (implies --bam)
  -u, --uncompressed         Uncompressed BAM output (implies --bam)
  -h, --with-header          Include header in SAM output
  -H, --header-only          Print SAM header only (no alignments)
      --no-header            Print SAM alignment records only [default]
  -c, --count                Print only the count of matching records
  -o, --output FILE          Write output to FILE [standard output]
  -U, --unoutput FILE, --output-unselected FILE
                             Output reads not selected by filters to FILE
Input options:
  -t, --fai-reference FILE   FILE listing reference names and lengths
  -M, --use-index            Use index and multi-region iterator for regions
      --region[s]-file FILE  Use index to include only reads overlapping FILE
  -X, --customized-index     Expect extra index file argument after <in.bam>

Filtering options (Only include in output reads that...):
  -L, --target[s]-file FILE  ...overlap (BED) regions in FILE
  -r, --read-group STR       ...are in read group STR
  -R, --read-group-file FILE ...are in a read group listed in FILE
  -N, --qname-file FILE      ...whose read name is listed in FILE
  -d, --tag STR1[:STR2]      ...have a tag STR1 (with associated value STR2)
  -D, --tag-file STR:FILE    ...have a tag STR whose value is listed in FILE
  -q, --min-MQ INT           ...have mapping quality >= INT
  -l, --library STR          ...are in library STR
  -m, --min-qlen INT         ...cover >= INT query bases (as measured via CIGAR)
  -e, --expr STR             ...match the filter expression STR
  -f, --all-flags INT        ...have all of the FLAGs in INT present
  -F, --excl[ude]-flags INT  ...have none of the FLAGs in INT present
  -G, --no-all-flags INT     EXCLUDE reads with all of the FLAGs in INT present
      --subsample FLOAT      Keep only FLOAT fraction of templates/read pairs
      --subsample-seed INT   Influence WHICH reads are kept in subsampling [0]
  -s INT.FRAC                Same as --subsample 0.FRAC --subsample-seed INT

Processing options:
  -x, --remove-tag STR       Strip tag STR from reads (option may be repeated)
  -B, --remove-B             Collapse the backward CIGAR operation

General options:
  -?, --help   Print long help, including note about region specification
  -S           Ignored (input format is auto-detected)
      --no-PG  Do not add a PG line
      --input-fmt-option OPT[=VAL]
               Specify a single input file format option in the form
               of OPTION or OPTION=VALUE
  -O, --output-fmt FORMAT[,OPT[=VAL]]...
               Specify output format (SAM, BAM, CRAM)
      --output-fmt-option OPT[=VAL]
               Specify a single output file format option in the form
               of OPTION or OPTION=VALUE
  -T, --reference FILE
               Reference sequence FASTA FILE [null]
  -@, --threads INT
               Number of additional threads to use [0]
      --write-index
               Automatically index the output files [off]
      --verbosity INT
               Set level of verbosity

Notes:

1. This command now auto-detects the input format (BAM/CRAM/SAM).
   Further control over the CRAM format can be specified by using the
   --output-fmt-option, e.g. to specify the number of sequences per slice
   and to use avoid reference based compression:

	samtools view -C --output-fmt-option seqs_per_slice=5000 \
	   --output-fmt-option no_ref -o out.cram in.bam

   Options can also be specified as a comma separated list within the
   --output-fmt value too.  For example this is equivalent to the above

	samtools view --output-fmt cram,seqs_per_slice=5000,no_ref \
	   -o out.cram in.bam

2. The file supplied with `-t' is SPACE/TAB delimited with the first
   two fields of each line consisting of the reference name and the
   corresponding sequence length. The `.fai' file generated by 
   `samtools faidx' is suitable for use as this file. This may be an
   empty file if reads are unaligned.

3. SAM->BAM conversion:  samtools view -bT ref.fa in.sam.gz

4. BAM->SAM conversion:  samtools view -h in.bam

5. A region should be presented in one of the following formats:
   `chr1', `chr2:1,000' and `chr3:1000-2,000'. When a region is
   specified, the input alignment file must be a sorted and indexed
   alignment (BAM/CRAM) file.

6. Option `-u' is preferred over `-b' when the output is piped to
   another samtools command.

7. Option `-M`/`--use-index` causes overlaps with `-L` BED file regions and
   command-line region arguments to be computed using the multi-region iterator
   and an index. This increases speed, omits duplicates, and outputs the reads
   as they are ordered in the input SAM/BAM/CRAM file.

8. Options `-L`/`--target[s]-file` and `--region[s]-file` may not be used
   together. `--region[s]-file FILE` is simply equivalent to `-M -L FILE`,
   so using both causes one of the specified BED files to be ignored.
```

### DIFF
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -381,24 +381,30 @@ is specified in the same was as the
 .B "--add-flag"
 option.
 .TP
-.BI "-s " FLOAT
-Output only a proportion of the input alignments.
+.BI "--subsample " FLOAT
+Output only a proportion of the input alignments, as specified by 0.0 \(<=
+.I FLOAT
+\(<= 1.0, which gives the fraction of templates/pairs to be kept.
 This subsampling acts in the same way on all of the alignment records in
 the same template or read pair, so it never keeps a read but not its mate.
-.IP
-The integer and fractional parts of the
-.BI "-s " INT . FRAC
-option are used separately: the part after the
-decimal point sets the fraction of templates/pairs to be kept,
-while the integer part is used as a seed that influences
+.TP
+.BI "--subsample-seed " INT
+Subsampling seed used to influence
 .I which
 subset of reads is kept.
-.IP
 .\" Reads are retained based on a score computed by hashing their QNAME
 .\" field and the seed value.
 When subsampling data that has previously been subsampled, be sure to use
 a different seed value from those used previously; otherwise more reads
 will be retained than expected.
+[0]
+.TP
+.BI "-s " FLOAT
+Subsampling shorthand option:
+.BI "-s " INT . FRAC
+is equivalent to
+.BI "--subsample-seed " INT " --subsample
+.RI 0. FRAC .
 .TP
 .BI "-@ " INT ", --threads " INT
 Number of BAM compression threads to use in addition to main thread [0].
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -105,9 +105,9 @@ alignments that match certain criteria.
 The
 .BR -x ,
 .BR -B ,
-.B --add-flag
+.BR --add-flags ,
 and
-.B --remove-flag
+.B --remove-flags
 options modify the data which is contained in each alignment.
 
 The
@@ -367,18 +367,18 @@ Read tag to exclude from output (repeatable) [null]
 .BR -B ", " --remove-B
 Collapse the backward CIGAR operation.
 .TP
-.BI "--add-flag " FLAGS
+.BI "--add-flags " FLAG
 Adds flag(s) to read.
 .I FLAG
 can be specified in hex by beginning with `0x' (i.e. /^0x[0-9A-F]+/),
 in octal by beginning with `0' (i.e. /^0[0-7]+/), as a decimal number
 not beginning with '0' or as a comma-separated list of flag names.
 .TP
-.BI "--remove-flag " FLAG
+.BI "--remove-flags " FLAG
 Remove flag(s) from read.
 .I FLAG
-is specified in the same was as the
-.B "--add-flag"
+is specified in the same way as with the
+.B "--add-flags"
 option.
 .TP
 .BI "--subsample " FLOAT
@@ -501,7 +501,7 @@ samtools view -d RG:grp2 -o /data_folder/data.rg2_only.bam /data_folder/data.bam
 .IP o 2
 Remove the actions of samtools markdup.  Clear the duplicate flag and remove the \fBdt\fR tag, keep the header.
 .EX 2
-samtools view -h --remove-flag DUP -x dt -o /data_folder/dat.no_dup_markings.bam /data_folder/data.bam
+samtools view -h --remove-flags DUP -x dt -o /data_folder/dat.no_dup_markings.bam /data_folder/data.bam
 .EE
 
 .SH AUTHOR
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -158,27 +158,32 @@ Output all alignments.
 
 .SH OPTIONS
 .TP 10
-.B -b
+.BR -b ", " --bam
 Output in the BAM format.
 .TP
-.B -C
+.BR -C ", " --cram
 Output in the CRAM format (requires -T).
 .TP
-.B -1
+.BR -1 ", " --fast
 Enable fast BAM compression (implies -b).
 .TP
-.B -u
+.BR -u ", " --uncompressed
 Output uncompressed BAM. This option saves time spent on
 compression/decompression and is thus preferred when the output is piped
 to another samtools command.
 .TP
-.B -h
+.BR -h ", " --with-header
 Include the header in the output.
 .TP
-.B -H
+.BR -H ", " --header-only
 Output the header only.
 .TP
-.B -c
+.B --no-header
+When producing SAM format, output alignment records but not headers.
+This is the default; the option can be used to reset the effect of
+.BR -h / -H .
+.TP
+.BR -c ", " --count
 Instead of printing the alignments, only count them and print the
 total number. All filter options, such as
 .BR -f ,
@@ -187,14 +192,14 @@ and
 .BR -q ,
 are taken into account.
 .TP
-.B -?
+.BR -? ", " --help
 Output long help and exit immediately.
 .TP
-.BI "-o " FILE
+.BI "-o " FILE ", --output " FILE
 Output to
 .I FILE [stdout].
 .TP
-.BI "-U " FILE
+.BI "-U " FILE ", --unoutput " FILE ", --output-unselected " FILE
 Write alignments that are
 .I not
 selected by the various filter options to
@@ -203,7 +208,7 @@ When this option is used, all alignments (or all alignments intersecting the
 .I regions
 specified) are written to either the output file or this file, but never both.
 .TP
-.BI "-t " FILE
+.BI "-t " FILE ", --fai-reference " FILE
 A tab-delimited
 .IR FILE .
 Each line must contain the reference name in the first column and the length of
@@ -215,7 +220,7 @@ defines the order of the reference sequences in sorting. If you run:
 can be used as this
 .IR FILE .
 .TP
-.BI "-T " FILE
+.BI "-T " FILE ", --reference " FILE
 A FASTA format reference
 .IR FILE ,
 optionally compressed by
@@ -256,11 +261,11 @@ The usage of a BED file is optional and its path has to be preceded by
 .BR -L
 option.
 .TP
-.BI "-N " FILE
+.BI "-N " FILE ", --qname-file " FILE
 Output only alignments with read names listed in
-.I FILE.
+.IR FILE .
 .TP
-.BI "-r " STR
+.BI "-r " STR ", --read-group " STR
 Output alignments in read group
 .I STR
 [null].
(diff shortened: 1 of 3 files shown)


daviesrob:
Looks good in the main.  I'm not sure if I like `--all-flags` and `--no-all-flags` though.  Would `--required-flags` and `--avoid-flags` be better?

jmarshall:
I was pretty unimpressed that I couldn't come up with anything better than `--no-all-flags`!

`--require-flags` is probably an improvement, but while I like `--avoid-flags` better than `--no-all-flags` I'm not sure it conveys the “all“ aspect of the test — `avoid` sounds like “avoid any of these” which would make it a synonym of `--exclude-flags` (I think).

Let's all mull those ones over for a while…

jmarshall:
I think `--require-flags` is better, and I've realised how to describe `-G` positively and clearly (obvious in retrospect):

```
Filtering options (Only include in output reads that...):
[…]
  -f, --require-flags INT    ...have all of the FLAGs in INT present
  -F, --excl[ude]-flags INT  ...have none of the FLAGs in INT present
  -G, --something-flags INT  ...have any of the FLAGs in INT not present
```

Hmmm… what for that last something…
`--some-unset-flags`? `--absent-flags`? `--unset-flags`? `--any-unset-flags`?
Frankly it's pretty hard to do better than `--not-all-flags`…

jmarshall:
> Firstly "not present" is better expressed as "absent", and secondly the double negative may be better expressed as a positive.

I had it as “absent” but felt that it was too easy for the eye to skip over the half-a-word difference from the other lines without noticing. I'll call this a vote (and I'm inclined to agree) for the status quo:

```
Filtering options (Only include in output reads that...):
[…10 other filtering options…]
  -f, --require-flags INT    ...have all of the FLAGs in INT present
  -F, --excl[ude]-flags INT  ...have none of the FLAGs in INT present
  -G, --something-flags INT  EXCLUDE reads with all of the FLAGs in INT present
```

`--exclude-any/all-flags` is not bad, but `--excl[ude]-flags` is so-named because it is the same as pileup's `--excl-flags` option. Hmmm…

jkbonfield:
Compatibility with mpileup is good, but I note it has also reversed the filtering language to describe what to skip, rather than what to keep.  Also I wouldn't be averse to changing mpileup's description if we come up with something better, but changing long-opts is obviously not good unless we add aliases.  That does mean of course `--excl-flags`, `--exclude-flags` and `--exclude-any-flags` could all be synonyms if we wished.

It just feels odd that we have a long-opt `--exclude-flags` with a description of "only include in output reads that have none of the FLAGs in INT present".  But then again maybe phrasing it in both directions means people can consider it from either side of the fence and go with what works for their brain.

jmarshall:
That phrasing was introduced by you in 8b779df31fbcea305923e293070d4e41233070eb.

The `samtools view` usage has always phrased these filter options in terms of what to keep. This PR doesn't change the language much, it mostly just factors “only include reads” out of every line into the subheading. There are 13 filtering options here, and IMHO the keep phrasing is overall definitely the right one — e.g. consider this 🤮 for `-d`:

> `-d, --tag STR1[:STR2`]    Skip reads that do not have a tag STR1 (or that do but its value is not STR2)

That `mpileup`'s usage describes similar options in terms of skipping is a separate minor niggle.

Anyway, that mostly unchanged usage phrasing is not the major problem here. The thing that's holding this PR up is what words to choose for `--require-flags` and `--not-all-flags` / `--exclude-all-flags` / something.

jkbonfield:
As a side note, if we're pointing the finger of blame at mpileup bad phrasing, my pet peeve has always been the overlap removal description:

>   -x, --ignore-overlaps   disable read-pair overlap detection

By default, bases from overlapping reads in a read pair are ignored (filtered out).  To include them instead, we enable `--ignore-overlaps`.  Boggle!

jkbonfield:
> Anyway, that mostly unchanged usage phrasing is not the major problem here. The thing that's holding this PR up is what words to choose for `--require-flags` and `--not-all-flags` / `--exclude-all-flags` / something.

My choice would still be for `--exclude-all-flags` as it matches the description the best.

jmarshall:
OMG James, this PR is about `samtools view` only! I'd just like a little help workshopping the last couple of option words so that this usability improvement can get into the hands of users.

jmarshall:
Rebased now that #1441 has been merged. Still no final epiphany on the last couple of long options, and the new `--add-flag`/`--remove-flag` haven't been adjusted to match the other `--foo-flags` option names' plurality (but I did fix a couple of man page typos there).

whitwham:
Just to get things moving, how about --reject-all-flags?

daviesrob:
... or we could remove the long opt for `-G` for now, so we can merge the rest and go back to fix it when inspiration hits ...

jmarshall:
Yep, happy to do that just to get this moving. With `-f`/`--require-flags` and just `-G` for now? And should the new `--add/remove-flag` be pluralised?

whitwham:
> 
> And should the new `--add/remove-flag` be pluralised?

Yes, probably.

daviesrob:
Actually, it may be best to leave all of `-f`, `-F` and `-G` for now, as I see we currently have:

```
amplicon_stats.c:        {"flag-require", required_argument, NULL, 'f'},
amplicon_stats.c:        {"flag-filter", required_argument, NULL, 'F'},
bam_plcmd.c:        {"rf", required_argument, NULL, 1},   // require flag
bam_plcmd.c:        {"ff", required_argument, NULL, 2},   // filter flag
bam_plcmd.c:        {"incl-flags", required_argument, NULL, 1},
bam_plcmd.c:        {"excl-flags", required_argument, NULL, 2},
coverage.c:        {"rf", required_argument, NULL, 1}, // require flag
coverage.c:        {"ff", required_argument, NULL, 2}, // filter flag
coverage.c:        {"incl-flags", required_argument, NULL, 1}, // require flag
coverage.c:        {"excl-flags", required_argument, NULL, 2}, // filter flag
stats.c:        {"required-flag", required_argument, NULL, 'f'},
stats.c:        {"filtering-flag", required_argument, NULL, 'F'},
```
so it's probably best to avoid adding another variant for now.  We can make a follow-up pull request to try to unify them, as much as we can.

jmarshall:
I think the situation is not as irregular as all that:

Skip when: | (FLAG & _N_) != _N_ | (FLAG & _N_) != 0 | (FLAG & _N_) == 0 | (FLAG & _N_) == _N_
-|-|-|-|-
ampliconstats | -f/--flag-require | -F/--flag-filter |  | 
coverage |  | --ff/--excl-flags | --rf/--incl-flags | 
mpileup |  |--ff/--excl-flags | --rf/--incl-flags | 
stats | -f/--required-flag | -F/--filtering-flag |  | 
view | -f/`--require-flags` | -F/`--excl-flags` | | -G

(With the proposed `view` additions shown `in code`.)

I had previously categorised only `coverage` and `mpileup`; adding the other two into the table shows:

* Using _require_ as in your `--require-flags` suggestion is clearly the right choice
* `-G` is a weird one-off so we should be comfortable in ignoring it for long-option purposes
* _filter[ing]_ is not a great choice (as it implies nothing about the test's meaning), so following `--excl[ude]` is better

So I think adding `--require-flags` and `--excl[ude]-flags` (but no long form for `-G`) is a step towards what a desired unification would look like.

Need to decide whether to add them as verbs (`‑‑require-flags ‑‑excl[ude]-flags`) or adjectives (`‑‑required-flags ‑‑excl[uded]-flags`) or just implement both. The mpileup options were originally abbreviations of the verbs (see e9ea1ba3dff6c3d68023b4669d3ac9235f2baaac) but in a way using the adjectival form à la `stats` is a useful distinction from the modifying verbs of `‑‑add/remove-flag[s]`.

(I can never remember which option I want to use and have to look it up each time. What I would really like to have for this is to come up with some `[+-=].*` syntax something like `chmod(1)` symbolic permissions so that all these options could be unified into a single `-f`/`--flags` option — at which point these legacy options would become somewhat moot.)

jmarshall:
Reading over the existing filtering long options in other subcommands, I think actually the simpler verb form for these is generally the way to go. So for your consideration, I've pushed a version of this that adds `--require-flags` for `-f`, `--excl[ude]-flags` for `-F`, no long option for `-G`, and pluralises the `--add-flags`/`--remove-flags` options.

Hopefully this may be suitable for merging and then we can try to approach cross-subcommand nirvana later :smile:

(Note that GNU `getopt_long` and probably other implementations accept an unambiguous prefix of long options, so all these options can in fact be spelt as either `--add-flag` or `--add-flags` as the user prefers… though they shouldn't really do that in scripts unless they like living on the edge!)

daviesrob:
I guess we can live with `--require-flags` and sort out the the other subcommands later.  incidentally, did you forget to move it to its correct position in the alphabetically-ordered `lopts[]` array?

jmarshall:
> incidentally, did you forget to move it to its correct position in the alphabetically-ordered `lopts[]` array?

D'oh. Thanks.

PR #1446, opened 2021-06-11
[coverage] Fix option -l

### LINKED ISSUE #1443: Typo in samtools coverage help
Hi,

Just found a typo in help text.
The line
```
-l, --min-read-len INT  ignore reads shorter than INT bp [0]
```
should read
```
-L, --min-read-len INT  ignore reads shorter than INT bp [0]
```
https://github.com/samtools/samtools/blob/8e8b07400fd7d148e4243f38e7e0fc3cb9e124e1/coverage.c#L114
https://github.com/samtools/samtools/blob/8e8b07400fd7d148e4243f38e7e0fc3cb9e124e1/coverage.c#L330

valeriuo:
Although advertised in the man page, the option `-l` was non-functional, due to wrong internal encoding. This is now fixed.

Fixes #1443

### DIFF
--- a/coverage.c
+++ b/coverage.c
@@ -327,7 +327,7 @@ int main_coverage(int argc, char *argv[]) {
         {"incl-flags", required_argument, NULL, 1}, // require flag
         {"excl-flags", required_argument, NULL, 2}, // filter flag
         {"bam-list", required_argument, NULL, 'b'},
-        {"min-read-len", required_argument, NULL, 'L'},
+        {"min-read-len", required_argument, NULL, 'l'},
         {"min-MQ", required_argument, NULL, 'q'},
         {"min-mq", required_argument, NULL, 'q'},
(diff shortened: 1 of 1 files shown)


jmarshall:
Is it better to change the behaviour to match the documentation, or to update the documentation to match the existing behaviour?

Note that `-l` was changed to `-L` (but not in the usage, leading to this issue) in d4566c32c6f69ece97a173bb72a1e731a4f2c8b4. There's no explicit conversation about this, but the commit message says “Add long version of most options in concordance with mpileup” so I suspect `-l` was changed to `-L` so that it would use a different short letter from mpileup's `-l`.

jkbonfield:
IMO making it use `-l` is preferably as mpileup's `-l` does a different thing, while `depth -l` is exactly the same functionality as this.  I assume the change was in error.

Issue #1226, opened 2020-04-14
ampliconclip ideas and todo list


whitwham:
Since ampliconclip is new and under development for the virus tracking project this is a place to put ideas and a reminder of things to do.

Initial list (in no real order)

- man page
- tests
- both ends clipping
- calmd fix for clipped reads

whitwham:
Options:
Add OA tags to keep the original alignment data.
Remove MD and NM tags when they become invalid.

whitwham:
The calmd fix is not done and not going to be done.  It is probably best left to calmd itself.  The rest have all been done.

nathanhaigh:
I've just come to this tool and have a few suggestions which might help others trying to figure out how to run the tool.

I think the word "amplicon" is being somewhat misused in the help: according to the Sequence Ontology (http://www.sequenceontology.org/browser/current_release/term/SO:0000006) amplicon is the PCR product which will include the bit covered by primers. 

- It's unclear from the help what should actually be in the BED file. It says "bedfile of amplicons to be removed" but to me, it's the amplicon primers
- It's unclear what `--both-ends` and `--strand` achieve. I kind of assume the former requires both the 5' and 3' end of a read to match the boundary of one of the primers defined in the BED file

whitwham:
> I think the word "amplicon" is being somewhat misused in the help: according to the Sequence Ontology (http://www.sequenceontology.org/browser/current_release/term/SO:0000006) amplicon is the PCR product which will include the bit covered by primers.

We know, the list of potential names is [here](https://github.com/samtools/samtools/pull/1219#issuecomment-609649307).  It was all to do with the amplicon based sequencing of Sars-Cov-2 and we were in a hurry.  We are probably not going to change it now.
 
>     * It's unclear from the help what should actually be in the BED file. It says "bedfile of amplicons to be removed" but to me, it's the amplicon primers

That is a good point.  I should include what entries are being used and why.
 
>     * It's unclear what `--both-ends` and `--strand` achieve. I kind of assume the former requires both the 5' and 3' end of a read to match the boundary of one of the primers defined in the BED file

I didn't think `--both-ends` was unclear.  Maybe the man page should read "Clip from the 5' and the 3' ends"?

`--strand` could do with more explanation, probably in conjunction with what we use from the BED file .
 
>     * It would be useful to state how much "slop" there is in matching the boundary coordinates of the read alignment with a primer region. Looking at the code, it looks like this is `tol = 5`

[...]

PR #1448, opened 2021-06-11
Update ampliconclip documentation.

whitwham:
Update the ampliconclip documentation to be clearer and more informative about what it does.  Done at the suggestion of @nathanhaigh in issue #1226.


daviesrob:
Generally looks better, although I noticed that there are a few places in the usage message where it says "amplicons primers" that you might want to fix.

whitwham:
So it did.  Chronic sleeplessness strikes again.  Fixed now.

PR #1456, opened 2021-06-25
Added a tolerance option and removed dead code.

whitwham:
The inbuilt matching tolerance of 5 now has an option to change it.  Code
left over from an earlier change has been deleted.

### DIFF
--- a/bam_ampliconclip.c
+++ b/bam_ampliconclip.c
@@ -993,7 +993,7 @@ int amplicon_clip_main(int argc, char **argv) {
             case 1012: param.rejects_file = optarg; break;
             case 1013: param.oa_tag = 1; break;
             case 1014: param.del_tag = 0; break;
-            case 1015: param.tol = atoi(optarg);  if (param.tol < 0) param.tol = 0; break;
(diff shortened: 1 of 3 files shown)


bam_ampliconclip.c
```diff
@@ -1113,6 +993,7 @@ int amplicon_clip_main(int argc, char **argv) {
             case 1012: param.rejects_file = optarg; break;
             case 1013: param.oa_tag = 1; break;
             case 1014: param.del_tag = 0; break;
+            case 1015: param.tol = atoi(optarg);  if (param.tol < 0) param.tol = 0; break;
```
valeriuo, line 996: Perhaps it would be better to go back to the default (5), instead of 0.

whitwham: My thinking was that 0 was the nearest valid value to a negative number.  Going back to the default is reasonable too.  I should probably make it print out an error message when it changes the value.

valeriuo (approved):

valeriuo:
Merged as 5f186d0dc868ed46e07e80a1569e219ea7a70f4a

PR #1464, opened 2021-07-07
Document "endpos" filter expression

jmarshall:
Companion PR to samtools/htslib#1307, adding the new item to the documentation.

### DIFF
--- a/doc/samtools.1
+++ b/doc/samtools.1
@@ -999,6 +999,7 @@ Valid variable names and their data types are:
 
 .TS
 lb l l .
+endpos	int	Alignment end position (1-based)
 flag	int	Combined FLAG field
 flag.paired	int	Single bit, 0 or 1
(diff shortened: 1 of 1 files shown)


jkbonfield:
Is it actually the last base, or the first base after the end?   Caveats aside it's implemented as `pos + bam_cigar2rlen(...)`.  I'm assuming rlen is the number of reference bases with an aligned read against them so a 1M cigar would return 1, so a read at pos 100 with 1M would yield endpos as 101.  I haven't called the code to double check though.

I think that would make it "1-based exclusive", but I've never really liked those terms.  It's fine to use them, but better to also explicitly state what it means.

jmarshall:
Yep, the +1 and -1 cancel out, as mentioned in the htslib commit message and probably should have been mentioned on this one as well :smile:

PR #1466, opened 2021-07-09
Quote test case --index-format arguments

jmarshall:
Quote these `i*i*` arguments so they aren't misinterpreted as filename globs, which can lead to test failures when there are files matching `i*i*` in the source directory.

D'oh :smile:


jkbonfield:
D'oh! indeed.

Thanks

jmarshall:
Fortunately people building the recent release will be doing so from their freshly unpacked source directory that probably won't have any extra homegrown files in it.

PR #1311, opened 2020-09-10
Initial draft of base modifications in mpileup.

jkbonfield:
These are enabled with -M or --output-mods.  This adds them to any
sequence bases (both matching ref and insertions).  This can make
parsing more complex, especially insertions, so --no-output-ins-mods
can be used to turn them off there.

Also extended mpileup to provide further control of output format via
--no-output-ins and --no-output-del, which may be used to make parsing
the sequence easier for specific applications that do not wish to
handle indels.

They both have two levels, to change from eg CCC-3NNNGCC to CCC-3GCC
to CCCGCC.  This will greatly simplify scripts that are simply trying
to count bases.

Similarly --no-output-ends removes the ^MQUAL and $ markup, which are
pretty unreliable frankly as they don't always appear on range
requests, when soft-clips are present, or when bases have been
quality filtered via -Q.

PR #1473, opened 2021-07-29
"Samtools consensus" subcommand

jkbonfield:
This is something of a work in progress, but could do with feedback still.

Some thoughts.

- It provides a way to generate a basic consensus from a BAM(et al) file.  This can be pure or with ambiguity codes, and be a basic simple counting method or a more rigorous probabilistic analysis (based on Gap5's consensus algorithm).

- It is reference coordinate based.  Thus 100% deletion generates "*" and insertions are ignored.  This makes it deal for generating reference sequences that can be used and embedded within a CRAM file.
  - If you do this, you'll probably want -a to prepend/append Ns and keep the reference coordinates in sync.
  - However... do we also have a need for a more general consensus that includes insertions?

- It supports methylation, thus this PR also neccessitates inclusion of all the commits in #1311.  Once that's merged I can rebase this accordingly.  This was infact one of the reasons I wrote it: as a test bed for the base modification PR to see if it's fit for purpose.
  - TODO: test it actually works!  Fundamentally I don't have sufficient so I'm flying blind here.  Anyone able to help with data files containing the new methylation Mm tag?  I've been using a grand total of 2 reads. :/

- There is a format option between fasta / fastq, but it also has pileup format.  That's probably due to be culled, but it's a useful debugging tool for now and to judge how it differs to `samtools mpileup`.

- TODO:
  - Man page
  - Tests
  - What options do we need with regards to filtering, eg min/max qual, mapping qual, etc?


daviesrob:
I think basic support for insertions would be good, even if it's only most-frequent-wins.  We should have an interface so an improved method can be selected later, when we come up with one.  I also think the default for a deleted base should be to output nothing, with an option for pad characters.  This makes it less likely that someone will accidentally treat the pads as N's, which is usually not the correct way they should be interpreted.

daviesrob:
Rebased and removed trailing spaces...

bam_consensus.c
```diff
@@ -0,0 +1,1522 @@
...
+        // NB no support for opts->use_qual yet here.  We either need
+        // a better bam_plp_insertion or inline it here with added
+        // quality value processing.
+        int il = bam_plp_insertion(p, &is, NULL), j;
+        if (il < ins_len)
+            continue;
+
+        for (j = 0; j < il; j++) {
+            if (is.s[j] == '*') {
+                ins[j][16] += 8;
```
daviesrob, line 210: There's a compiler error on this line:
```
bam_consensus.c:210:28: error: array subscript 16 is above array bounds of 'int[16]' [-Werror=array-bounds]
  210 |                 ins[j][16] += 8;
      |                            ^~
```

Should the array size be 17, or should this be index 0?  (Changing to 0 would also require adjustments below).

PR #1484, opened 2021-08-24
Add mpileup --output-BP-5 option.

### LINKED ISSUE #1481: Base positions on reads mapped in reverse strand in the pileup format
I would like to get base mismatch positions on reads (5'->3'). It can be done easily in the pileup format with the option -O enabled. However, in case of reads mapped to reverse strand, it is difficult because the positions are still numbered in the  forward strand direction, not in 5'->3' direction of the target strand. 
mpileup supports read length output or calculate the positions on reads in 5'->3' direction.

jkbonfield:
Normally the sequence base positions output are in left to right orientation, as shown in the SAM file.  The --output-BP-5 option outputs positions as based on the original 5' to 3' orientation, which helps identification of sequence position-specific biases.

Fixes #1481


valeriuo (approved):

valeriuo:
Merged as ae2a60315bd798eec6a0da5265a3e785cbd205fa

PR #1485, opened 2021-08-26
Add support for the SRA "@RUN.1234 proper:name:100:200" FASTQ syntax

jkbonfield:
We don't know why the name got moved from the first field to the second field, but the new "-N" option to "samtools import" now parses this.  Needs the corresponding htslib PR too.


daviesrob:
Rebased, with a trivial change to the man page.  I swapped `.B --name2, -N` to `.B -N, --name2` to match other cases where we list both short and long options.

PR #1490, opened 2021-08-31
Flush output samFiles before printing error messages

jmarshall:
When output is to the terminal (common when e.g. `samtools view` is used to view SAM), flush it to ensure error messages are not obscured by buffered output that is printed later when the output stream is closed.

For example, with samtools 1.13 on an intentionally corrupted BAM file:

```
$ samtools-1.13 view ex2_corrupt.bam
EAS56_57:6:190:289:82   73      chr1    100     73      35M     =       100     0       AGGGGTGCAGAGCCGAGTCACGGGGTTGCCAGCAC     <<<<<<;<<<<<<<<<<;<<;<<<<;8<6;9;;2;     MF:i:64 Aq:i:0  NM:i:0  UQ:i:0  H0:i:1  H1:i:0
[…]
EAS56_61:1:119:880:781  147     chr2    1312    99      35M     =       1157    -190    ACAAsamtools view: error reading file "ex2_corrupt.bam"
ATCTGCGCTTGTACTTCTAAATCTATAACAA ;8<<;<<<<:<84<<<<:<<<<<<<<<<<<<5<<<     MF:i:18 Aq:i:45 NM:i:1  UQ:i:27 H0:i:0  H1:i:1
[…
 …approx 170 more lines of SAM output…
 …]
EAS114_26:7:37:79:581   83      chr2    1533    68      35M     =       1349    -219    TTTTTTTTTTTTTTTTTTTTTTTCATGCCAGAAAA     3,,,===6===<===<;=====-============     MF:i:18 Aq:i:27 NM:i:2  UQ:i:23 H0:i:0  H1:i:1
$
```

With this patch (which requires PR samtools/htslib#1326) the error message appears after all the SAM output, prominently just before the following `$` prompt.

This PR fixes the issue for a couple of likely candidate subcommands, `samtools view` in particular.

(`vprint_error_core()` is able to flush `stdout` itself because there is a well-known global variable referring to standard output's `FILE *`. So if _sam_utils.c_ (or another source file) provided a `samFile *sam_stdout` global variable that the various subcommands could set to their output handles that point to standard output, `vprint_error_core()` would be able to `sam_flush()` it itself too. But this is probably more trouble than it's worth, as care would have to be taken to set the global to NULL when the subcommand's output handle was closed.)

### DIFF
--- a/bam_fastq.c
+++ b/bam_fastq.c
@@ -480,6 +480,7 @@ static bool init_state(const bam2fq_opts_t* opts, bam2fq_state_t** state_out)
                     return false;
                 }
                 set_sam_opts(state->hstdout, state, opts);
+                autoflush_if_stdout(state->hstdout, "-");
             }
             state->fpr[i] = state->hstdout;
         }
@@ -546,6 +547,7 @@ static bool destroy_state(const bam2fq_opts_t *opts, bam2fq_state_t *state, int*
         }
     }
     if (state->hstdout) {
+        release_autoflush(state->hstdout);
         if (sam_close(state->hstdout) < 0) {
             print_error_errno("bam2fq", "Error closing STDOUT");
             valid = false;
@@ -787,7 +789,7 @@ static bool bam2fq_mainloop(bam2fq_state_t *state, bam2fq_opts_t* opts)
     while (true) {
         int res = sam_read1(state->fp, state->h, b[n]);
         if (res < -1) {
-            fprintf(stderr, "[bam2fq_mainloop] Failed to read bam record.\n");
+            print_error("bam2fq", "Failed to read bam record");
             goto err;
         }
         at_eof = res < 0;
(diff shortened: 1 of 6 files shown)


jmarshall:
Now thinking that with a couple of wrapper functions the _sam_utils.c_ global variable approach is very practical. Will revisit this once the HTSlib PR is merged.

jmarshall:
IMHO much improved implementation that just needs a couple of calls when output is opened/closed rather than spraying save-`errno`+`sam_flush()` in front of every error message call.

valeriuo (approved):

valeriuo:
Merged as a78376c5fe659bd53819fc0a960ec699830a3ecb

jmarshall:
Thanks for merging. FYI for future reference:

Having rebased the PR, if you force-push to the PR's branch (i.e., **flush** in **jmarshall/samtools**; this is why GitHub added the “Allow edits by maintainers” button) and afterwards push to **develop**, the PR will be marked as Merged (purple) rather than Closed (red). Then everyone can take advantage of GitHub's commit tracking instead of you needing to comment with “Merged as xyz”.

PR #1512, opened 2021-10-05
Unmap filtered alignments in samtools view

### LINKED ISSUE #1469: add option to samtools view to unmap aligned reads rather than remove them
Pretty self-explanatory.  When using -f/F/G or any other filters, I want to keep the reads in the bam, just render them unaligned.  This would be useful for downstream analyses that use "total reads" in the denominator, as well as downsampling bams that have already been "filtered".

whitwham:
Set the UNMAP flag on filtered alignments and write them to the same output as the unfiltered alignments.  Closes #1469 where this feature was requested.


daviesrob:
Looks OK, thanks.

PR #1523, opened 2021-10-21
Cull the long deprecated VCF and BCF output from samtools mpileup.

jkbonfield:
This was marked as deprecated in 1.9, with both the usage statement and documentation components having been removed already.   The bcftools mpileup code has diverged considerably and should be used instead.

We still need bam2bcf.[ch] for the slimmest of reasons: to report ambiguity codes in tview consensus column.  However the bulk of the code has been trimmed to their minimum.  Frankly a basic summing algorithm is probably good enough for this modest use case, but for now it's left in-situ.

NB: the plan is to merge this as soon after the 1.14 release goes out, so third party developers get sufficient time to assess the impact.


bam_plcmd.c
```diff
@@ -1407,26 +1113,8 @@ int bam_mpileup(int argc, char *argv[])
...
-                            "'--open-prob INT' is functional, but deprecated. "
-                            "Please switch to using bcftools mpileup in future.\n");
-                } else {
-                    mplp.output_fname = optarg;
-                }
-            }
-            break;
-        case 'e': mplp.extQ = atoi(optarg); deprecated(c); break;
-        case 'h': mplp.tandemQ = atoi(optarg); deprecated(c); break;
+        case 'o': mplp.output_fname = optarg; break;
```
jmarshall, line 1116: There's a paragraph of “distinguishing -o” documentation to delete from _samtools-mpileup.1_ too (in the description of `--output`).

bam_plcmd.c
```diff
@@ -1338,7 +1054,7 @@ int bam_mpileup(int argc, char *argv[])
         {NULL, 0, NULL, 0}
     };
 
-    while ((c = getopt_long(argc, argv, "Agf:r:l:q:Q:uRC:BDSd:L:b:P:po:e:h:Im:F:EG:6OsVvxXt:aM",lopts,NULL)) >= 0) {
+    while ((c = getopt_long(argc, argv, "Agf:r:l:q:Q:uRC:Bd:b:o:EG:6OsVvxXaM",lopts,NULL)) >= 0) {
```
jmarshall, line 1057: Did you mean to leave `g…u…v` in here? They're gone from the switch cases, so they just fall through to “Unexpected global option. … Usage: …”.

I was guessing there might be going to be new `case` labels for these three, just printing out “These options have been removed. Blah blah…”.

jmarshall:
The documentation still mentions `--open-prob`, which no longer exists.

I think @jkbonfield was planning to address the review comments, but this had not occurred before this PR was merged.

jkbonfield:
Tbh I thought I'd dealt with the other comments too, but apparently not (although this is the first I've known about `--open-prob` being forgotten). Oops.  I'll make another PR to fix the few remaining glitches and review again the man page.  Thanks John

jmarshall:
> although this is the first I've known about `--open-prob` being forgotten

It's just a different (and more searchable :smile:) way of describing the previously mentioned “paragraph of _distinguishing -o_ documentation to delete” problem.

PR #1531, opened 2021-10-28
Consensus part1

jkbonfield:
This is the "samtools consensus" sub command.

The purpose is to provide an easy way to compute the consensus from a SAM (etc) file, potentially for use in embedded reference based compression, or as a target sequence to align against in an assembly improvement tool.

There is hidden functionality here which will make it proper into part 2 (and 3) at some point.  This includes a pileup format ("samtools consensus -f pileup"), but I expect this to change before advertising it.  Right now it's omitted from the help and usage statements.  The intention is to get a pileup-ish style output, but one that is easier to parse and in the spirit of the consensus command has one row per consensus column and not one row per reference locus (as mpileup gives).  Currently it doesn't do this part.  [That may also entail switching to io_lib's pileup algorithm as that's the native processing style.]

Future work also includes the ability to call modifications at the consensus level.  This functionality was present in the initial version of this PR, but has been removed pending further work on base modification support in htslib.


bam_consensus.c
```diff
@@ -0,0 +1,1426 @@
...
+    static int seqi2A[16] = { 0,8,0,4, 0,4,0,2, 0,4,0,2, 0,2,0,1 };
+    static int seqi2C[16] = { 0,0,8,4, 0,0,4,2, 0,0,4,2, 0,0,2,1 };
+    static int seqi2G[16] = { 0,0,0,0, 8,4,4,1, 0,0,0,0, 4,2,2,1 };
+    static int seqi2T[16] = { 0,0,0,0, 0,0,0,0, 8,4,4,2, 8,2,2,1 };
+
+    int ins_[128][16] = {0}, (*ins)[16] = ins_;
+    int i;
+
+    if (ins_len > 128)
+        if (!(ins = calloc(ins_len * 16, sizeof(int))))
```
daviesrob, line 185: This is quite a big multiplier.  On 32-bit platforms you could get an overflow with a carefully-crafted insertion.  It would be better to do `calloc(ins_len, 16 * sizeof(int))`.

bam_consensus.c
```diff
@@ -0,0 +1,1426 @@
...
+    if (y < -500)
+        y = -500;
+    if (y > 500)
+        y = 500;
+
+    return e_tab[(int)y];
+}
+
+/* Taylor (deg 3) implementation of the log:
+ * http://www.flipcode.com/cgi-bin/fcarticles.cgi?show=63828
```
daviesrob, line 417: This link is dead.  The new location looks like `https://www.flipcode.com/archives/Fast_log_Function.shtml`.

It's not clear what the copyright status of this example is, although given that this version works on doubles instead of floats it might not be a problem.

bam_consensus.c
```diff
@@ -0,0 +1,1426 @@
...
+static inline double fast_log2(double val)
+{
+   register int64_t *const     exp_ptr = ((int64_t*)&val);
+   register int64_t            x = *exp_ptr;
+   register const int      log_2 = ((x >> 52) & 2047) - 1024;
+   x &= ~(2047LL << 52);
+   x += 1023LL << 52;
+   *exp_ptr = x;
+
+   val = ((-1.0f/3) * val + 2) * val - 2.0f/3;
```
daviesrob, line 428: You probably don't need the `f`s here any more.

jkbonfield:
Replaced by #1557 which incorporates Io_lib's pileup algorithm for an easier to use interface.  This is strictly one column of data supplied per *consensus* position, rather than per *reference* position.  This makes handling indels different - easier, but possibly not as fully robust as the complex analysis I had before with a two stage approach to compute the most common indel length and then compute the consensus of only those indels.

However I'm not wanting a full caller, just a consensus algorithm that reflects the data it's given.  Fixing up variable sized indels is probably best done to a separate realigner stage prior to running this code, if we chose to go down that road.  (I have a basic one, so ought to test whether it even helps.)

PR #1537, opened 2021-11-02
Add `samtools head -e EXPR` header filtering option [DRAFT]

jmarshall:
A draft PR on top of #1517 that adds a `-e, --expr EXPR` option to the `samtools head` command, to allow arbitrary filtering of which headers to display.

Requires PR samtools/htslib#1351, and adds draft documentation of that PR's proposed header filter variables to _doc/samtools.1_.

This enables useful functionality such as:

```sh
# View only read groups
samtools head -e 'type == "RG"' foo.bam

# View only "interesting" chromosomes
samtools head -e '@SQ && [SN] !~ "^GL"' bar.bam
samtools head -e '@SQ && [LN:i] > 1000000' baz.bam
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ padding.o: padding.c config.h $(htslib_kstring_h) $(htslib_sam_h) $(htslib_faidx
 phase.o: phase.c config.h $(htslib_hts_h) $(htslib_sam_h) $(htslib_kstring_h) $(sam_opts_h) $(samtools_h) $(htslib_hts_os_h) $(htslib_kseq_h) $(htslib_khash_h) $(htslib_ksort_h)
 sam_opts.o: sam_opts.c config.h $(sam_opts_h)
 sam_utils.o: sam_utils.c config.h $(samtools_h)
-sam_view.o: sam_view.c config.h $(htslib_sam_h) $(htslib_faidx_h) $(htslib_khash_h) $(htslib_thread_pool_h) $(htslib_hts_expr_h) $(samtools_h) $(sam_opts_h) $(bam_h) $(bedidx_h)
+sam_view.o: sam_view.c config.h $(htslib_sam_h) $(htslib_faidx_h) $(htslib_khash_h) $(htslib_kstring_h) $(htslib_thread_pool_h) $(htslib_hts_expr_h) $(samtools_h) $(sam_opts_h) $(bam_h) $(bedidx_h)
 sample.o: sample.c config.h $(sample_h) $(htslib_khash_h)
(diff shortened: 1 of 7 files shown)


jkbonfield:
I think we never reviewed this and the associated htslib PR because they're labelled as draft.

Was this something you still had plans to continue John but never got time to continue, or was the draft tag something forgotten about.

For what it's worth, I think it's a nice idea and I like to reuse of the filtering language (although obviously I'm biased).  I'm not convinced the audience would be huge though as it's quite specialist and could be replaced with grep, awk or perl one-liners.  It also looks like it was written before "samtools head" was merged, so was this the first iteration which then later got dialed back to something simpler?

jmarshall:
I think perhaps I was planning to refactor the proposed HTSlib API in some way. Thanks for the reminder — this is a good point in the release cycle to revisit this.

jkbonfield:
I noticed this too, but I haven't investigated why yet.  It worked for me in my fork of this repository, but not when making a PR against samtools, which makes sense as it's the same owner.

It could well be trickier now we've moved from AppVeyor to GitHub Actions for windows, but the cirrus-ci stuff broke too so I'm expecting it's something to do with the CI helper script.

jkbonfield:
It looks like if you added `htslib#1351` into the first line of your commit message that it'd automatically pull out the named htslib PR and use that.  This was added in 078225abffe7d3bd8ecd45eeb9f1b619cab3513b

I assume at the time I couldn't figure out the environment variable needed to automatically use the author's comparably named htslib branch, but I'll make an dummy PR to experiment on.

jkbonfield:
Interesting that it's the title here rather than the actual commit message that works for Cirrus-CI!

Sadly this isn't available within GitHub Actions from what I can tell.  However the commit message is, with a bit of manual poking.  See #2115.

I have a PR which should be fixing this therefore (provided you modify the commit message rather than the github title).  If we can get that reviewed and merged then you could rebase on top of develop and it should start passing tests for windows too.   Sorry for the huge amount of spam from trial and error yesterday for anyone who got it!  Really debugging CI is the most painful thing :/

jmarshall:
[CIRRUS_CHANGE_TITLE](https://cirrus-ci.org/guide/writing-tasks/#environment-variables) is described as coming from the PR title in a PR event context (though it doesn't really document which contexts that is). So that was fun; happy to rework it again to handle the GH Action one in due course, or just not worry about the CI too much…

jkbonfield:
Yeah it's not an issue really, just a convenience.  We'd need to merge htslib one first anyway in which case we can just retrigger the test to get this checked.  (Hence why I've been ignoring this known issue for so long)

PR #1538, opened 2021-11-03
Added fix for odd sequence length hardclip bug.

### LINKED ISSUE #1527: ampliconclip bug when hardclipping odd-length sequences from the left
When hardclipping on the left, should not instead of the line 390 in bam_ampliconclip.c, https://github.com/samtools/samtools/blob/develop/bam_ampliconclip.c#L390, there be line 392 saying:

`odd_base = rec->core.l_qseq  & 1;`

otherwise when doing the memcopy in the line 396, https://github.com/samtools/samtools/blob/develop/bam_ampliconclip.c#L396, when the sequence length is odd, the last letter in the sequence gets left out.

This fixes the  hardclippinng, not sure if there are any side effects for the softclipping.

whitwham:
This should fix #1527.


daviesrob:
Looks good, thanks.

Issue #1497, opened 2021-09-07
tview: Catalog of loci


lskatz:
#### Is your feature request related to a problem? Please specify.

When viewing in tview, you can use `?` for a nice help menu including navigation and including a helpful menu.
However, there is no listing the seqnames or their lengths.

#### Describe the solution you would like.

Is there already a way in tview to show the seqnames and their lengths?
If not, I would like to request this sort of menu, maybe when `g` is pressed.
Or a new key like `G` to list the seqnames and then navigate to highlight, and then enter to go to that seq.

Thank you!

PR #1539, opened 2021-11-04
Tview ref list

absanger:
Improves on #1530 and hopefully provides a better solution to #1497.

Here's what it should look like:
refs


whitwham (changes requested):
Looks good and a definite improvement on what was there before.  As I am the samtools copyright police, could you update the copyright notice in bam_tview_curses.c please.

whitwham (approved):
Thanks you.

PR #1540, opened 2021-11-04
Fixes the documentation for mpileup --output-BP-5.

### LINKED ISSUE #1534: mpileup: --output-BP-5 switches output column 7 and 8
when using the --output-BP-5 option, the position and read id columns are switched
```
samtools mpileup --output-QNAME --output-BP in.bam
test_chr	25	A	1	^!,	I	1	id_1

```
```
samtools mpileup --output-QNAME --output-BP-5 in.bam
test_chr	25	A	1	^!,	I	id_1	24
```

jkbonfield:
The original intention of this was to modify the orientation of the `--output-BP` field, as originally requested.  Due to lack of research into the original code, sadly this was moved to the end of the list of columns.  We didn't test in conjunction with everything else turned on.

It was decided that we should not fix this as it would become incompatible with the previous release, so it's now been documented.

However given it's now its own column instead of modifying the existing column, `--output-BP-5` is no longer mutually exclusive with `--output-BP`.

Fixes #1534 (albeit not in the ideal way)

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -1100,14 +1100,8 @@ int bam_mpileup(int argc, char *argv[])
         case '6': mplp.flag |= MPLP_ILLUMINA13; break;
         case 'R': mplp.flag |= MPLP_IGNORE_RG; break;
         case 's': mplp.flag |= MPLP_PRINT_MAPQ_CHAR; break;
-        case 'O':
-            if (!(mplp.flag & MPLP_PRINT_QPOS5))
-                mplp.flag |= MPLP_PRINT_QPOS;
-            break;
-        case  14:
-            mplp.flag |=  MPLP_PRINT_QPOS5;
-            mplp.flag &= ~MPLP_PRINT_QPOS;
-            break;
+        case 'O': mplp.flag |= MPLP_PRINT_QPOS; break;
+        case  14: mplp.flag |= MPLP_PRINT_QPOS5; break;
         case 'M': mplp.flag |= MPLP_PRINT_MODS; break;
         case 'C': mplp.capQ_thres = atoi(optarg); break;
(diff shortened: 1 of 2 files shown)


jmarshall:
Both documentation paragraphs need a “(Column only present when *…option…* is used.)” sentence, and looking at the code suggests that in 1.14 the `--output-BP-5` column actually appears **before** the “Additional read tag field columns”. (So in between the two halves of `--output-extra` — brillant! :smile:)

(#1281 is an example of where `samtools-1.10 mpileup` column ordering was altered in 1.11 to agree with its documentation — except in that case there was the advantage that it was also being altered to agree with the ≤1.9 behaviour.)

jkbonfield:
Hmm, I'll check again to make sure I've got the documentation right then.

My personal preference, given it's recent and somewhat obscure, would be to fix as intended, but I think we're fearful of the mess that could ensue with it being in different positions.  I have sympathy for that if I believe it was in common usage, but I expect it really isn't!

jkbonfield:
Updated the ordering again.

The code is pretty cryptic and it's not immediately obvious that the bit-field also represents a partial ordering too.  (I say partial as some fields still appear out of order.)

bam_plcmd.c
```diff
@@ -1106,7 +1106,6 @@ int bam_mpileup(int argc, char *argv[])
             break;
         case  14:
             mplp.flag |=  MPLP_PRINT_QPOS5;
-            mplp.flag &= ~MPLP_PRINT_QPOS;
```
daviesrob, line 1109: You also need to remove the test on `MPLP_PRINT_QPOS5` on line [1104](https://github.com/samtools/samtools/blob/e56146ca/bam_plcmd.c#L1104)

PR #1542, opened 2021-11-09
Add new `view --fetch-pairs` option

pd3:
Proof of principle implementation that allows to pull reads from
regions including their mates, even when they fall outside the regions
or are unmapped.

Note
- so far this was tested on small data only
- tests need to be added
- everything is stored in memory, with many regions an external
  temporary storage may be needed
- some unmapped reads may be missed, this depends on the success of
  https://github.com/samtools/htslib/pull/1352 pull request


sam_view.c
```diff
@@ -404,47 +413,243 @@ int parse_aux_list(auxhash_t *h, char *optarg) {
...
+    while ( sam_itr_multi_next(conf->in, iter, rec)>=0 )
+    {
+        int drop = 1;
+        if ( rec->core.tid >=0 && bed_overlap(conf->bed, sam_hdr_tid2name(conf->header,rec->core.tid), rec->core.pos, bam_endpos(rec)) ) drop = 0;
+        if ( drop && khash_str2int_get(mate_names,bam_get_qname(rec),&ret)==0 ) drop = 0;
+        if ( !drop && check_sam_write1(conf->out,conf->header,rec,conf->fn_out,&ret) < 0 ) error("aborting");
+    }
+
+    int i;
+    for (i=0; i<nreg; i++) free((char*)reglist[i].reg);
```
daviesrob, line 599: This can be dropped if we can get rid of all the `strdup()` calls when setting the `reg` fields.

sam_view.c
```diff
@@ -404,47 +413,243 @@ int parse_aux_list(auxhash_t *h, char *optarg) {
...
+                change_flag(b, conf);
+                if (check_sam_write1(conf->out, conf->header, b, conf->fn_out, &ret) < 0) break;
+            }
+            conf->count++;
+        } else if (conf->unmap) {
+            b->core.flag |= BAM_FUNMAP;
+            if (check_sam_write1(conf->out, conf->header, b, conf->fn_out, &ret) < 0) break;
+        } else {
+            if (conf->un_out) { if (check_sam_write1(conf->un_out, conf->header, b, conf->fn_un_out, &ret) < 0) break; }
+        }
```
daviesrob, line 625: I think this part is also common to the old-iterator and non-iterator versions, so it would be good to pull it out into its own inline function.

sam_view.c
```diff
@@ -855,188 +1068,137 @@ int main_samview(int argc, char *argv[])
...
-                    if (!is_count) {
+    }
+    else    // retrieve alignments in specified regions using the generic iterator
+    {
+        int i;
+        bam1_t *b;
+        b = bam_init1();
+
+        for (i = (has_index_file)? optind+2 : optind+1; i < argc; ++i) {
+            int result;
```
daviesrob, line 1140: I think it should be possible to get rid of most of this version by making it call `multi_region_init()` followed by `multi_region_view()` for each region separately.  If it works it would be a very useful simplification.

daviesrob (commented):
It looks like the fetch-pairs option currently disables all other filters.  We might want to consider how they might work together, and what corner cases that may introduce.

Of course this needs tests, documentation etc.

pd3:
> It looks like the fetch-pairs option currently disables all other filters. We might want to consider how they might work together, and what corner cases that may introduce.

Allowing filters is problematic, one of the reads from the pair can fail and the other pass, which goes against the spirit and the intent of the feature. One can pipe one view command into another if filtering is required.

jmarshall:
If the feature is separate from all the other view filtering options, perhaps it should be a separate command.

daviesrob:
I think it could be compatible with filters if we say that if a read passes the filtering then every read with the same name will be returned, even if the filter would normally have rejected it.  I think the only thing we'd need to be careful about would be accidentally including other reads from the regions where we're getting pairs - so we would need to check against the original region specification and only allow the specific pairs we're after if they're outside it.

sam_view.c
```diff
@@ -404,47 +413,243 @@ int parse_aux_list(auxhash_t *h, char *optarg) {
...
+{
+    int nreg = *_nreg;
+    int i = _reglist_find_tid(reg,nreg,tid);
+    if ( i>=nreg || reg[i].tid!=tid )
+    {
+        nreg++;
+        reg = (hts_reglist_t*)realloc(reg,sizeof(hts_reglist_t)*nreg);
+        if ( i+1 < nreg )
+        if ( i+1 < nreg )
+            memmove(reg + i + 1, reg + i, sizeof(hts_reglist_t)*(nreg - i - 1));
```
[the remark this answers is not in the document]
pd3: Then it would not be possible to perform a binary search to identify the chromosome block. I was thinking of the typical case of finished assemblies.

pd3:
The pull request has been updated and includes tests and documentation.

daviesrob:
I've rebased this, along with some small adjustments and a bit more refactoring.  Now three copies of the process-and-write code have been merged into one, which is a big improvement.

What we need to sort out now is how exactly `--fetch-pairs` should interact with other options.  Currently there may be inconsistencies because filters apply to all reads from the original regions, but not to pairs that get pulled in, which could lead to inconsistencies in the output data.  For example, if you used the `--remove-tag` option, records from the original region would have tags removed, but pairs would not.

I think there are two ways that this could work:
1. The `--fetch-pairs` option first expands the region list to get the extra pairs, then `process_aln()` is applied to both the original set and the paired reads.  That's not quite what happens at the moment.
2. `--fetch-pairs` uses `process_aln()` to select which pairs to return, and then you get both of them if either one of them passed the filter.

Option 2 might be a bit more consistent, but has the disadvantage that you have to store all of the read names, so you can't get away with dropping the ones that hit the original region of interest.  So it might be better to go with option 1, which may be possible by adjusting when `process_aln()` gets called the second time through.

One other thing I noticed when testing is that while secondary/supplementary reads from outside the original region are not usually included, it's possible for them to appear if they happen to be somewhere that gets added to the region list due to another paired read at the same location.  This can happen with either of the options above, but it's much more likely with 2 due to the number of reads in the hash table.  Should we make the `--fetch-pairs` option always turn on secondary and supplementary read filtering, as I think they're going to be difficult to deal with otherwise?

sam_view.c
```diff
@@ -405,47 +414,292 @@ int parse_aux_list(auxhash_t *h, char *optarg) {
     return 0;
 }
 
+#define error(...) { \
+   print_error_errno("view",__VA_ARGS__); \
+   exit(EXIT_FAILURE); \
```
jmarshall, line 419: Unlike bcftools, it is very uncommon in samtools to call `exit` to abort subcommands. This makes a difference for wrappers such as pysam, so it would be good to follow the existing samtools convention: use `print_error` or `print_error_errno` directly and return an error indication from subroutines etc, and this `error` macro will be unnecessary.

pd3: Can you provide more detail about what the wrappers expect? It makes a big difference for the code structure. The `view` command is written in a way that is very hard to refactor and this was an easy solution to that.

pd3:
I see there are two operations, filtering and editing hidden in `process_aln`. The original code considered filtering but not the editing part.

The option 1, as far as I understand it, is not good in that it may result in single reads on the output, which defeats the purpose.
The option 2 requires storing the names, also not good.
Best might be option 3 which refactors `process_aln` into two separate functions, one for filtering and one for editing. The behavior then will be as described in the man page ("filters are applied only on reads overlapping the requested regions but not their mates"). 

Filtering of secondary and supplementary reads automatically when `--fetch-pairs` is requested is a good idea.

sam_view.c
daviesrob: pysam has a habit of [calling samtools' `main`](https://github.com/pysam-developers/pysam/blob/50f784ed/samtools/samtools.pysam.c#L66) from python.  This of course causes problems if samtools calls `exit` as you get dumped out of the python interpreter.  Arguably it would be safer to call `fork` and run samtools in a separate process, but I guess they may have their reasons for using the method that they do.

The `error()` macro is only used in four functions, so it wouldn't be too difficult to adjust them so that they clean up any memory used and return error codes instead.

daviesrob:
The editing part is currently just a bit wasteful, as you throw away the results of the first pass through.  It wouldn't be too difficult to put the tag adjustments in their own function so they only get called when needed.  The `bam_remove_B()` bit would need to stay, as it can affect later tests.

Yes, option 1 could give you single reads, but if that was asked for in the filters, should we argue?  Note that currently you can also get single reads, if they both overlap the originally-defined region.

sam_view.c
jmarshall: Pysam rewrites `exit` via `longjmp` so you are no longer dumped out of the interpreter, but `exit` does then cause leaks.

The usual convention in samtools is to return from `main` rather than call `exit`.

daviesrob: I've added a commit that changes the error handling so it doesn't call `exit()`, which should make pysam happier.

While going through this I also spotted and fixed a few places where error checking had been missed, so it was a useful exercise in that regard.  The most important of these was checking the return value from the iterator, so that it's possible to detect read errors.

daviesrob:
After a short diversion around aux tag removal, I've also adjusted the `--fetch-pairs` option to apply filters to all of the reads.

I've made quite a few changes to this now.  It would be good if someone else cast an eye over it to make sure it's OK.

doc/samtools-view.1
```diff
@@ -442,6 +446,23 @@ is equivalent to
...
 .TP
+.BR -P ", " --fetch-pairs
+Retrieve pairs even when the mate read is outside of the requested region.
+Enabling this option also turns on the multi-region iterator (\fB-M\fR).
+A region to search must be specified, either on the command-line, or using
+the \fB-L\fR option.
+The input file must be an indexed regular file.
+
+This option first scans the requested region to look for mate pairs outside
+the original search area.
```
pd3, line 457: Perhaps it should be noted that the mate pairs look up is done only for reads passing filters if any were given.

pd3 (commented):
The changes looks good to me

d-cameron:
> Should we make the --fetch-pairs option always turn on secondary and supplementary read filtering, as I think they're going to be difficult to deal with otherwise?

IMO, `--fetch-pairs` on a region should work the same way as grabbing all the read (names) in a region then pulling out them and their mates with `samtools view -N` (this is what I do in VIRUSBreakend and why I submitted the `-N` PR). Secondary and supplementary are indeed difficult to deal with and that's why my extract pass reads the entire file.

I don't think any option should silently set filters. A better approach would be to check the filters and do either an indexed lookup or a linear scan depending on whether the selected filters are compatible with indexed lookup.

Always filtering secondary/supplementary has behaviour that is likely to break legitimate use cases. For example, a split read spanning a breakpoint should have all 3 records (2*soft clipped + mate) extracted regardless of which side of of the breakpoint was in the region filter.


Essentially, my concern is about whether `--fetch-pairs` is really `--fetch-fragment` or `--fetch-mate`. If it's the latter then you can say that you only extract the mate alignment. If it's the former, then you really do need to extract everything and take the performance hit to do so (when necessary).

d-cameron:
Some edge cases I've had to deal with:

1) GATK indel realignment invalidating PNEXT/MC
```
mate_position_lies	65	chrA	1	0	10M	chrA	1	0	*	*
mate_position_lies	129	chrB	1000	0	10M	chrA	1	0	*	*
```

2) spec-compliant secondary alignment of read
```
secondary_read_alignment	65	chrA	1	0	10M	chrB	1000	0	*	*
secondary_read_alignment	129	chrB	1000	0	10M	chrA	1	0	*	*
secondary_read_alignment	385	chrC	1000	0	10M	chrA	1	0	*	*
```

3) spec-compliant secondary alignment of fragment
```
secondary_fragment_alignment	65	chrA	1	0	10M	chrB	2	0	*	*
secondary_fragment_alignment	129	chrB	2	0	10M	chrA	1	0	*	*
secondary_fragment_alignment	321	chrD	3	0	10M	chrC	4	0	*	*
secondary_fragment_alignment	385	chrC	4	0	10M	chrD	3	0	*	*
```

4) bwa reports secondary alt alignments as supplementary
```
bwa_secondary_as_supp	65	chrA	1	0	10M	chrUn	3	0	*	*
bwa_secondary_as_supp	129	chrB	2	0	10M	chrA	1	0	*	*
bwa_secondary_as_supp	2177	chrUn	3	0	10M	chrA	1	0	*	*
```

5) split read alignments
```
supp1	65	chrA	1	0	10M	chrB	2	0	*	*
supp1	129	chrB	2	0	5M5S	chrA	1	0	*	*	SA:Z:chrC,3,+,5S5M,0,0
supp1	2177	chrC	3	0	5S5M	chrA	1	0	*	*	SA:Z:chrB,2,+,5M5S,0,0
```

5) multi split reads
```
supp2	65	chrA	1	0	10M	chrB	2	0	*	*
supp2	129	chrB	2	0	5M10S	chrA	1	0	*	*	SA:Z:chrC,3,+,10S5M,0,0;chrD,4,+5S5M5S,0,0
supp2	2177	chrC	3	0	10S5M	chrA	1	0	*	*	SA:Z:chrB,2,+,5M10S,0,0;chrD,4,+5S5M5S,0,0
supp2	2177	chrD	4	0	5S5M5S	chrA	1	0	*	*	SA:Z:chrB,2,+,5M10S,0,0;chrC,3,+,10S5M,0,0
```

doc/samtools-view.1
d-cameron: Can we use different terminology than "mate pairs"? It's unclear as to whether secondary/supplementary alignment should or should not be included and [mate pair sequencing](https://sapac.illumina.com/science/technology/next-generation-sequencing/mate-pair-sequencing.html) isn't what this PR is referring to.

d-cameron:
You can pretty much reuse the `-N` codepath for the slow way. Once you've got the relevant read names for the first pass region lookup, it's just plumbing the filters into a samtools view call and pass through all filters except the region ones and `-N`.

d-cameron:
And if we really want to a spec-compliant `--fetch-fragment` implementation, then we'd need to consult `CC`/`CP`/`FI`/`FS`/`HI`/`IH` and not just `RNEXT`/`PNEXT`/`SA`. AFAIK, these aren't actually used by anyone and custom UMI barcode fields are used instead (e.g. 10X Genomics Chromium).

pd3:
> IMO, `--fetch-pairs` on a region should work the same way as grabbing all the read (names) in a region then pulling out them and their mates with `samtools view -N` (this is what I do in VIRUSBreakend and why I submitted the `-N` PR). Secondary and supplementary are indeed difficult to deal with and that's why my extract pass reads the entire file.

No, it shouldn't. The purpose of the option is different: retrieve complete pairs so that one can realign the region. The fuctionality you are describing would be slow and, as you note, it can be already achieved with the help  of `-N,--qname-file` option. Making `--fetch-pairs` operate the same way would be redundant.

d-cameron:
Can we call it `--fetch-mates` or `--include-mates` (my preference) then?

We could also update the documentation to explicitly state which records we are retrieving:

"For the purposes of this option, a read pair mate is the primary alignment of the other
read in the read pair. That is, the record and mate must have the 0x1 flag set, different 0x2 and 0x4
flags, the mate 0x100 and 0x800 flags not set, and the mate alignment
overlaps the position specified in the record RNEXT/PNEXT fields.

daviesrob:
I've rebased again (sorry) so that this PR gets the AppVeyor tests fix.  I've also added another commit to better describe how the option works in the man page.

At the moment, it doesn't use any flags when selecting extra reads, just `QNAME` matches.  I think this is alright, as checking flags for the additional reads might be complicated.  I think there is a case to be made for checking that the 0x01 bit is set, though, as we shouldn't really be using `PNEXT` and `RNEXT` for single-ended reads.  It wouldn't be too hard to add a check for that.

daviesrob:
Another small adjustment to check the `BAM_FPAIRED` flags bit.  If this is unset, there shouldn't be a paired read and following `RNEXT` and `PNEXT` isn't likely to lead anywhere useful.

daviesrob:
Now squashed to a slightly tidier set of commits.

PR #1543, opened 2021-11-10
Minor fixes to cullage of mpileup vcf/bcf output.

jkbonfield:
- The -g, -u and -v options are no longer in the getopt_long option  list.

- The man page no longer mentions the two uses of -o (`--output` vs `--open_prob`).

Fixes comments originally raised in #1523.

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -1054,7 +1054,7 @@ int bam_mpileup(int argc, char *argv[])
         {NULL, 0, NULL, 0}
     };
 
-    while ((c = getopt_long(argc, argv, "Agf:r:l:q:Q:uRC:Bd:b:o:EG:6OsVvxXaM",lopts,NULL)) >= 0) {
+    while ((c = getopt_long(argc, argv, "Af:r:l:q:Q:RC:Bd:b:o:EG:6OsxXaM",lopts,NULL)) >= 0) {
         switch (c) {
         case 'x': mplp.flag &= ~MPLP_SMART_OVERLAPS; break;
         case  1 :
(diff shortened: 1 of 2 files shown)


bam_plcmd.c
```diff
@@ -1054,7 +1054,7 @@ int bam_mpileup(int argc, char *argv[])
         {NULL, 0, NULL, 0}
     };
 
-    while ((c = getopt_long(argc, argv, "Agf:r:l:q:Q:uRC:Bd:b:o:EG:6OsVvxXaM",lopts,NULL)) >= 0) {
+    while ((c = getopt_long(argc, argv, "Af:r:l:q:Q:RC:Bd:b:o:EG:6OsVxXaM",lopts,NULL)) >= 0) {
```
whitwham, line 1057: What is the **V** for?

jkbonfield: Oops, it's another one unused too.  Good spot.  It used to be this:

```
        case 'V': mplp.fmt_flag |= B2B_FMT_DV; fprintf(stderr, "[warning] samtools mpileup option `-V` is functional, but deprecated. Please switch to `-t DV` in future.\n"); break;
```

I though when I was removing the case statements I'd also removed all from the getopts list, but I guess some more crept through.  I should probably double check the entire thing again then.

jkbonfield: Found a bunch more in the long opts too.

I've gone through long options, short options and case statements and they're all in sync again.  Plus all differences can be related back to things I failed to do originally, so differences have been explained.

jkbonfield: I've double checked long opts, short opts and case statements (this time on the correct branch!).  The 'V' is the only rogue one I could spot.  Squashed it in.

Thanks.

PR #1553, opened 2021-11-30
Fixes a bug detected by clang 13 with -Wformat-security.

jkbonfield:
The mvwprintw message needs to be an argument to a "%s" format, to
avoid the str itself containing % rules.  In particular a reference
containing %n will cause it to write to illegal (or worse, legal)
addresses.

### DIFF
--- a/bam_tview_curses.c
+++ b/bam_tview_curses.c
@@ -208,9 +208,8 @@ static void tv_win_listref(curses_tview_t *tv, int* matches, int num_matches,
             mvwprintw(tv->wlistref, i+1, 1, ">");
         }
         snprintf(str, TV_MAX_GOTO, "%s  length: %d", ref, len);
-        mvwprintw(tv->wlistref, i+1, 2, str);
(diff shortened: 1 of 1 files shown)


jmarshall:
Or could reduce it to `mvwaddstr()` instead, though I see _bam_tview_curses.c_ doesn't currently use that…

jkbonfield:
I thought about folding the snprintf command into the mvwprintw call, but that then makes the restriction of size a bit tricky to do as we wouldn't know what `%.*s` value to use unless we can get a handle on the `%d` bit too.  None of it is ideal as I'm not even sure it works properly once you get references that no longer fit in the dialogue box, but I simply took the smallest change as a quick fix to unbreak the tests.

I hadn't hunted for alternative curses commands though.  I did check elsewhere and sure the "%s" meme was used throughout, so just went with it.  It's in now, but agreed mvwaddstr is probably more suitable, assuimng it's portable.

Issue #1544, opened 2021-11-11
B aux tag not supported by mpileup


jkbonfield:
Mpileup has an option of `--output-extra` which permits writing of aux tags fields.

If given a tag which is a B array, this field is not output, so an empty field is present ("", not "\*").  If they're not supported they should be replaced with "\*" perhaps (incase that aux tag does exist in another record with another type, to ensure the number of columns is fixed).

I also note that bases that are below the `--min-BQ` figure, and perhaps other filters, get "*" for all these additional output columns.  That's clearly deliberate, but as far as I can see undocumented.

An example file:

```
@HD	VN:1.4	SO:coordinate
@SQ	SN:1	LN:249250621
@RG	ID:rg
r1	99	1	10001	2	10M	*	0	0	ACCGGGTTTT	AA>%=#F?I=	NM:i:1	MD:Z:18T81	RG:Z:rg	FF:f:1.23e21	BB:B:i,-10,10,20,-100
```

With command:

```
$ samtools mpileup _.sam --output-extra RG,BB,RG --output-BP --output-QNAME
[mpileup] 1 samples in 1 input files
1	10001	N	1	^#A	A	1	r1	rg		rg
1	10002	N	1	C	A	2	r1	rg		rg
1	10003	N	1	C	>	3	r1	rg		rg
1	10004	N	0	*	*	*	*	*	*	*
1	10005	N	1	G	=	5	r1	rg		rg
1	10006	N	0	*	*	*	*	*	*	*
1	10007	N	1	T	F	7	r1	rg		rg
1	10008	N	1	T	?	8	r1	rg		rg
1	10009	N	1	T	I	9	r1	rg		rg
1	10010	N	1	T$	=	10	r1	rg		rg
```

Edit: verified i, f, Z, H and A tag types all work.

PR #1554, opened 2021-12-02
Fix empty field on unsupported aux tags.

absanger:
Fixes: #1544 

Replaces the empty fields on unsupported aux tags with a '*'.


jkbonfield:
Looks like a nice minimal fix.  Thanks.

PR #1557, opened 2021-12-07
Add a "samtools consensus" sub-command.

jkbonfield:
This support simple frequency counting as well as a more complex
Bayesian algorithm used by Gap5 and Crumble.

We also permit use of the mapping quality, but this is too errorprone
to take at face value, so it may be capped at low/high values or
adjusted via other means such as a localised NM figure.

The main proposal is for fasta / fastq, but there is also a simplified
pileup-ish format output.  This contains the consensus column (like
old pileup, rather than mpileup).  It's also strictly one row per
consensus position rather than one per reference coordinate (by
default).

The possible use cases for this command include:

- Generating a consensus suitable for draft assembly refinement,
  similar to ICORN, where the reads are aligned to the assembly, we
  generate a revised consensus, and then realign back to that
  consensus again.

- Generation of a consensus suitable for embedding within a file for
  purposes of storing reference-based CRAM files, without needing to
  submit a reference.  This may be suitable for more adhoc work where
  we don't yet have a final assembly we wish to submit, but still wish
  to provide aligned data.  Or alternatively where we have a very
  crude assembly produced solely for the purposes of data compression
  and the contigs are not intended to be a usable product.

- Simple consensus generation for things like pandemic tracking; eg
  as a replacement of ivar consensus.  Typically data here is so deep
  that it's easy to achieve accuracy via simple methods.

It also turns out however that it's possible to use it as a basic
rapid diploid variant caller.  (It's approx 8x faster than bcftools
for this.)  This commit does not include a way to generate VCF output,
but I do have a hacky perl script to convert from pileup to VCF.  If
this is deemed useful it could become a basic option in the tool, but
I don't think it's really the intention.  Generation of VCF was purely
as a test for evaluating the robustness of the consensus, which only
needs to be adequate for the use cases above.

If we wished to persue that evenue, it would be best as part of a
pipeline of tools, such as some simple mapq / baseq recalibration and
a local realignment step.  These are the sorts of things internally
that the top quality callers need to do to ensure accuracy.

For the curious however, here is an example for SynDip chr1, after
conversion of pileup to a noddy VCF.  SNP only as I haven't converted
indels (plus I'd expect them to be much less robust).

```
gatk 4:
SNP          Q>0 /  Q>=150
SNP   TP  265050 /  263351
SNP   FP    6819 /    3991
SNP   GT     891 /     785
SNP   FN    4605 /    6304

bcftools (2021?):
SNP          Q>0 /   Q>=46
SNP   TP  265363 /  263345
SNP   FP    5666 /    3453
SNP   GT     676 /     505
SNP   FN    4292 /    6310

freebayes (2018):
SNP          Q>0 /   Q>=12
SNP   TP  264313 /  263355
SNP   FP    6018 /    5468
SNP   GT    1673 /    1637
SNP   FN    5342 /    6300

Samtools consensus
SNP          Q>0 /   Q>=51
SNP   TP  265792 /  263377
SNP   FP    5393 /    3155
SNP   GT     686 /     342
SNP   FN    3863 /    6278
```

Q>=x is with bcftools filter of QUAL>=x.  It differs per tool as I
tried to normalise for the same FN figure of approx 6300, so we can
then judge how TP, FP and GT errors change.  GT is genotype error,
where the call is correct, but mislabelled as het / hom.

The above was 45x.  At 15x it's quite a bit poorer than other tools.
Similarly GIAB is a more mixed bag.  Ie this works sometimes, but it's
a not a fully robust caller.  So use with caution!


bam_consensus.c
```diff
@@ -0,0 +1,1685 @@
...
+    err |= kputc_('\t', ks) < 0;
+    err |= kputc_(cb, ks)   < 0;
+    err |= kputc_('\t', ks) < 0;
+    err |= kputw(cq, ks)    < 0;
+    err |= kputc_('\t', ks) < 0;
+    if (err)
+        return -1;
+
+    /* Seq + qual at predetermined offsets */
+    ks_resize(ks, ks->l + depth*2 + 2);
```
daviesrob, line 1223: The return value needs to be checked here.

consensus_pileup.c
```diff
@@ -0,0 +1,606 @@
...
+                /* Another indicator for unmapped */
+                continue;
+            } else if (b->core.tid == last_ref) {
+                pos = b->core.pos+1;
+                //printf("New seq at pos %d @ %d %s\n", pos, b->core.tid,
+                //       bam_name(b));
+                last_in_contig = 0;
+            } else {
+                //printf("New ctg at pos %d @ %d\n",b->core.pos+1,b->core.tid);
+                pos = (b->core.pos > col ? b->core.pos : col)+1;
```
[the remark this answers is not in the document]
jkbonfield: A reasonable comment on `last_in_contig`.  It's set when we change contigs, so isn't so much that _this_ read is the last but that the previous one was, so a change of name is OK.

As for `col+1`, oddly it makes no difference as when `last_in_contig` is set we have `pos++` later on for this case and both col and pos march up until we hit the end of the contig.   Hence any value of pos is sufficient providing it's above col.  The simpler definition you suggest is therefore the obvious one.  (I'm struggling to remember precisely why that increment is there, but it's needed.  That probably means I should have commented that bit!)

`HTS_POS_MAX` is a bad idea though.  Yes it does work, but only because it's defined to be 0x7fffffff80000000 rather than 0x7fffffffffffffff, which luckily avoids the overflow when we increment pos beyond `HTS_POS_MAX`.

jkbonfield:
One thing I forgot to bring up at our weekly meeting was the default behaviour and option names.

When I was writing this, "gap5 consensus" was a useful placeholder and it makes sense to me given the origin of the algorithm, however I don't thin it's a useful term.  By all means document the origin in the man page, but it'd be better described as bayesian consensus or similar.  Secondly my instinct is that it should also be the default behaviour as it's the most accurate consensus, with "simple mode" (counting bases and/or just summing scores) being an option that can be selected.

Thoughts?  I'm happy to make these changes.

consensus_pileup.c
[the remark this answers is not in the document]
daviesrob: If you used `HTS_POS_MAX` then you could get rid of the line where it increments `pos` as it will already be bigger than any sensible value for `col`.  So I don't think `pos` could overflow.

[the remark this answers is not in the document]
jkbonfield: I did try this and it fails, but I've now figured out why.  (There's more than one place it needed changing in.)  This is indeed a better solution.  Thanks

daviesrob:
To avoid a proliferation of options in the future, might it be better to have something like `--mode` to set the consensus algorithm?  That way we wouldn't need to introduce more letters if we ever come up with a new consensus generationm method.

And irrespective of the above, I think it ould be good to have an option that explicitly selects the Bayesian algorithm.  That way it will be possible to select it, even if we decide to change the default.

jkbonfield:
I've added a `-m` / `--mode` option that accepts _simple_ and _gap5_ as options.  This does mean we need to actually name the default mode now, which brings back the question of what to call it.  I know you weren't happy with "bayesian" and I think calling it the "gap5" mode while accurate probably won't mean much to most people (sadly!).  Any suggestions?

daviesrob:
I'm not that worried about calling it "bayesian".

bam_consensus.c
```diff
@@ -0,0 +1,1702 @@
...
+
+            S[12] += MM;
+            S[13] += _M;
+            break;
+        }
+
+        depth++;
+
+        if (p->eof && p->cd)
+            free(p->cd);
```
daviesrob, line 803: It might be a good idea to set this to NULL after free.

jkbonfield: It's harmless, as it's initialised to zero when they're reused (https://github.com/jkbonfield/samtools/blob/consensus_part1b/consensus_pileup.c#L494).  However it's not really detrimental so can add it.

Arguably a better interface would be to not free it and cache the length it was allocate to, so it doesn't need reallocating, but that's quite a bit more to do and it would need a tidy up function at the end too.

jkbonfield:
Added the `p->cd = NULL` (and a const removal) as suggested.  Also changed "gap5" to "bayesian" based on comments in the weekly meeting.  Internally all the functions and variables are still the same, so this is just a TUI and man page change.

As these changes were minimal I also squashed and rebased so hopefully now ready for merging.

bam_consensus.c
```diff
@@ -1472,6 +1474,7 @@ int main_consensus(int argc, char **argv) {
         {"adj-MQ",             no_argument,       NULL, 'm'+100},
         {"no-adj-MQ",          no_argument,       NULL, 'm'+101},
         {"NM-halo",            required_argument, NULL, 'h'+100},
+        {"SC-cost",            required_argument, NULL, 'h'+101},
```
daviesrob, line 1477: I don't think there are any at the moment, but this way of selecting the value returned by long opts risks collisions...

bam_consensus.c
```diff
@@ -0,0 +1,1713 @@
...
+
+/* Taylor (deg 3) implementation of the log */
+static inline double fast_log2(double val)
+{
+    // FP representation is exponent & mantissa, where
+    // value = 2^E * M.
+    // Hence log2(value) = log2(2^E * M)
+    //                   = log2(2^E)+ log2(M)
+    //                   =        E + log2(M)
+    int64_t *const exponent = ((int64_t*)&val);
```
daviesrob, line 357: After exploring for probably far too long, it looks like modern compilers have an exception for this sort of type punning.  An ancient version of gcc did [compile it incorrectly](https://godbolt.org/z/q8WKoqjxM) though, so it may be safer to use a union instead.  Using a union is also [recommended by gcc](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-fstrict-aliasing), although it is a [bit controversial](https://davmac.wordpress.com/2010/02/26/c99-revisited/).

bam_consensus.c
jkbonfield: Ironically I started doing this when I did accidentally produce a clash when just using numerics.

I like to keep the options grouped in the list by function, but the numbers were being added sequentially as I added and extended the algorithm, which means you have to scan around to look for other numbers.

Note though that the compiler spots such errors (duplicate case statements) so it's very low risk

bam_consensus.c
jkbonfield: I'll have a play.

jkbonfield: Union looks solid.  I couldn't work out whether that gcc assembly is incorrect, as I can't find a system to test it on (short of a full download and compile from source).

I did verify though that icc, gcc7, gcc8 and gcc9 all produce correct code either way, but it's marginally *slower* with the non-union implementation.  So that's another reason to switch.

daviesrob:
Rebased with minor squashing to get the tests working again.

PR #1558, opened 2021-12-08
Optical Duplicate Marking: Add regex options for custom coordinates.

whitwham:
For the case of non standard read names (QNAME), add options to read the coordinates and, optionally, another part of the string to test for optical duplication.

Options are:

--read-coords REGEX
This takes a POSIX regular expression for at least x and y. but can include another part of the read name to test for equality, eg lane:tile elements.

--coords-order ORDER
The order of the elements captured in the regular expression.  Default is txy where t is a part of the read name selected for string comparison and x/y the coordinates used for optical duplicate detection.

An example to match the way read name reading currently works would be:
--read-coords '(\w+:[0-9]+:[0-9]+:[0-9]+:[0-9]+):([0-9]+):([0-9]+)'
--coords-order txy
To match the read format `machine:run:flowcell:lane:tile:x:y`

An alternative might be:
--read-coords '^([[:digit:]]):([[:digit:]])'
--coords-order xy
To match the read format `x:y:randomstuff`

Also removed an unused option in the help and a bug in the duplication chain code when the coordinates fail.


whitwham:
Partly added so people can test it and partly so the automated tests can run on non linux machines.

jkbonfield:
_\<Bleary-eye reminiscent mode>_ This suddenly reminded me of something from way back when - parsing Sanger-sequencing read names to extract data from within Pregap4 & Pregap4. :-)

Eg https://sourceforge.net/p/staden/code/HEAD/tree/staden/trunk/src/pregap4/naming_schemes/sanger_names_new.p4t

(Complete with broken URL as we all know web teams are rarely capable of keeping content working for more than a couple years.)

The approach was very similar:  A regexp to break a name into fields, and a list of field order.  Pregap4 had a bit more work to do as it then also had to map those fields from one value to another, eg "f" to "1".   Thankfully I think we have less complexity needed now.

daviesrob:
Looks good.  I've added a fix for a comment typo, and will merge shortly...

RandyHarr:
Firstly, I stand up and applaud this addition.  BGI/MGI/DNB sequencer QNAMES literally "hang" (infinite stall) samtools markdup nearly 50% of the time.  So we have to avoid the duplicate marking on these names; which are the majority of files we process.

I have tried every conceivable and simple form to get the REGEX matching for BGI QNAMEs with no success.  I either get the 
   "[markdup] warning: can not decipher X coordinate in E100003218L1C034R0294035623 ." or 
   "[markdup] warning: cannot decipher read name E100003218L1C034R0294035623 for optical duplicate marking." message. 
(side note: notice two forms of "cannot" and "can not" in the warning messages.)

As you can see, BGI/MGI/DNB sequencer names have much the same content as Illumina but just without colon separators.  I can easily match and manipulate the QNAMEs with the regex processor in Python. And have templates for the many varieties of QNAMES found.  (See http://bit.ly/2TfoAP2 if you really care.)  It does become a little tougher to write REGEX scripts as the length of the numerical fields can vary.  But the placement of the L, C and R are usually enough to figure out the rest reliably. 

I drop the template (t) processing from this example for simplicity.  The first warning shown above comes from using command line parameters of: 
  --read-coords '(C[0-9]{3})(R[0-9]{3})' --coords-order xy (and many other variations I have tried).  The second warning from using the BASIC POSIX REGEX form of parameters: 
  --read-coords '(C[0-9]\{3\})(R[0-9]\{3\})' --coords-order xy 

Your example uses "\w" which is PERL REGEX and not POSIX.  Otherwise, I would use "\d" instead of [0-9] but that generates the second warning message if used. I tend to abhor [[:digit:]] as it takes me back to the days of Ada.

As your examples do not have backslashes, I presume you are using EXTENDED REGULAR EXPRESSIONS in POSIX as opposed to BASIC. Would be useful to clarify. I tried both ways and think you cannot handle the backslashes of BASIC. Maybe see https://en.wikibooks.org/wiki/Regular_Expressions/POSIX_Basic_Regular_Expressions

Tried digging into the code changes but am not familiar with the libraries you are linking to and could not get very far.  Would love to get this resolved and working so I can turn back on the optical duplicate marking for BGI/MGI sequencer reads (which over 50% of the sequencer reads we see these days are coming from ...).

Thank you for the consideration.

whitwham:
Please can you add this as a new issue rather than as an addition to an already merged pull request.

PR #1566, opened 2021-12-15
Improves mpileup documentation.

### LINKED ISSUE #1564: mpileup command do not work properly when `--min-BQ` flag is set to 0
By default, `mpileup` command will only count overlap region of PE reads once, expect the `-x` flag is given. But when run `mpileup` commad with the `-Q/ --min-BQ` flag, and the argument is less than 1 (0 for example), **the mpileup command will count overlap region twice.**

jkbonfield:
- Document --min-BQ 0 disabling overlap removal.

- Document the lack of "B" aux tag support.

- Document "*" being used in fields where they would otherwise be empty.

Fixes #1564.  See also #1544.


daviesrob:
Just pushed a very minor formatting adjustment so the note added to `--min-BQ` is in its own paragraph.

PR #1579, opened 2022-01-20
Alphabetise samtools manual page synopses

jmarshall:
At <https://www.htslib.org/doc/samtools.html> these synopsis lines contain links to the individual manual pages. Alphabetising them enables users to find the right link when they're looking for details of a particular subcommand.

There is already an alphabetical list of subcommand links in the SEE ALSO section at the bottom of this page, but as the synopses are at the top of the page and more readably spaced out I suspect most users will be clicking on the links there rather than in SEE ALSO.

### DIFF
--- a/doc/samtools.1
+++ b/doc/samtools.1
@@ -43,75 +43,75 @@ samtools \- Utilities for the Sequence Alignment/Map (SAM) format
 .
 .SH SYNOPSIS
 .PP
-samtools view -bt ref_list.txt -o aln.bam aln.sam.gz
-.PP
-samtools tview aln.sorted.bam ref.fasta
-.PP
-samtools quickcheck in1.bam in2.cram
-.PP
-samtools head in.bam
-.PP
-samtools index aln.sorted.bam
-.PP
-samtools sort -T /tmp/aln.sorted -o aln.sorted.bam aln.bam
-.PP
-samtools collate -o aln.name_collated.bam aln.sorted.bam
-.PP
-samtools idxstats aln.sorted.bam
-.PP
-samtools flagstat aln.sorted.bam
-.PP
-samtools flags PAIRED,UNMAP,MUNMAP
-.PP
-samtools stats aln.sorted.bam
-.PP
-samtools bedcov aln.sorted.bam
-.PP
-samtools depth aln.sorted.bam
-.PP
-samtools ampliconstats primers.bed in.bam
-.PP
-samtools mpileup -C50 -f ref.fasta -r chr3:1,000-2,000 in1.bam in2.bam
-.PP
-samtools coverage aln.sorted.bam
-.PP
-samtools merge out.bam in1.bam in2.bam in3.bam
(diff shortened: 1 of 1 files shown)


jkbonfield:
I deliberately wrote them to not be alphabetical order, but clustered by type of command.  Eg all the indexing / dict things are together, fastq / import are adjacent, view and tview, etc.  See comment in 9813dbf1c007039cf07edb6240808e74be336d5c

jmarshall:
Thanks for the pointer. I am certainly not complaining that the per-subcommand summaries in _samtools.1_'s COMMANDS section is clustered by functional category.

What is your response to the suggestion that it is currently difficult for web page readers to find the right link to click on when they are hunting for a particular subcommand in the SYNOPSIS list? Do the maintainers agree that this is a problem that could be addressed?

Would it be better to revisit this, and use alphabetical ordering instead of subtle clustering in SYNOPSIS?

Alternatively subheadings could be added identifying each type-of-command cluster and setting the clusters off from each other (e.g. as in the usage display you get when you type just `samtools` by itself). Then the reader's eye would (hopefully) be drawn to the right cluster to find the subcommand they're looking for.

jkbonfield:
I take your point that it's duplicated in two regions - a description section and a one line SYNPOSIS.  So I think I concur that we can apply different orderings in the two parts.

I'm not entirely convinced the groupings are sensible, although some are obvious (merge/split, sort/collate) and they're better than the random looking ordering it had originally.  That's a different problem though.  It may be best addressed another time by adding sub-section headers to break the text up a bit and then sorting within each sub-section.

jmarshall:
Please forget I brought up anything about the text in the DESCRIPTION section :smile: That is a separate discussion, if it needs anything at all.

This PR is about the SYNOPSIS section, and in particular the difficulty in finding the link you are looking for when you're in a web browser at [_samtools.html_](https://www.htslib.org/doc/samtools.html) and you're wanting to click through to the detailed web page for a particular subcommand. (It is good that there is a succinct list with links to click on at the top of the web page, but it is bad that it is not obvious how the list's entries are ordered — so you have to hunt through it linearly and slowly.)

So here are two proposals for improving the SYNOPSIS section (in particular for the web view of the _samtools.1_ man page, but also would affect the man page hopefully positively):

1. Alphabetise it. (That is this PR.)
2. Make the clustering **visible**.

If you prefer to keep the existing clustering, I am happy to prepare an alternative PR that adds subheadings _to the SYNOPSIS section_ (i.e., a PR implementing (2)).

jkbonfield:
Sorry maybe I wasn't clear.

Not forgetting (:-)) the DESCRIPTION section - it's because this exists and is (now) brief, that I don't think it is a requirement to have function based grouping in SYNOPSIS.  Back when the man page was a huge monolithic block of text, I think grouping SYNOPSIS made sense as there was no other short section listing the tools.

Ie I agree with this PR. :-)

PR #1582, opened 2022-01-25
Improve automatic file type detection with view -u / -1.

jkbonfield:
Previously `-u` and `-1` were documented to also enable BAM mode. When they were written this made sense.  There were two choices - uncompressed SAM or BAM (uncompressed or otherwise).  Specifying the compression level inevitably meant BAM.

Since then we've gained the ability to output SAM.gz and CRAM, and even compressed fasta and fastq.  Plus we've also gained automatic file type selection by output filename. Hence it really doesn't make much sense for

    samtools view -o out.cram -1 in.bam

to output a BAM file named out.cram.

To ensure compatibility, the default output format is still SAM, or BAM when `-1` / `-u` are used, but if and only if a filename is used and it doesn't indicate otherwise.  This means the recommended pipe method of `samtools view -u foo.bam | ...` still does the same thing.

In the process of doing this I also fixed some bug in `--write-index` when used in conjunction with `-U`.

### DIFF
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -171,12 +171,17 @@ Output in the BAM format.
 Output in the CRAM format (requires -T).
 .TP
 .BR -1 ", " --fast
-Enable fast BAM compression (implies -b).
+Enable fast compression.  This also changes the default output format to
+BAM, but this can be overridden by the explicit format options or
+using a filename with a known suffix.
 .TP
 .BR -u ", " --uncompressed
-Output uncompressed BAM. This option saves time spent on
-compression/decompression and is thus preferred when the output is piped
-to another samtools command.
+Output uncompressed data. This also changes the default output format to
+BAM, but this can be overridden by the explicit format options or
+using a filename with a known suffix.
+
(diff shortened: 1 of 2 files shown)


jkbonfield:
I noticed after removing some new code that I now have a pointless variable (`out_un_format`), and the old code can be simplified further.    I'll update it soon.

jkbonfield:
I think it's ready again now to review.

Things to check:

- Does stdout work still (SAM by default, BAM with -1/-u, CRAM etc with -C)
- Do -o and -u work, together or separate?
- Does -h work?  Header needed for BAM/CRAM, but not output by default for SAM (even if gzipped). It does default to headers when SAM.gz is indexed with --write-index. (Doesn't work in 1.14)

Probably more that I've forgotten about!

jkbonfield:
Note I've also looked at a couple of other commands.  Collate and sort both have -u options, and in both cases it doesn't change the output format.  I haven't exhaustively tested every subcommand, but view is the main tool for format conversion and it would appear to also be the outlier in doing the unexpected.

PR #1590, opened 2022-02-02
Fix consensus includes and man page typos, update dependencies [minor]

jmarshall:
Some minor fallout after the merge of the consensus command:

Using `strcasecmp()` may require including <strings.h> on some platforms (POSIX does not require <string.h> to provide these functions).

Only source files, not header files, should include <config.h>.

Update Makefile dependencies: As consensus_pileup.h includes another HTSlib/Samtools header, the makefile needs a `$(consensus_pileup_h)`. Also update bam2bcf.o and bam_lpileup.o rules to reflect recent changes.

Fix man page typos, noticed because `.B--show-del` elicited a warning from man2fhtml.


jmarshall:
Wow, you folks are really having fun with your CI base images lately :smile: — Now Ubuntu bites the dust for spurious reasons:

```
Package ca-certificates is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
```

Makefile
```diff
@@ -156,14 +156,15 @@ bam_lpileup_h = bam_lpileup.h $(htslib_sam_h)
 bam_plbuf_h = bam_plbuf.h $(htslib_sam_h)
 bam_tview_h = bam_tview.h $(htslib_hts_h) $(htslib_sam_h) $(htslib_faidx_h) $(bam2bcf_h) $(htslib_khash_h) $(bam_lpileup_h)
 bedidx_h = bedidx.h $(htslib_hts_h)
+consensus_pileup_h = consensus_pileup.h $(htslib_sam_h)
```
jkbonfield, line 159: Thanks for these updates.

However given the samtools Makefile is entirely an internal thing, I'm wondering why we use this more complex approach compared to something more trivial such as:

```
consensus_pileup.h:  $(htslib_sam.h)
```

Why are we avoiding Make's builtin dependency analysis?  It's something I've never understood the background to.

jmarshall: I think if you try that, you'll find that it doesn't have the desired effect of causing _bam_consensus.o_ to be rebuilt if _htslib/sam.h_ gets updated.

Something like the following would do the trick, but IMHO changing the date stamps on the samtools headers would be artificial and confusing:

```make
consensus_pileup.h: $(htslib_sam_h)
	@touch $@
```

daviesrob:
Thanks for the updates.  And yes, it's been a bad CI week so far, it got there in the end though.

PR #1592, opened 2022-02-04
Fix/stats memory fail read header

mphschmitt:
Some blocks weren't freed when stats failed to read the header of a file.


daviesrob:
Thanks.  This is a step in the right direction, although I note that there are still plenty of other places where `remove_overlaps` can leak memory.  Eventually we really ought to make it return an error code as well, along with `collect_stats` so that they can properly report problems instead of attempting to continue as if nothing is wrong...

PR #1603, opened 2022-02-21
Add consensus to the samtools.1 man page.

jkbonfield:
Also fix a few minor problems in the samtoiols-consensus man page itself.

### DIFF
--- a/doc/samtools-consensus.1
+++ b/doc/samtools-consensus.1
@@ -58,7 +58,6 @@ samtools consensus
 .RB [ -H
 .IR het-fract ]
 .I in.bam
-.RI [ in2.bam ]
 
 .SH DESCRIPTION
 .PP
@@ -122,7 +121,7 @@ absent if it is a gap).  The
 option permits generation of ambiguity codes instead of "N", provided
 the minimum fraction of the second most common base type to the most
 common is above the
-.BI "--het-fract " fraction "" \fR.
+.BI "--het-fract " fraction .
 
 .SH OPTIONS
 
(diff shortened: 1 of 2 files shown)


babarlelephant:
Hi I'm having a hard time understanding the effect of min-depth, call-fract,  commands with -m bayesian

It is confusing because with -m simple it fits mostly my naive script parsing the .sam

jkbonfield:
As per the command line usage, they have no impact on the Bayesian mode.

```
For simple consensus mode:
  -q, --(no-)use-qual   Use quality values in calculation [off]
  -c, --call-fract INT  At least INT portion of bases must agree [0.75]
  -d, --min-depth INT   Minimum depth of INT [1]
  -H, --het-fract INT   Minimum fraction of 2nd-most to most common base [0.5]

For default "Bayesian" consensus mode:
  -C, --cutoff C        Consensus cutoff quality C [10]
   ...
```

The man page is more explicit "The  following  options  apply  only  to the simple (default) consensus mode:"

Although I note that it wrongly claims the simple mode is the default.  We changed this to the more accurate Bayesian method late in development but apparently didn't spot this line in the man page.

babarlelephant:
What is the unit of min-depth? I can retry but it seems to me that the unit is the number of reads in simple and something unclear in bayesian

jkbonfield:
Sorry if this still isn't clear.  It *does nothing* in Bayesian mode, so the unit is irrelevant.  (But yes you're correct in simple mode "depth" is a couting bases)

The Bayesian method is purely working on probabilities and doesn't make use of heuristics such as depth or fractions/ratios of call vs SNPs.

babarlelephant:
Thank you! This should explain the behavior :-)

Issue #1591, opened 2022-02-04
Appetite for a new sort order: TemplateCoordinate


nh13:
I wanted to gauge the appetite for a pull request for a new sort order, which we have in `fgbio` called [TemplateCoordinate](https://github.com/fulcrumgenomics/fgbio/blob/c26d405c55872b683dfa16ffe8d3c360db5ce898/src/main/scala/com/fulcrumgenomics/bam/api/SamOrder.scala#L167-L198).

This sort order is used by downstream tools that try to identify reads that observe the same source molecule.  This occurs when the original source molecule has one or more UMIs, and so the multiple observations (i.e read pairs) of the source molecule have the same UMI.  The reads observing the same source molecule can be used to create consensus reads.

For example UMI-tools and [`fgbio GroupReadsByUmi`](http://fulcrumgenomics.github.io/fgbio/tools/latest/GroupReadsByUmi.html) have to sort the input into this order, which is slow for large datasets.  

Sorts reads by:

1. the earlier unclipped 5' coordinate of the read pair
2. the higher unclipped 5' coordinate of the read pair
3. library (from read group)
4. the molecular identifier
5. read name
6. if R1 has the lower coordinates of the pair

In the SAM header, the sort order (SO) is `unsorted`, the group order (GO) is `query`, and sub-sort (SS) is `template-coordinate`.

I'd be open to attempting make a PR (command line option `-c`?), but don't want to invest effort unless there's commitment to review and accept a PR.

daviesrob:
I think this looks OK in principle.  I think a long option would be better than a short one, so we don't make the inconsistencies between samtools short options even worse.  It would also need a way of setting the UMI tag name.

I don't really like using the single string `template-coordinate` for the sub-sort order.  The [SAM specification](https://samtools.github.io/hts-specs/SAMv1.pdf#subsection.1.3) suggests that you can use a colon-separated sequence of descriptions, which would allow for a richer description of exactly how the file was sorted.  Unfortunately it doesn't have anything for "the earlier unclipped 5' coordinate of the read pair" though, so you would have to make up a name for that part.  @jkbonfield may want to comment on this aspect.

nh13:
[...]

PR #1605, opened 2022-02-27
Sort template coordinate

nh13:
This adds a new sort order to `samtools sort` and `samtools merge` called template-coordinate (see #1591 for more details).  I did a bit of refactoring to make it clearer (to me) the sort order being used in `bam_sort.c`, now that we have few different sort orders.

This also fixes a small bug where sub-sort is not removed from the header when the input header has a sub-sort but the output sort order does not.


bam_sort.c
```diff
@@ -2636,6 +2849,8 @@ int bam_sort_core_ext(int is_by_qname, char* sort_by_tag, const char *fn, const
         free(fns);
     }
     bam_destroy1(b);
+    if (sam_order == TemplateCoordinate) {
```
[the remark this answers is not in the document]
daviesrob: The number if items in use would be `k`.  You would need to free them after spilling to disk as well as at the end, or maybe check to see if the pointer has already been set so you can reuse them.  Although I'm not too enthusiastic with the idea of having a `malloc` per read as it potentially reintroduces the problems solved by PR #593, which were that `sort` didn't keep track of memory usage very well due to a lot of `malloc` overhead, and it spent a lot of time allocating and freeing small memory blocks.

You could get around this by putting `template_coordinate_key_t` directly in the `bam1_tag::u` union instead of using a pointer, but that would have the disadvantage of making the struct much bigger than it needs to be for sort orderings that don't need to look up the extra data.  I think better solutions might be one of:

1. Allocate `template_coordinate_key_t` structs in larger blocks, so you don't have so many of them to free.

2. Store the `template_coordinate_key_t` structs in `bam_mem` along with the BAM records.  This would have the advantage that they would be correctly accounted for in the memory usage.  You would need a way of dealing with the last record on each sorting round that fails to fit in the `bam_mem` block.

Option 1 is probably the easiest...

bam_sort.c
```diff
@@ -2740,12 +2739,20 @@ int bam_sort_core_ext(SamOrder sam_order, char* sort_tag, int minimiser_kmer,
...
             bam1_tag *new_buf;
+			template_coordinate_key_t *new_keys;
             max_k = max_k? max_k<<1 : 0x10000;
             if ((new_buf = realloc(buf, max_k * sizeof(bam1_tag))) == NULL) {
                 print_error("sort", "couldn't allocate memory for buf");
                 goto err;
             }
             buf = new_buf;
+			if (sam_order == TemplateCoordinate) {
+				if ((new_keys = realloc(keys, max_k * sizeof(template_coordinate_key_t))) == NULL) {
```
daviesrob, line 2750: This `realloc` is a bad idea because you're storing pointers into `keys` later on, which would become invalid if the data gets moved during reallocation.  Instead, you'd need to store a sequence of fixed buffers along with a list or array of where they start for use at demolition time.  Making them all the same size might be easiest as then you can just allocate a new one at the appropriate values of `k`, and the start points actually get stored in the `buf[k].u.key` pointers for the multiples of the array size you choose so there's no need to store them separately.

nh13:
@daviesrob thank-you thank-you for your patience and guidance, it's much appreciated.  I _think_ I've taken the approach you've asked for in terms of memory allocation.  Let me know what you think!

bam_sort.c
```diff
@@ -51,16 +51,85 @@ DEALINGS IN THE SOFTWARE.  */
...
+  template_coordinate_key_t **buffers; // the list of buffers
+} template_coordinate_keys_t;
+
+// Gets the idx'th key; does not OOB checking
+static template_coordinate_key_t* template_coordinate_keys_get(template_coordinate_keys_t *keys, size_t idx) {
+    size_t buffer_idx = idx / keys->buffer_size; // the index of the buffer to retrieve in buffer
+    size_t buffer_offset = idx % keys->buffer_size; // the offset into the given buffer to retrieve
+    return &keys->buffers[buffer_idx][buffer_offset];
+
+
```
nh13, line 85:
```suggestion
```

bam_sort.c
```diff
@@ -51,16 +51,85 @@ DEALINGS IN THE SOFTWARE.  */
...
+    assert(new_m * keys->buffer_size >= max_k);
+    if ((new_buffers = malloc(new_m * sizeof(template_coordinate_key_t*))) == NULL) {
+        print_error("sort", "couldn't allocate memory for template coordinate key buffers");
+        return -1;
+    }
+    // copy over existing buffers
+    int j;
+    for (j = 0; j < keys->n; ++j) {
+        new_buffers[j] = keys->buffers[j];
+    }
```
daviesrob, line 101: This looks a bit more complicated than strictly necessary - there's nothing wrong with calling `realloc()` on `keys->buffers`.

bam_sort.c
```diff
@@ -1916,16 +2010,131 @@ static inline int bam1_cmp_by_minhash(const bam1_tag a, const bam1_tag b)
...
+            return NULL;
+        }
+        // remove training /1 or /2, for duplex sequencing data
+        int i;
+        for (i = strlen(key->mid) - 1; 0 <= i; --i) {
+            if (key->mid[i] == '/') {
+                key->mid[i] = '\0';
+                break;
+            }
+        }
```
daviesrob, line 2059: Editing `key->mid` here is a bad idea as it edits the pointed-to aux data in a way that can't be undone. Depending on where the `MI` tag is, it either silently removes everything after the `/`, or it results in a `Corrupted aux data` error.

You need to either take a copy of the `MI` tag, or to avoid more memory-related faff write a custom string comparison function that ignores the `/1` or `/2` on the end.

jmarshall: I'm actually surprised that there's any need for trimming a `/1` or `/2` from an `MI` value. (In fact this code trims `/lots.of.words` too — anything from the last `/`.) There's no hint in _SAMtags.pdf_ that that is something that might be necessary.

If in practice you find that `MI` values are often just read names, including the `/1` or `/2`, and that you want to ignore that suffix for the purposes of picking out that unique ID for “All reads with the same MI tag represent the group of reads”, then this is something that the specification needs to address (or the tools doing this need fixing!). Please consider raising this as an hts-specs issue if so.

nh13: Will do. For duplex sequencing, it is useful to distinguish which strand on the original source molecule the reads originate, so a /1 or /2 is used. I’ll add a spec issue later today.

bam_sort.c
```diff
@@ -1591,13 +1683,17 @@ typedef struct {
...
     int i = heap->i, res;
     if (i < nfiles) { // read from file
         res = sam_read1(fp[i], hout, heap->entry.bam_record);
     } else { // read from memory
         if (in_mem[i - nfiles].from < in_mem[i - nfiles].to) {
-            heap->entry.bam_record = buf[in_mem[i - nfiles].from++].bam_record;
+            size_t from = in_mem[i - nfiles].from;
+            heap->entry.bam_record = buf[from].bam_record;
+            if (g_sam_order == TemplateCoordinate) heap->entry.u.key = template_coordinate_keys_get(keys, from);
+            in_mem[i - nfiles].from++;
```
daviesrob, line 1696: I'm also getting a segmentation fault from the heap sort stage.  I think the problem is that while you've still got the `template_coordinate_keys_t` data for everything in-memory, unfortunately you've lost it for reads that were spilled to disk.  It would therefore need to be recreated for the reads that are read back in again.  You wouldn't need that many - it should be one for every temporary file if the heap is working correctly.

nh13:
@daviesrob 

I updated the PR to:
    * update mid comparison to not modify the string and use a custom comparison function
    * simplify reallocate keys->buffers
    * create the key when reading from disk while merging
    
I ran on a large file, and made sure it wrote to disk (`-M 10M`), and back in again, with and without threads.

@jmarshall I'll get to the htsspecs request today

nh13:
@jmarshall here you go: https://github.com/samtools/hts-specs/pull/633

bam_sort.c
```diff
@@ -51,16 +51,71 @@ DEALINGS IN THE SOFTWARE.  */
...
+    size_t buffer_offset = idx % keys->buffer_size; // the offset into the given buffer to retrieve
+    //assert(buffer_idx < keys->m);
+    //assert(buffer_offset < keys->buffer_size);
+    return &keys->buffers[buffer_idx][buffer_offset];
+}
+
+// Rellocates the buffers to hold at least max_k entries
+static int template_coordinate_keys_realloc(template_coordinate_keys_t *keys, int max_k) {
+    size_t cur_m = keys->m;
+    keys->m += 0x100;
```
nh13, line 91: TODO: figure out how to expand memory

nh13:
@daviesrob another useful addition to this sort would be to fall back to the `RX` tag for the molecular identifier if the `MI` tag does not exist (see https://github.com/nh13/samtools/pull/4). The reason being that for some data types, this can lead to a memory reduction in the UMI grouping (i.e. assigning reads to source molecules and setting the MI tag) step.  


In the grouping step, where we actually set the MI tag, the reads with the same outer coordinates and library (keys 1-3 from https://github.com/samtools/samtools/issues/1591) are read in together.  Then the raw molecular barcodes are compared, sometimes with exact matching, and sometimes with using a [directed adjacency graph](https://doi.org/10.1101/gr.209601.116).  In the case that exact matching is used, if the reads with the same exact molecular barcode (RX tag) are grouped together in the input stream of reads, we don't need to read in all the reads into memory for a given set of outer coordinates.  For data where there are many reads with the same outer coordinate, this can be a big savings.

If not implemented, I _think_ this can also be done by copying the `RX` tag into the `MI` tag before running `samtools sort --template-cordinate`.  This may be preferred, but I want to raise this before we finalize this PR in case the fallback to RX when MI is not present is acceptable.

daviesrob (commented):
Actually, I am wondering if all the extra complication of ignoring characters after a trailing `/` is necessary.  If you only expect a single one, wouldn't the records collate right next to each other using `strcmp()`, anyway?

bam_sort.c
```diff
@@ -1916,16 +2005,158 @@ static inline int bam1_cmp_by_minhash(const bam1_tag a, const bam1_tag b)
...
+    if (0 == retval) retval = key_a->tid2 - key_b->tid2;
+    if (0 == retval) retval = key_a->pos1 < key_b->pos1 ? -1 : (key_a->pos1 > key_b->pos1 ? 1 : 0);
+    if (0 == retval) retval = key_a->pos2 < key_b->pos2 ? -1 : (key_a->pos2 > key_b->pos2 ? 1 : 0);
+    if (0 == retval) retval = key_a->neg1 == key_b->neg1 ? 0 : (key_a->neg1 ? -1 : 1);
+    if (0 == retval) retval = key_a->neg2 == key_b->neg2 ? 0 : (key_a->neg2 ? -1 : 1);
+    if (0 == retval) retval = template_coordinate_key_compare_mid(key_a->mid, key_b->mid);
+    if (0 == retval) retval = strcmp(key_a->name, key_b->name);
+    if (0 == retval) retval = strcmp(key_a->library, key_b->library);
+    if (0 == retval) retval = key_a->is_upper_of_pair == key_b->is_upper_of_pair ? 0 : (key_a->is_upper_of_pair ? 1 : -1);
+    return retval < 0 ? -1 : (retval > 0 ? 1 : 0);
```
daviesrob, line 2139: The order of comparisons here doesn't match the comment above...

bam_sort.c
```diff
@@ -1591,13 +1669,26 @@ typedef struct {
...
         res = sam_read1(fp[i], hout, heap->entry.bam_record);
+        if (res >= 0 && g_sam_order == TemplateCoordinate) { // file read OK and TemplateCoordinate order
+            if (keys->n >= keys->m * keys->buffer_size) res = template_coordinate_keys_realloc(keys, keys->n + 1); // need more memory
+            if (res >= 0) {
+                template_coordinate_key_t *key = template_coordinate_keys_get(keys, keys->n); // get the next key to use
+                heap->entry.u.key = template_coordinate_key(heap->entry.bam_record, key, hout); // update the key
+                if (heap->entry.u.key == NULL) res = -1; // key could not be created, error out
+                else keys->n++; // key created OK, increment the number of keys
+            }
+        }
```
daviesrob, line 1685: It looks like you'll use much more memory than needed here, as you're adding a new key for every single record read back in.  It should instead be possible to recycle them in the same way as the bam_record structs.  The easiest way to do this would be to use entry `keys->n + i` instead of just `keys->n`, without incrementing `keys->n`.

bam_sort.c
```diff
@@ -1916,16 +2005,158 @@ static inline int bam1_cmp_by_minhash(const bam1_tag a, const bam1_tag b)
...
+static inline int template_coordinate_key_compare_mid(const char* mid1, const char* mid2) {
+    int i = 0;
+
+    // compute the length
+    int len1 = strlen(mid1);
+    int len2 = strlen(mid2);
+
+    // shortcut: if the lengths differ, the shorter one is less than
+    if (len1 < len2) return -1;
+    else if (len1 > len2) return 1;
```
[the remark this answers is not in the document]
nh13: Moved to after trimming.  In practice, either all MIs have a trailing /1 or /2, or none do.

bam_sort.c
```diff
@@ -1916,16 +2005,158 @@ static inline int bam1_cmp_by_minhash(const bam1_tag a, const bam1_tag b)
...
+
+    // trim trailing slash and character
+    if (len1 >= 2 && mid1[len1-2] == '\\') len1 -= 2;
+    if (len2 >= 2 && mid2[len2-2] == '\\') len2 -= 2;
+
+    // find first mismatching character
+    while (mid1[i] != '\0' && mid2[i] != '\0' && mid1[i] != mid2[i]) {
```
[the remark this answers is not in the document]
nh13: fixed

bam_sort.c
nh13: Are you talking about the the fact we compare `(tid1, tid2, pos1, pos2, neg1, eng2)` versus the comment could be interpreted as `(tid1, pos1, neg1, tid2, pos2, neg2)`?  If so, I'll update the description here and hts specs.

bam_sort.c
```diff
@@ -1916,16 +2005,158 @@ static inline int bam1_cmp_by_minhash(const bam1_tag a, const bam1_tag b)
...
+    int len1 = strlen(mid1);
+    int len2 = strlen(mid2);
+
+    // shortcut: if the lengths differ, the shorter one is less than
+    if (len1 < len2) return -1;
+    else if (len1 > len2) return 1;
+
+    // trim trailing slash and character
+    if (len1 >= 2 && mid1[len1-2] == '\\') len1 -= 2;
+    if (len2 >= 2 && mid2[len2-2] == '\\') len2 -= 2;
```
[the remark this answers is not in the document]
nh13: fixed both

bam_sort.c
nh13: Are you suggesting to use the additional keys between `[k->n,k->m)` ?  I.e. the keys that have not been used but have been allocated?  I think that's what I'm doing.

We need enough memory to store the existing in-memory keys (already stored in`[0, k->n)`), and any additional keys that need to be made from bam records read from the file.  For the latter keys, why not add them into `keys` right after the existing in-memory keys?  

Or you saying I don't need to re-allocate the keys at all, and I can somehow use keys within `[0, k->n)`?  Or just assume `heap->entry.u.key` is not `NULL`?  

My apologies as I don't understand your suggestion.

nh13 (commented):
@daviesrob we want to ignore the trailing slash so the reads are query group ordered, but not sorted by query name.  R1 and R2 will have the same MI prefix, so they'll end up grouped if we ignore `/` and beyond, but not query grouped if we include them.

bam_sort.c
daviesrob: It's more that the comment says `library, MI, name` while the code does `MI, name, library`.  It sounds like you want what the code does? Although if `library` comes after `name` then I'd have thought it would be redundant and could be removed, as reads with the same name should all come from the same library.

nh13: Fixed.

bam_sort.c
nh13: Done!

nh13:
@daviesrob and @jmarshall I think I've addressed your latest comments.  Thank-you sincerely for your help so far!

nh13:
@daviesrob and @jmarshall no urgency implied, but simply the most gentle reminder.

jmarshall:
It's nowt to do with me. My only real interest is in samtools/hts-specs#633; BTW the most gentle reminder that there are some unanswered questions for @nh13 there…

bam_sort.c
```diff
@@ -1591,13 +1669,24 @@ typedef struct {
...
+                                uint64_t *idx, sam_hdr_t *hout) {
     int i = heap->i, res;
     if (i < nfiles) { // read from file
         res = sam_read1(fp[i], hout, heap->entry.bam_record);
+        if (res >= 0 && g_sam_order == TemplateCoordinate) { // file read OK and TemplateCoordinate order
+            // It is assumed that there are nfiles more keys allocated than keys->n; see allocation in bam_merge_simple
+            template_coordinate_key_t *key = template_coordinate_keys_get(keys, keys->n + i); // get the next key to use
+            heap->entry.u.key = template_coordinate_key(heap->entry.bam_record, key, hout); // update the key
+            if (heap->entry.u.key == NULL) res = -1; // key could not be created, error out
+            else keys->n++; // key created OK, increment the number of keys
```
daviesrob, line 1682: This line is no longer needed, and currently causes a crash as `keys->n` can go beyond the number allocated.
```suggestion
 ```

**Review by daviesrob — COMMENTED** [review]
Sorry, I've been a bit busy over the last few days.  I've found a couple of changes that are needed to fix crashes found by testing on an input big enough to cause data to be written out to temporary files.

**Comment by daviesrob** [comment]
I think we're getting close to finished with this, barring the issues listed above and discussions [elsewhere](https://github.com/samtools/hts-specs/pull/633) on `MI` tag usage.  One outstanding item here is that a section describing this will need to be added to the [sort manual page](https://daviesrob.github.io/www.htslib.org/doc/samtools-sort.html).  If you don't want to mess around with [roff](https://www.man7.org/linux/man-pages/man7/roff.7.html) you could always post some suggested text here and I can attempt to man-ify it.

**Comment by nh13** [comment]
@daviesrob I committed your two suggestions, then took a shot at updating the `samtools-sort` man page.  I also synced a missing note from the latter into `samtools`' man page in 4722382

**Comment by daviesrob** [comment]
This is looking better, and seems to be behaving itself now.  [`man 7 man`](https://linux.die.net/man/7/man) suggests putting a period after the number in numbered lists, but I can easily fix that up.  At the moment the `MI` tag discussion on samtools/hts-specs#633 is still ongoing, but I guess we can make changes later should the outcome be different to the anticipated one.

There's rather a lot of bugfix commits here.  I'll see if I can squash it down into something a bit easier to follow.

**Comment by daviesrob** [comment]
So, after a bit of messing around with git, I've managed to come up with a tidied up [set of commits](https://github.com/daviesrob/samtools/commits/152aa065).  The first does all the refactoring, then you get the template-coordinate change, then the new memory allocation and finally tests and man page changes.  Apart from some small man page adjustments, it should be identical to what you current have here.

If you're happy with it, I'll force-push it back here (so github records the merge properly) and then merge it in.

**Comment by nh13** [comment]
@daviesrob I like that history, thank-you for doing that.  I think only [this issue](https://github.com/daviesrob/samtools/commit/213df45c2f179f0f84ada3e6ccf2923d3cfa3a3f#r71889266) remains.

**Comment by daviesrob** [comment]
Thanks.  I have assimilated it into my local branch.  To make things easier I could push it back here - that will mildly confuse your local checkout, but you should be able to fix it up again by doing a `git fetch --force` from your `nh13/samtools` github repo to resynchronise it.  (To be safe, rename your existing local `sort-template-coordinate` branch first so you don't lose it in the process).

@jmarshall notes that the existing test doesn't have any `MI` tags in it.  How easy would it be to provide a more realistic (but not too big) test input?

**Comment by nh13** [comment]
@daviesrob I'd be happy to add some more tests.  Can you force push your local branch and then I'll add a few more tests?

**Comment by daviesrob** [comment]
OK, force-push done.  The only difference between the two branches is the trivial man page changes.

I've been looking at an improved test myself, which has spotted a couple of remaining issues:

- The `template_coordinate_key_compare_mid()` function doesn't work, the main problem being that the string comparison bails out on the first matching character instead of the first mismatching one.  There's also still an issue with the short-cut based on the string length, which I think could lead to inconsistent comparison results and probably isn't a good idea.  Fixing this shouldn't be too difficult.
- The library name comparison doesn't work.  This is because the `bam_get_library()` function returns the name in a [static buffer](https://github.com/samtools/samtools/blob/44b11db4/bam.c#L49) which gets overwritten on each call.  That means the values stored in the `template_coordinate_key_t::library` field will point to the same value.  Unfortunately this will be a bit more difficult to fix, at least if we want to be efficient and avoid making multiple copies of the same string value.

File: bam_sort.c [hunk 14]
```diff
@@ -1916,16 +2007,158 @@ static inline int bam1_cmp_by_minhash(const bam1_tag a, const bam1_tag b)
...
+    }
+
+    // compare last characters
+    if (mid1[i] == mid2[i]) return 0; // all characters match
+    else if (mid1[i] == '\0') return -1; // mid1 shorter
+    else if (mid2[i] == '\0') return 1; // mid2 shorter
+    else if (mid1[i] < mid2[i]) return -1; // mid1 earlier
+    else return 1;
+
+}
```
daviesrob, line 2040: I think this fixes the MI comparison, while keeping the slash-and-single-character behaviour.

```suggestion
/ * if mid1 is less than mid2, then -1 will be returned                         
// * if mid1 is greater than mid2, then 1 will be returned                       
static inline int template_coordinate_key_compare_mid(const char* mid1, const char* mid2) {
    size_t i = 0;
    size_t len1 = strlen(mid1);
    size_t len2 = strlen(mid2);
    size_t shortest;

    // Snip off trailing slash followed by a single character, if present        
    if (len1 >= 2 && mid1[len1-2] == '/') len1	-= 2;
    if (len2 >= 2 && mid2[len2-2] == '/') len2 -= 2;
    shortest = len1 < len2 ? len1 : len2;

    // find first mismatching character                                          
    while (i < shortest && mid1[i] == mid2[i]) i++;

    // compare last characters                                                   
    if (i == len1 && i < len2) return -1; // mid1 shorter                        
    if (i == len2 && i < len1) return  1; // mid2 shorter                        
    if (i == len1 && i == len2) return 0; // all characters match                
    if (mid1[i] < mid2[i]) return -1; // mid1 earlier                            
    else return 1;
}

```

nh13:
@daviesrob is there a better way to write unit tests than test files? I’d love to write a swath of test cases for the compare method.

jmarshall:
You could follow the example of _test/merge/*.c_ which do `#include "../../bam_sort.c"` and then add their own `main()` that contains tests exercising the functions defined in the included source file. They demonstrate writing unit tests that exercise C functions, rather than _test/test.pl_ which mostly contains tests exercising samtools commands.

jkbonfield:
To echo what John says and for completeness, there are 3 basic ways of doing tests.

1. Direct C API tests via custom C code.  The only way of testing at the function level.  This is probably what you need for your scenario.  Note the direct inclusion of the C file is required in order to test static functions.  Don't look too hard or you'll feel queasy. :-)   (This is much cleaner in htslib given it's designed to be used as a library.)

2. The perl test/test.pl script to drive samtools and validate the output.  Being perl it can do pre- and post-processing and is flexible.   Personally I find this script rather impenetrable and more than a little bit tedious to edit.

3. A simpler bourne shell test harness which contains a series of commands and their expected output, eg tests/mpileup/*.reg.  Limited and restrictive, but concise.  See "regression.sh" in Makefile for example usage.

Linking against a small bit of samtools in order to validate it is a little bit cumbersome so often methods 2 and 3 are easier to do.   Fortunately in this case there is a previous demonstration of testing bam_sort.c directly so it's a good starting point to copy.

jmarshall:
+1 re queasy :smile:

Linking against what you need from samtools is what `libst.a` is for. So you can copy what the _Makefile_ does for `test/merge/test_bam_translate` et al.

jkbonfield:
> Linking against what you need from samtools is what `libst.a` is for. So you can copy what the _Makefile_ does for `test/merge/test_bam_translate` et al.

True, but it also tempts people into monstrosities such as turning samtools into a library to embed without their own applications (and then getting shocked that incorrect arguments to a function leads to `exit(1)` calls, hah).   I'd rather it was done via a Makefile `$(OBJ)` type macro instead of .a file creation.

That's not to say that some of the functionality wouldn't be good to add to a library, but none of the code is implemented with that in mind and it doesn't follow any of the ABI compatibility validation that we apply to things like htslib, nor restrictions such as not calling exit and even sometimes memory leak tidyness on broken input data.

I dread to think how many ghastly GUI apps or language bindings there are that wrap up samtools this way, that display usage followed by exit if you attempt to specify a filename starting in a minus sign ;-).  I bet they exist!

daviesrob:
I think this is drifting somewhat off topic, which considering the length of conversation so far here is not a good thing.  At the moment I'm using the files in the attached [template-coordinate.zip](https://github.com/samtools/samtools/files/8611940/template-coordinate.zip) archive for test material.  The `.expected.sam` was made by manually reordering the lines according to the documented sort order.  It's a start but it isn't perfect as while it shows up the library problem it misses the `MI` one.  It could really do with a bit of adjustment so that name comparison always gives the wrong answer if it gets there because one of the earlier ones incorrectly returned zero.

Plugging it into the test harness is easy, you just do:
```
    test_cmd($opts, out=>"sort/template-coordinate.sort.expected.sam", ignore_pg_header => 1, cmd=>"$$opts{bin}/samtools sort${threads} --template-coordinate -m 10M $$opts{path}/sort/template-coordinate.sort.sam -O SAM -o -");
```
in place of the existing test.  It's already had fixmate run on it, so there's no need for that bit.

jmarshall:
> True, but it also tempts people into monstrosities such as turning samtools into a library

On-topic: _libst.a_ Itself tempts no such thing, because very little is in _libst.a_.

73c4054a91d6ea357802e85b8a68b2a5c97ce061 added this as a library to prevent potential unused warnings from the linker if nothing from a particular translation unit was used by a particular test executable, as might occur if such translation units were listed directly on the linker command line; but if you prefer you could change the _Makefile_ to use `$(LIBST_OBJS)` directly (and probably rename it to avoid `LIB…`) and hope that the linker does not improve to emit such warnings in your `-Werror` builds.

Off-topic: Seriously, @jkbonfield, you should stop whining about that minor aspect of pysam. It's very boring. (And the `exit(1)` exiting issue was fixed three pysam releases ago.)

daviesrob:
My proposed fix to the `MI` comparison above, and one for the library name problem can currently be found [here](https://github.com/daviesrob/samtools/commits/pr1605_squash).  That branch includes a corrected version of my fairly basic test.  (There's also an unrelated commit to fix a recent CI problem, which can mostly be ignored as it should disappear on merge).

If you're happy with them, please copy the new commits to your branch which will add them in here.

nh13:
@daviesrob I've pushed your branch to this one, so that should all be set.  I still want to spend a few hours writing some comprehensive tests.

nh13:
@daviesrob this is still on my list todo, but myself and my family have been sick for this past month, so will try to get to it in June.  My apologies for the delay

daviesrob:
@nh13 I'm sorry to hear that, hope you're all feeling better.  If necessary, I could try expanding the minimal test we already have to make sure it hits all of the ordering rules.  I think that would give reasonable confidence that this works.

nh13:
I’d welcome help at this point. Again, my apologies.

daviesrob:
I've had a go at expanding the test file.  The new unsorted file can be seen [here](https://github.com/daviesrob/samtools/blob/ecc5d5076b/test/sort/template-coordinate.sort.sam) together with the [expected output](https://github.com/daviesrob/samtools/blob/ecc5d5076b/test/sort/template-coordinate.sort.expected.sam).

While working on this, there were a couple of things that I noticed:

- `tid1` and `tid2` are both [compared](https://github.com/samtools/samtools/blob/1f76f4a185/bam_sort.c#L2137) before `pos1` and `pos2`.  This means pairs that span two different references get sorted into their own group.  So, for example,  in the test file, pairs that span `ref1` and `ref2` [appear after](https://github.com/daviesrob/samtools/blob/ecc5d5076b/test/sort/template-coordinate.sort.expected.sam#L18) all the `ref1`-only pairs.
- I found it quite difficult to get the [`neg` comparison](https://github.com/samtools/samtools/blob/1f76f4a185/bam_sort.c#L2141) to do anything.  As far as I can see, they set by the relative direction of the reads in a pair.  For pairs pointing towards each other `neg1` is always false and `neg2` is always true.  So for these comparisons to show anything other than a match, you need pairs where the positions coincide exactly at each end and where the two reads in each pair are in different relative orientations.  Do you know how likely this is to happen in practice?  If it's not often then these comparisons may not be adding much.

If the output order is in line with your expectations, then we can pull the new test file in here.

nh13:
@daviesrob Thank-you for all the hard work you have done.  I think these align with my expectations, so feel free to pull the new test file in here!  Details below.

The behavior you noticed in your first bullet point is by design.  We are relying on mapping information to sort into groups so then we can look at read pairs that have the same start/stop, which includes reads that map inter-reference.  For consensus calling, we create groups of read pairs that are equal to the [mid comparison](https://github.com/samtools/samtools/blob/1f76f4a18505b530feef70b018cfa1abb1ddfcbf/bam_sort.c#L2144), and then consensus call their bases/sequence.

The neg comparison occurrence is very very rare.  We haven't profiled how often each of these comparisons actually differ, but from a correctness point of view (and to match the existing implementation in fgbio), I think we'll need to keep it.  To be honest, I think we are so used to having a primary sort key be chr/pos/strand.  I wouldn't say such things are impossible, as you do have inversions (biological) and template switching (library prep) that can lead to strange things.  And when you're trying to find very very rare events, you want to be sure those rare events are not due to ignoring read-pair orientation.

daviesrob:
OK, I'll pull it in and see if I can get the result in a state where it can be merged.

daviesrob:
Merged at last.  thanks for all your work on this.

nh13:
Herculean effort on your part @daviesrob, kudos and thank-you!

PR #1617, opened 2022-03-16
Fix view -L in.bed -M.

jkbonfield:
The -M operator upgrades the -L filtering to become an index-jumping query.  However this was broken by the fetch-pairs PR (487f2019).

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -1230,7 +1230,7 @@ int main_samview(int argc, char *argv[])
         print_error("view", "Incorrect number of arguments for -X option. Aborting.");
         return 1;
     }
-    if ( settings.fn_idx_in || nregs )
+    if ( settings.fn_idx_in || nregs || settings.multi_region )
     {
         settings.hts_idx = settings.fn_idx_in ? sam_index_load2(settings.in, settings.fn_in, settings.fn_idx_in) : sam_index_load(settings.in, settings.fn_in);
         if ( !settings.hts_idx )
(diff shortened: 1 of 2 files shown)


jmarshall:
Is there a way to add a test case to ensure nothing similar happens again?

(It's not obvious how to construct a test case that distinguishes between streaming and index jumping just by looking at the output. It might need to use timing or `strace`, or perhaps add some instrumentation that could be turned on via `hts_verbose` levels.)

jkbonfield:
Timing isn't going to be a robust test case without the file being sufficiently large and slow to make it reliably significant.  I don't want that in our test harness really.

However as you say `-U` could work.

jmarshall:
For reasons that you've been discussing on Twitter recently, it seems important for the test suite to verify that samtools invocations that should use index-jumping actually do so.

jkbonfield:
Agreed, although that specific issue isn't cause by this bug.  Without this fix the command simply fails because no index was loaded and it promptly errors out:

```
samtools view: Iterator could not be created. Aborting.
```

I haven't yet worked out what scenario causes -M plus -L to ignore the index but still continue to completion.  Possibly just an old release, but it's embarassing the very latest code doesn't work.   Version 1.6 predates -M and it just issues a warning but still continues (slowly) without index hopping.  1.7 onwards uses -M to optimise -L.  Removing the index gives a hard error too - the same deal as we have in 1.15 basically.

I'm just confused how -M could ever *not* utilise the index without also at least warning somewhere.

daviesrob:
The problem was more that there were no tests for the `-M` option, so we didn't spot the bug at the time #1542 went in.  I've added a commit here to add a couple of checks, so we're less likely to get an accidental regression again.

PR #1635, opened 2022-04-18
coverage: update format string [trivial].

mphschmitt:
Requires 'int' but the argument type is 'unsigned int'.

### DIFF
--- a/coverage.c
+++ b/coverage.c
@@ -252,7 +252,7 @@ void print_hist(FILE *file_out, const sam_hdr_t *h, const stats_aux_t *stats, in
         fprintf(file_out, full_utf ? VERTICAL_LINE : "|");
         fputc(' ', file_out);
         switch (i) {
-            case 9: fprintf(file_out, "Number of reads: %i", stats[tid].n_selected_reads); break;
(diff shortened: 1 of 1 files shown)


daviesrob:
This is correct, although out of interest how did you detect this?  If you have an input with more than 2Gi reads, it might be worth upgrading the variable to a 64-bit one so that the problem doesn't reassert itself if your input doubles in size.

Compiling with `gcc -Wformat-signedness` finds a few more instances of this including on the next line.  If we decide to stick to 32 bits here, they can be easily sorted in a follow-up pull request after merging this one.

mphschmitt:
I run cppcheck on a project that uses samtools (coverage, mostly) as a dependency, and this came out as a warning.
I do not expect the input to be that big, so there is probably no need to increase the size of the variables.

This did not happen during software execution, however, so we can discard this commit if you don't think it's worth it.
Otherwise, I may find the time to do the follow-up pull request if you wish.

daviesrob:
OK, In that case I'll just push this one as-is and try to remember to fix the other ones later.

PR #1649, opened 2022-04-29
Cram embed ref

### LINKED ISSUE #723: How to extract embedded reference from CRAM file?

jkbonfield:
Adds a cram2ref subcommand which extracts the embedded reference out of a CRAM file.  (Ie one that was built using `view -O cram,embed_ref=1`)

Fixes #723 (or at least partially).  A full fix would require whoever produced the assembly and aligned data against it to look for zero-coverage regions and produce synthetic reads to patch the holes so the consensus is 100% covered, but arguably if you align raw data against your assembly and find places with zero coverage then I'd argue the consensus there wasn't worth storing as it's apparently wrong!


jkbonfield:
NB also see https://github.com/jkbonfield/htslib/blob/cram_embed_ref2/cram/cram_encode.c#L2739-L2851

This is an htslib copy of two functions first implemented here.  I'll work on tidying up the API there some and then cull them here.

That code permits CRAM to embed references by dint of computing it on-the-fly (assuming MD:Z tag exists - todo: make sure it does!).

doc/samtools-reference.1
```diff
@@ -0,0 +1,110 @@
...
+.  in
+
+..
+.
+.SH SYNOPSIS
+.PP
+samtools reference
+.RB [ -q ]
+.RB [ -r
+.IR region "] "
```
jmarshall, line 47:
```suggestion
.IR region ]
```

doc/samtools-reference.1
```diff
@@ -0,0 +1,110 @@
...
+will hold the entire reference that spans the slice coordinates, but
+gaps in coverage can lead to gaps between slices.  However this
+reference should be suitable for passing into a CRAM decode
+(\fBsamtools view -T ref.fa\fR).
+
+For SAM/BAM files or CRAMs without reference, using the MD:Z tag may
+also produce an incomplete reference.  Unlike embedded reference, this
+reference may not be sufficient for decoding a CRAM file as the CRAM
+slice headers store the MD5sum of the portion of reference than spans
+that slice, but the slice may not have 100% coverage leadng to Ns in
```
jmarshall, line 69:
```suggestion
that slice, but the slice may not have 100% coverage leading to Ns in
```

doc/samtools-reference.1
```diff
@@ -0,0 +1,110 @@
...
+Specifies a single region to produce the reference from.  If
+specified, an index file must be present.
+
+.TP 8
+.BI "-o " FILE
+Write the FASTA records to \fIFILE\fR.  By default this is sent to stdout.
+
+.TP 8
+-BI "-@ " INT
+The number of BAM/CRAM decompression threads to use in addition tot he
```
jmarshall, line 95:
```suggestion
.BI "-@ " INT
The number of BAM/CRAM decompression threads to use in addition to the
```

reference.c
```diff
@@ -0,0 +1,591 @@
...
+        }
+    }
+
+    while ((r = iter
+                ? sam_itr_next(in, iter, b)
+                : sam_read1(in, h, b)) >= 0) {
+        // check b->core.tid and flush old seq.
+        if (b->core.tid != last_tid) {
+            if (last_tid >= 0)
+                dump_ref(h, iter, last_tid, ref, ref_len, outfp, 1);
```
whitwham, line 451: Should dump_ref verbose always be 1 here?  The same question for the other dump_refs in this function.

jkbonfield:
Thanks.  I added a remove to fix the `M` option and honour `verbose`.  I also added more error checking to `dump_ref` calls.

I can squash and rebase this if you've finished review, but did a separate commit initially so you can review the fixes.

whitwham:
Look good to me.  If you squash it I will merge it.

jkbonfield:
Rebased and squashed.

PR #1659, opened 2022-05-26
Add documentation on CRAM compression profiles.

### LINKED ISSUE #1656: Compression presets for cram
To make cram compression more user friendly, some algorithm presets for `--output-fmt-option` would be nice to have.

Add preset options for `--output-fmt-option` as implemented in [scramble](https://github.com/jkbonfield) by @jkbonfield.
Examples for presets are in the README, with corresponding settings in the code.

https://github.com/jkbonfield/io_lib/blob/master/io_lib/cram_io.c#L6392-L6422
https://github.com/samtools/htslib/blob/develop/cram/cram_io.c#L5805-L5838

jkbonfield:
Also document the newer options that appeared with CRAM 3.1 and above.

Fixes #1656


doc/samtools.1
```diff
@@ -937,18 +937,56 @@ block compression.
 CRAM output only; defaults to 0 (off).  Permits use of lzma in CRAM
 block compression.
 .TP
+.BI use_fqz= 0|1
+CRAM >= 3.1 output only; enables and disables the fqzcomp quality
```
jmarshall, line 941: These could be pretty-printed as `≥`, which makes a difference to the HTML view on http://htslib.org/doc and also in `man samtools` if your terminal is in an UTF-8 locale (which it probably is, these days):

```suggestion
CRAM \(>= 3.1 output only; enables and disables the fqzcomp quality
```

jkbonfield: Thanks John.

Issue #1569, opened 2021-12-22
standardize I/O nomenclature across all tools?


eboyden:
Although most tools can read from either a file or stdin and write to either a file or stdout, the manner in which to specify that seems to vary from tool to tool.  It's not consistent when stdin/stdout is assumed unless a file is specified (as with `sort`), vs. when a file or stdin/out must be specified but an option is not required (as with `fixmate`), vs. when an option is required for output but not input (as with `addreplacerg`).

Is there any possibility of standardizing the I/O across all tools to lessen the confusion?  With over 30 tools it's difficult to remember how they all behave.  Personally I would vote to always require `-i` and `-o` for specifying a file name (but perhaps assuming stdin/out in the absence of either option, though `-i stdin` and `-o stdout` would also be supported).

Poshi:
I strongly agree with this idea. In fact, I've been tempted to open an issue for similar considerations.
Using `-i stdin` could be ok, but I think that `-i /dev/stdin` would be a better option, as it just means that you should read the given file, and not interpret the `stdin` string as the marker for standard input. Keeping `-` as standard also sounds good, as it has been used for a long time.

But unifying input and output parameters is a must for ease of use and to lessen the (potentially catastrophic) errors: let's say that we want to index all bam files in a folder. Common sense say you can try `samtools index *.bam`. But this will likely cause the overwrite of the second bam in the list by the index of the first bam (OK, maybe I will get an error saying that the file already exists, but the point is that it is very dangerous and error prone).
Another fun trick is to run `samtools merge *.bam > merged.bam`. This will result on the first bam being overwritten by the merge of the other bam files, and an empty `merged.bam` file.
Because sometimes the output file is the first in the list (like in merge) or the last of the list (like in index).

I would advise having all the output files specified via parameter; no bare output files in the command line. And for homogeneity, the input files should follow the same policy.

jmarshall:
Samtools follows the usual [Unix conventions](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) and in particular the [Unix filter conventions](https://en.wikipedia.org/wiki/Filter_(software)).[^1] Like the overwhelming majority of Unix tools, non-option arguments are input filenames, `-` in the input filenames indicates standard input, an empty list of input filenames means to read from standard input, and output can be redirected with `-o FILE` (typically) or via the shell (which is usually an inferior technique for production use).

[...]

PR #1674, opened 2022-07-05
Enable `samtools index` to index multiple alignment files at once

jmarshall:
Given a directory containing three BAM files, _sampleA.bam_,  _sampleB.bam_,  and _sampleC.bam_, one might expect the following to index the three of them individually:

```
samtools index sample*.bam
```

Instead it will overwrite _sampleB.bam_ with A's index, and ignore its third argument — as reported [here](https://twitter.com/_nicoromano_/status/1496859773727301633), [here](https://twitter.com/AbigailRamsoe/status/1511666622502785035), and in https://github.com/samtools/samtools/issues/1569#issuecomment-1000201931, in at least two cases likely from bitter experience.

It would be easy to catch this case (i.e., to detect that the arguments are not the expected `<in.bam> [out.index]` and fail with an error message) by actually validating `argc`. However if there were only two files matching `sample*.bam` it would be harder as `argc` would have its expected value and you would need to check whether the second filename actually referred to an extant BAM/CRAM file. If you were going to do that, it's equally easy to implement the `*.bam` behaviour that these users are expecting.

Hence this PR, which regularises `samtools index` to follow the usual Unix conventions (in addition, for compatibility reasons, to its existing baroque command line syntax):

Accept either `FILE [FILE]...` or `FILE [OUTINDEX]`. When exactly two filenames are given, interpret the second as OUTINDEX if it does not exist or if it already contains index data. In particular, do not interpret an existing BAM/CRAM file as an output index filename!

Add optional `-o OUTINDEX` option that can be used instead of OUTINDEX, to align this subcommand with others that use `-o` to specify output files. At present this can only be used when 1 input file is given. In future, it could include expansion metacharacters similarly to the split subcommand.

In principle, we could use a thread pool to index multiple input files in parallel (as well as multi-threading the input reading). This would be a larger change and, at least initially, probably not worthwhile.

### DIFF
--- a/bam_index.c
+++ b/bam_index.c
@@ -28,10 +28,10 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include <htslib/hts.h>
 #include <htslib/sam.h>
+#include <htslib/hfile.h>
 #include <htslib/khash.h>
 #include <stdlib.h>
 #include <stdio.h>
-#define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 #include <unistd.h>
 #include <getopt.h>
@@ -44,12 +44,30 @@ DEALINGS IN THE SOFTWARE.  */
 static void index_usage(FILE *fp)
 {
     fprintf(fp,
-"Usage: samtools index [-bc] [-m INT] <in.bam> [out.index]\n"
+"Usage: samtools index [-bc] [-m INT] [-o <out.index>] <in.bam>...\n"
+"   or: samtools index [-bc] [-m INT] <in.bam> [out.index]\n"
 "Options:\n"
 "  -b       Generate BAI-format index for BAM files [default]\n"
 "  -c       Generate CSI-format index for BAM files\n"
 "  -m INT   Set minimum interval size for CSI indices to 2^INT [%d]\n"
+"  -o FILE  Write index to FILE [alternative to <out.index> as an argument]\n"
 "  -@ INT   Sets the number of threads [none]\n", BAM_LIDX_SHIFT);
+}
+
+// Returns 1 if the file does not exist or can be positively
+// identified as an index file.
+static int nonexistent_or_index(const char *fn)
+{
+    int ret1, ret2;
+    htsFormat fmt;
+    hFILE *fp = hopen(fn, "r");
+    if (fp == NULL) return 1;
+
+    ret1 = hts_detect_format2(fp, fn, &fmt);
+    ret2 = hclose(fp);
+    if (ret1 < 0 || ret2 < 0) return 0;
+
+    return fmt.category == index_file;
 }
 
 int bam_index(int argc, char *argv[])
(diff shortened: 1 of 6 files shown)


bam_index.c
```diff
@@ -28,10 +28,10 @@ DEALINGS IN THE SOFTWARE.  */
 
 #include <htslib/hts.h>
 #include <htslib/sam.h>
+#include <htslib/hfile.h>
 #include <htslib/khash.h>
 #include <stdlib.h>
 #include <stdio.h>
-#define __STDC_FORMAT_MACROS
```
jmarshall, line 34: (This macro only ever had an effect in C++; there's never a need for it in a C source file.)

delocalizer:
I like the intent but solely to address principle of least violation it would be simpler to not accept multiple positionals at all and have the custom index name specified with the -o OUTINDEX style argument.
Supporting processing multiple inputs doesn't really seem like it adheres to "make each program do one thing well". Also, if you ever support `-o -` to write to stdout what would that mean for multiple bams
[disclaimer: not a core samtools dev]

jkbonfield:
I haven't reviewed it yet, but I'm in agreement that we cannot deprecate the old syntax, and it seems sensible to add error checking to protect against silly accidents such as `index -o *.bam`. 

Samtools merge has the hideous `-X` option, which permits explicit naming of the indices after the bams, so a.bam b.bam c.bam a.bam.bai b.bam.bai c.bam.bai.  It's pretty horrific, but was required by people using custom index names.  I don't think that's really important here though as merge *requires* multiple files while in index it's just a nicety and you can always do a shell loop instead.

jmarshall:
The proposed code does not check for unintended overwriting if `index -o *.bam` expands to exactly two filenames. That didn't occur to me, and IMHO it's not greatly important to guard against it: most people typing `-o` will be primed to type their desired output filename after it, so this will not be a common failure mode.

Re `-X`, there is also the hideous `##idx##` syntax. This could probably be used to specify the index filenames when doing lots at once, but I dislike it so I haven't tested it. But agreed there's no pressing need to enable that when a shell loop is available for the fiddly cases.

daviesrob:
I'm afraid I'm not as optimistic about how quickly new versions get adopted.  Experience shows that old versions tend to hang around for rather a long time.

daviesrob:
Out of caution, I'd prefer to see the gated `-M *.bam` solution.  Possibly with some assistance so that if you forget it, `samtools index` outputs something like "I see you're trying to index lots of files.  Please try using the `-M` option."

jkbonfield:
A small data point: it turns out that the old (ie existing) code will take `samtools index *.bam` with 100 bam files and not complain.  Sigh.  It does the annoying thing of overwriting the second bam.  Blergh.

Even so, I favour making life easy for future users more than easy for historic ones who are reading the wrong documentation. :)

daviesrob:
Despite James' comments above, I still want to have a `-M` option to enable the multi-file mode.  Looking at the revised manual page, there's a lot of explanation about when the various methods of invoking program can and can't be used.  I think this gets simpler if it becomes a choice of one of these three methods:

The new way, one file only:
`samtools index [-o file.csi] file.bam`

The old way, one file only:
`samtools index file.bam [file.csi]`

Multiple files:
`samtools index -M file1.bam [file2.bam ...]`

The `-M` ensuring that the multi-file one is never confused with either single-file version.

jmarshall:
The man page explanations added are my interpretation of your previous request in https://github.com/samtools/samtools/pull/1674#issuecomment-1178736692. It is true that the explanation could be largely simplified if `-M` were required, at the cost of making this small part of the user interface suck a little more for future users.

Could the maintainers please just tell me what they collectively want me to do. :smile: Rather than Individual maintainers expressing their preferences and wants and what they favour, please instruct the external contributor by making a decisive imperative statement of the form “please change the PR so that `-M` is required for the multiple file form” — if that is what is required for this PR to progress.

daviesrob:
The precise statement _is_ “please change the PR so that -M is required for the multiple file form”.  If anyone contradicts this, please ignore them.

jmarshall:
Updated to require `-M` to enable the multiple-files functionality. It is fair to say that this does remove the need for a lot of excess verbiage in the manual page. :smile:

Hopefully in future when affected old versions have receded into the past, we may be able to make `-M` a no-op, similarly to `view -S` which is ignored as it has not been necessary for many years.

daviesrob:
Thanks for adding the `-M` option.  Would it be OK to rebase/squash this to resolve the conflict and get all the tests working again?

jmarshall:
Personally I'd quite like to preserve the story on this one, although I suppose the separate commits are recorded for posterity in this PR.

How about I rebase it past the other NEWS addition and rocky fix?

daviesrob:
That would be fine, thanks.

jmarshall:
Done, tests should all succeed this time.

Feel free to rebase it all the way to **origin/develop**'s current head and/or squash it if you prefer.

PR #1676, opened 2022-07-11
Add dict -l/--alt option to add AH based on bwa .alt file

jmarshall:
A companion to lh3/bwa#348 that further extends `samtools dict`'s ability to produce `@SQ` headers that represent everything that BWA's built-in default headers can represent. (In particular, `bwa`'s built-in header writing code will add `AH:*` for sequences listed in the _.alt_ file.) —

New option for the 'dict' command that reads a bwa(1)-style .alt file and adds `AH:*` to the alternate loci sequences listed in the file.


jkbonfield:
It's a tad unfortunate that `--alt` matches the prefix of `--alternative-name`, as using this on an older samtools gives totally junk output due to insufficient error checking in earlier releases.  I don't think there's a great deal we can do about that sadly as `--alt` is absolutely the right long opt to use.

I don't know how commonly `AH` gets used in the various modes, but `AH:*` is listed as the format to use when you don't know the chromosome (or chr:pos).  When we do know it, this code ignores this information.  If this matters then it's perhaps the subject of a future issue if you wish to get this PR in as it stands (it's still clearly useful as-is).  If so, is it one you'd feel you want to address yourself?  Now, or at a later stage?

jmarshall:
The `AH:*` currently implemented produces the same output as bwa itself (see _bwa.c_'s `bwa_print_sam_hdr()`), which only ever outputs a value of `*`. I would see producing the actual coordinates (where available) as a future enhancement.

PR #1679, opened 2022-07-14
Allow '*' for -T to write all tags to fastq

cjw85:
The command `samtools import` allows `-T "*"` to read all aux data from fastq headers. This PR makes `samtools fastq` symmetric by allowing all tags from SAM records to be written to fastq headers.

The motivation for this was for use with minimap2 to align reads with modified base data stored in unaligned SAM files (and not burden the user with knowing all tags).

### DIFF
--- a/bam_fastq.c
+++ b/bam_fastq.c
@@ -349,7 +349,7 @@ void set_sam_opts(samFile *fp, bam2fq_state_t *state,
 
     hts_set_opt(fp, FASTQ_OPT_BARCODE, opts->barcode_tag);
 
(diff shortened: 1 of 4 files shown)


cjw85:
@jkbonfield I presume the rockylinux test is failing here only because things are running from my fork?

cjw85:
Ooh, I think I noticed that `-T ''` would have the effect but forgot about it.

I'm on vacation at the moment, feel free to tidy this up.

daviesrob:
Ok, I'll make the necessary adjustments.

daviesrob:
Adjustments made.  Apologies for the forced push.

PR #1683, opened 2022-07-21
Fix missing `@RG` headers when using `samtools merge -r`

daviesrob:
Caused by a broken test for an empty klist.  Fix by checking the size stored in the klist struct.  It's not clear if it's an officially supported way of testing for an empty klist (there is currently no macro to expose size), but it works for now.

Adds a test, as the case of no @RG headers in the inputs was previously not covered.

See samtools/htslib#1479

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -630,7 +630,7 @@ static klist_t(hdrln) * trans_rg_pg(bool is_rg, sam_hdr_t *translate,
     }
 
     // If there are no RG lines in the file and we are overriding add one
-    if (is_rg && override && kl_begin(hdr_lines) == NULL) {
+    if (is_rg && override && hdr_lines->size == 0) {
         kstring_t new_id = {0, 0, NULL};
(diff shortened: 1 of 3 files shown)


jmarshall:
If you wanted to use the existing macros to check for emptiness, I think this would be the way to do it:

```c
if (kl_begin(hdr_lines) == kl_end(hdr_lines)) { …list is empty… }
```

daviesrob:
I did consider doing it that way, but decided it's probably worse.  It wouldn't be too unreasonable to expect `kl_begin(x) == kl_end(x)` to be true for a one-element list, and it's only by carefully inspecting the klist code that you find it isn't.  Checking `size` seems more clear.

I'm tempted to add a macro to expose `size` in the klist code, so it becomes more officially supported.

jmarshall:
Most of the existing uses of klists is in iterator-based loops, so is already effectively using that same test (which is pretty familiar from other k-iterator code):

```c
```

PR #1684, opened 2022-07-25
add mapq histogram to samtools stats

lindenb:
Hi all,
unless I'm wrong, there is no histogram for the mapping qualities in `samtools stats`. This small PR adds a histogram(mapq,count).

Thanks.

example:
```
./samtools stats HG02260.transloc.chr9.14.bam | grep MAPQ 
# Mapping qualities. Use `grep ^MAPQ | cut -f 2-` to extract this part. The columns are: mapq, count
MAPQ	0	3
MAPQ	7	1
MAPQ	25	2
MAPQ	29	25
MAPQ	37	42
MAPQ	60	398
MAPQ	70	6
```

### DIFF
--- a/doc/samtools-stats.1
+++ b/doc/samtools-stats.1
@@ -82,11 +82,11 @@ IS	Insert sizes
 RL	Read lengths
 FRL	Read lengths for first fragments only
 LRL	Read lengths for last fragments only
+MAPQ	Mapping qualities
 ID	Indel size distribution
 IC	Indels per cycle
 COV	Coverage (depth) distribution
 GCD	GC-depth
-MAPQ    Mapping qualities
 .TE
 
 Not all sections will be reported as some depend on the data being
@@ -336,6 +336,9 @@ observed length (up to the maximum specified by the \fB-l\fR option).
 Columns are read length and frequency.  FRL and LRL contains the same
 information separated into first and last fragments.
 
+MAPQ reports the mapping qualities for the mapped reads, ignoring the
+duplicates, supplementary, secondary and failing quality reads.
+
 ID reports the distribution of indel sizes, with one row per observed
 size. The columns are size, frequency of insertions at that size and
 frequency of deletions at that size.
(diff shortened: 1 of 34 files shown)


daviesrob:
You're right, there's no `MAPQ` histogram at the moment, but it does [count the number with zero mapping quality](https://github.com/samtools/samtools/blob/1c5fbedad5/stats.c#L969).  I note that `nreads_mq0` is only calculated on "primary" (i.e. non-secondary or supplementary) reads.  Would you also want to do that for your histogram?  In particular I'd have thought that secondary alignments would skew the 0 bin as it's likely that they'd all have mapping quality zero.

You're currently getting some [test failures](https://cirrus-ci.com/task/5292274044108800) because the `expected` test files haven't been updated to include the new section.  This can be easily fixed by running the tests and capturing the output somewhere:
```
make test 2>&1 | tee /tmp/test.out
```

Then search the output for `The outputs stdout differ`.  For each instance, compare the `expected` and `expected.new` files and if it looks OK copy the `.new` file over the original.  When they're all done you can `git add` them, commit and push.

The other missing element here is a manual page update, describing the new section in [`samtools-stats.1`](https://github.com/samtools/samtools/blob/1c5fbedad5/doc/samtools-stats.1).

lindenb:
@daviesrob thanks for the suggestions, I'll have a look at this in the next days.

lindenb:
I had the time to look back at this. I added a test to test the flags for `BAM_FUNMAP|BAM_FSECONDARY|BAM_FSUPPLEMENTARY|BAM_FQCFAIL|BAM_FDUP` , I updated the expected output files for tests and I updated the documentation.

I hope it's ok :-)

daviesrob:
Thanks for the update.  It looks like it's working as described.  I've taken the liberty of pushing a commit to make some small formatting adjustments:

* Fix man page table formatting (it needs a tab to work...)
* Move MAPQ in man page table and text to match its position in the output file.  All other entries follow this ordering.
* Adjust if-statement formatting to match (for better or worse) the lines near to it.

If you're happy with that I'll squash everything together and merge it.

lindenb:
@daviesrob many thanks for the little changes. It looks ok to me. 

:+1:

PR #1696, opened 2022-08-15
Fix unaligned access in dump_ref()

daviesrob:
Found building on SPARC, but it may cause slowdowns even on platforms that allow unaligned access.  Fix by putting in an
initial loop to get to an eight-byte boundary before switching to the fast mode.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -68,6 +68,9 @@ Bug Fixes:
  * Fix missing RG headers when using samtools merge -r.
    (PR#1683, addresses htslib#1479.  Reported by Alex Leonard)
 
+ * Fix a possible unaligned access in samtools reference.
+   (PR#1696)
+
 Documentation:
 
  * Add documentation on CRAM compression profiles and some of the newer options
--- a/reference.c
+++ b/reference.c
@@ -88,7 +88,10 @@ static int dump_ref(sam_hdr_t *h, hts_itr_t *iter, int ref_id,
     // (was ~20%).
     if (verbose) {
         int n4[8] = {0};
-        for (j = 0; j < (ref_len&~7); j+=8) {
+        for (j = 0; j < ref_len && (((uintptr_t) &ref[j] & 7) != 0); j++)
+            N += ref[j] == 'N';
+        uint64_t fast_end = ((ref_len - j) & ~7) + j;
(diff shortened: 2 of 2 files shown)


jkbonfield:
In assessing the speed, I wonder if it's even worth it.

We use this unrolling only because we insist on having O2 as our default optimisation instead of O3.  With O3 it gets vectorised and runs at 10x faster.  With O2 we have to manually do such shenanigans.  Clang vectorises at O2 so doesn't need such things.  Infact the int64 hackery to speed things up harms the vectorisation, so it's detrimental to clang anyway.

Maybe we should just use the naive loop:

```
        for (j=0; j < ref_len; j++)
            N += ref[j] == 'N';
```

Benchmarks from perf record for the dump_ref function.  "None" means `samtools reference -q` so no coverage stats.

```
dump_ref counts, gcc12, 3 runs
naive	2132 2125 2122
"-O3	 233  221  211
rob	 525  521  513
" -O3	 533  517  515
jkb	 413  413  407
" -O3	 406  397  411
none	  29   32   43

dump_ref counts, clang13
naive	234 245 251
rob	522 532 534
jkb 	345 349 360
none	 38  39	 47
```

We can see the refactoring of this code does slightly harm performance compared to the earlier one, but in all cases bar gcc -O2 the naive loop is faster still than either the new code or the current develop variant.

Overall impact on speed, with gcc12 -O2 `reference` is 7.3% slower than `reference -q` on my low coverage BAM file.  With the int64 hackery it's just 1.1% slower, so negligible.  With -O3 and a naive loop it's 0.9% slower, so also negligible.  Maybe just being naive and let people use -O3 if they care for performance is the right solution here.

jkbonfield:
Thinking again, that said the difference between a vectorised naive loop and a manual unrolled replacement is tiny in the grand scheme of things, so either is fine.  The difference between a default gcc build with the two methods is quite significant, so I guess maybe we do want to keep this until when/if we change the default optimisation levels for samtools builds.

So maybe we just keep as in this PR.  (The slow down vs previous code isn't worth worrying about as a portion of the overall CPU cost.)

daviesrob:
It should be fine to just squash the two commits together.

PR #1698, opened 2022-08-19
Add a bam_sanitize function to fixmate and view.

jkbonfield:
The original fixmate already did some basic sanitization, such as
checking for ref "*" or pos <=0 to indicate unmapped data, but it only
did so sporadically (e.g. not for secondary alignments).  It also
treated reads overhanging the end of the reference by a single base as
worthy as marking as unmapped, but it didn't clear things like CIGAR
or MAPQ.

This is more consistent and sane interface, which can turn reads
slightly overhanging the chromosome into soft-clipped alignments.

Samtools view uses the same code, but has the sanitizing disabled by
default.  (Note the -s option is taken on view but -z is free, so I
went with the Oxford spelling of sanitize.)

The rules are:

    RNAME *           => unmapped
    POS 0             => unmapped
    POS > ref len     => unmapped
    End pos > ref_len => soft-clip
    Unmapped          => MQUAL 0, CIGAR *
    Unmapped          => drop NM, MD, CG and SM tags.


jkbonfield:
Draft as this is for discussion, and other work probably needs to take place first.

The aux tag iterator in https://github.com/samtools/htslib/pull/1354 is an ideal fit for this.

Additionally if the concept is agreeable, then it's probably best if `bam_sanitize()` is part of htslib itself (which avoids the need to copy functions around, although that won't be such an issue once the iterator is merged).  That way any tool could just open the BAM file and sanitize as it goes.  Maybe automatically via an `hts_set_opt` request.

Comments welcomed.

Inspiration comes from https://gatk.broadinstitute.org/hc/en-us/articles/360040510231-CleanSam-Picard-

jkbonfield:
This is now working with @jmarshall's tag iterator.  The tag removal code here should be efficient too, as it's copying tags back over the same buffer in a single pass (with no O(N^2) issues based on number or length of tags), although I haven't done any benchmarks to see what cost that has.

It's still draft however as we need to make a decision on where `bam_sanitize` actually belongs.  Choices are.

1. Here.  We can always move it later if desired.

2. In Htslib.  Either as an explicitly callable function as is written here, or as an automatic ability enabled by e.g. `hts_set_opt(fp, HTS_OPT_SANITIZE_STR, "mqual,cigar")`. (Or `"all,-pos"` could be everything bar the POS modifications that break sort.)

jkbonfield:
Undrafting as decision at weekly meeting was to put into Samtools for now, and if we deem it to be useful we can migrate the function to htslib at a later date.

Issue #1143, opened 2019-11-26
Library level stats from Markdup


PedalheadPHX:
#### Is your feature request related to a problem? Please specify.

Not a problem but a limitation of the current stats summary when a given BAM file contains multiple libraries.

#### Describe the solution you would like.

Assuming duplicates are marked by library (RG-LB) within a given BAM, and optical duplicates are marked by the platform unit (RG-PU) then having at least the duplicate stats for each library defined in the BAM versus the sample or even multi-sample BAM would improve quality control processes.

Requested solution: `samtools markdup -f markdup.stats in.bam out.bam` produces a file with one column or row for each library within "in.bam".  I acknowledge the optical duplicate stats are then left in a similar issue where the numbers might differ by RG-PU but unlike different libraries that might have very different duplicate levels I'd expect the same library sequenced on multiple RG-PU will have very similar optical duplicate numbers so averaging the level observed by library seems reasonable.

In a perfect world the stats would be output in JSON with levels for the total BAM, each RG-SM (might be multiple samples in a BAM), and each RG-LB (might be multiple libraries).

whitwham:
It would take some reworking of the stats but that might not be a bad thing.

As for JSON, how about posting an example of the kind of output you would like to see?

PR #1699, opened 2022-08-22
Add an option to mark duplicates by read group.

whitwham:
This adds an option to take the read group into account when marking duplicates.

The stats have been changed to display duplicates by read groups after the totals.  There is also an option to output stats in JSON format.


whitwham:
Even though they are valid JSON, I'm not entirely convinced the JSON stats are usefully formatted

daviesrob:
Rebased and tidied up some odd strings that got into one of the commits.  Apologies for the forced push.

bam_markdup.c
```diff
@@ -265,7 +289,7 @@ static int64_t calc_score(bam1_t *b)
    read is leftmost of the pair. */
 
 
-static int make_pair_key(md_param_t *param, key_data_t *key, bam1_t *bam, long *warnings) {
+static int make_pair_key(md_param_t *param, key_data_t *key, bam1_t *bam, uint8_t rg_num, long *warnings) {
```
daviesrob, line 292: Why is this `uint8_t` and not `int`, as used elsewhere?  (Also in `make_single_key()`.)

whitwham: I don't remember.  I think it was part of an earlier development phase.  I changed how it worked part way through.  I'll change it.

daviesrob:
Thanks for the updates.  I've squashed in the fixes, and merged.

whitwham:
Hopefully should have fixed the requests in #1143.

PR #1701, opened 2022-08-24
Improve the speed of samtools sort

jkbonfield:
This is predominantly Rob Davies work from 3 years ago, but based on joint ideas from assessments both then and during the last week.

Sorting large files consists of a first phase of reading blocks of data, sorting them, and spilling to disk, followed by a second phase to merge the temporary files back together again.  This can lead to many files and consequentially lots of disk thrashing.  The best way to handle this is using `-m` to specify large memory usage, which reduces the number of files.  Another improvement is to increase the size of the blocks being read from the default 32KB used by htslib's hfile to 1MB.

Unfortunately the memory is per thread, so if we go from 1 thread to 16 threads then we need to reduce the block size by 16 to compensate.  That gives us 16 times more temporary files and a 16 fold bigger disk thrashing experience.  This PR merges the parallel sorted data together before writing to disk, dramatically reducing this.  It makes the sorting phase slower, but the merge stage considerably faster.

It's not perfect and there are more things that can be done (see below), but it's already a 3 fold speed increase for 16 threads on huge files (see commit message for lustre results).

Potential future work.

- The sort bit is parallel, but the merge step is in the main thread.  This is weak and causes the main thread to become the bottleneck so the overall parallelism is still poor.  We need a parallel merge (eg a naive recursive one) to remove this bottleneck.

- Even without being parallel, the merge is based on `ks_heapsort` which is quite poor performing.  A naive implementation just tracking the next N bam records was about 10% faster for `samtools merge`.  There is potential to speed that up further using a tree instead of a brute force search for the next lowest record.

- There are distinct phases of I/O bound and CPU bound during sort.  Ideally the I/O bit would be asynchronous and overlapping the CPU stage.  Using the thread-pool would work here, but it still needs a parallel merge too.

- Htslib: the `fdatasync` in hfile.c can be a slow down at times.  We should have a way to disable it as we don't need it for temporary files.  (It doesn't matter if the data is still in the disk cache provided we can process it and it ends correctly.)

- Htslib: the `qsize = nthreads*2` equation in bgzf blocks isn't optimal. `*10` gave better performance (about 10% when combined with the fsync change), but uses more memory.  Given the I/O in bgzf is asynchronous, boosting the queue size gives more potential for running I/O and CPU stages in parallel.  We need a better API for controlling this so we can increase it where helpful, without incurring higher memory costs across the board.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -160,30 +160,36 @@ typedef enum {Coordinate, QueryName, TagCoordinate, TagQueryName, MinHash, Templ
 static SamOrder g_sam_order = Coordinate;
 static char g_sort_tag[2] = {0,0};
 
-#ifdef isdigit
-#    undef isdigit
-#endif
-#define isdigit(c) ((c)<='9' && (c)>='0')
-
+#define is_digit(c) ((c)<='9' && (c)>='0')
 static int strnum_cmp(const char *_a, const char *_b)
 {
     const unsigned char *a = (const unsigned char*)_a, *b = (const unsigned char*)_b;
     const unsigned char *pa = a, *pb = b;
     while (*pa && *pb) {
-        if (isdigit(*pa) && isdigit(*pb)) {
+        if (!is_digit(*pa) || !is_digit(*pb)) {
+            if (*pa != *pb)
+                return (int)*pa - (int)*pb;
+            ++pa; ++pb;
+        } else {
+            // skip leading zeros
             while (*pa == '0') ++pa;
             while (*pb == '0') ++pb;
-            while (isdigit(*pa) && isdigit(*pb) && *pa == *pb) ++pa, ++pb;
-            if (isdigit(*pa) && isdigit(*pb)) {
-                int i = 0;
-                while (isdigit(pa[i]) && isdigit(pb[i])) ++i;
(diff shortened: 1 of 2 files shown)


bam_sort.c
```diff
@@ -159,6 +160,11 @@ typedef enum {Coordinate, QueryName, TagCoordinate, TagQueryName, MinHash, Templ
 static SamOrder g_sam_order = Coordinate;
 static char g_sort_tag[2] = {0,0};
 
+#ifdef isdigit
+#    undef isdigit
+#endif
+#define isdigit(c) ((c)<='9' && (c)>='0')
```
daviesrob, line 166: Overriding a libc function like this looks dodgy.  I think we should call the macro something like `is_digit`, replace `isdigit` with `is_digit` in `strnum_cmp()`, and remove `#include <ctype.h>` because it's no longer needed.

jkbonfield: Fair comment, although looking at the entire function I see it's unnecessarily convoluted anyway.

I have a rewrite which is considerably quicker in gcc, and no worse in clang which was already better than gcc (entire sort is 20% less CPU, including all the BAM/SAM parsing and I/O).

PR #1705, opened 2022-08-25
  bug fix: samtools merge --template-coordinate

nh13:
This includes the commit for the fix in #1703, so this should be merged only after that has been merged.

The keys used to compare based on template coordinate were not being created as the input sorted BAMs were being merged.

Tested via merging large sorted files that were created using the fix in #1703.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1686,7 +1686,7 @@ static inline int heap_add_read(heap1_t *heap, int nfiles, samFile **fp,
         if (in_mem[i - nfiles].from < in_mem[i - nfiles].to) {
             size_t from = in_mem[i - nfiles].from;
             heap->entry.bam_record = buf[from].bam_record;
-            if (g_sam_order == TemplateCoordinate) heap->entry.u.key = template_coordinate_keys_get(keys, from);
+            if (g_sam_order == TemplateCoordinate) heap->entry.u.key = buf[from].u.key;
             in_mem[i - nfiles].from++;
             res = 0;
         } else {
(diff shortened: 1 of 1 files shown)


bam_sort.c
```diff
@@ -1294,6 +1300,26 @@ int bam_merge_core2(SamOrder sam_order, char* sort_tag, const char *out, const c
...
+        }
+        keys->n = 0;
+        keys->m = 0;
+        keys->buffer_size = 0x10000;
+        keys->buffers = NULL;
+        // Make sure that there's enough memory for template coordinate keys, one per file to read
+        if (keys->n + n >= keys->m * keys->buffer_size) {
+            if (template_coordinate_keys_realloc(keys, keys->n + n) < 0) goto mem_fail;
+        }
+        lib_lookup = lookup_libraries(hout);
```
nh13, line 1317: I am unsure of which header to use, so I picked the output one. I _think_ that's correct, but it could use a second set of eyes.

bam_sort.c
```diff
@@ -1686,7 +1728,7 @@ static inline int heap_add_read(heap1_t *heap, int nfiles, samFile **fp,
         if (in_mem[i - nfiles].from < in_mem[i - nfiles].to) {
             size_t from = in_mem[i - nfiles].from;
             heap->entry.bam_record = buf[from].bam_record;
-            if (g_sam_order == TemplateCoordinate) heap->entry.u.key = template_coordinate_keys_get(keys, from);
```
nh13, line 1689: changes here and below are from #1703

bam_sort.c
```diff
@@ -1379,6 +1410,10 @@ int bam_merge_core2(SamOrder sam_order, char* sort_tag, const char *out, const c
             heap->idx = idx++;
             if (g_sam_order == TagQueryName || g_sam_order == TagCoordinate) {
                 heap->entry.u.tag = bam_aux_get(heap->entry.bam_record, g_sort_tag);
+            } else if (g_sam_order == TemplateCoordinate) {
+                template_coordinate_key_t *key = template_coordinate_keys_get(keys, heap->i); // get the next key to use
```
nh13, line 1414: I think `heap->i` is the correct index, but could use another set of eyes.

PR #1706, opened 2022-08-26
Improve merge stage of samtools sort

daviesrob:
Makes the merge stage make better use of the thread pool:
- Writing now uses the same pool as reading instead of starting its own
- Reading temporary files is now threaded.

The second change has the side effect of starting a reader thread for each temporary file.  To limit the impact of this, the spilling code is modified so that it limits the number of temporary files present by merging them if they're getting to numerous.  While this has a small impact on speed, it's only if a very large number of temporary files is created and it has extra benefits like fixing the problem of running out of file descriptors when trying to read a huge number of files back in.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -48,6 +48,7 @@ DEALINGS IN THE SOFTWARE.  */
 #include "htslib/sam.h"
 #include "htslib/hts_endian.h"
 #include "htslib/cram.h"
+#include "htslib/thread_pool.h"
 #include "sam_opts.h"
 #include "samtools.h"
 #include "bedidx.h"
@@ -1735,7 +1736,8 @@ static int bam_merge_simple(SamOrder sam_order, char *sort_tag, const char *out,
                             int n, char * const *fn, int num_in_mem,
                             buf_region *in_mem, bam1_tag *buf,
                             template_coordinate_keys_t *keys,
-                            khash_t(const_c2c) *lib_lookup, int n_threads,
+                            khash_t(const_c2c) *lib_lookup,
+                            htsThreadPool *htspool,
                             const char *cmd, const htsFormat *in_fmt,
                             const htsFormat *out_fmt, char *arg_list, int no_pg,
                             int write_index) {
@@ -1817,7 +1819,8 @@ static int bam_merge_simple(SamOrder sam_order, char *sort_tag, const char *out,
         return -1;
     }
 
-    if (n_threads > 1) hts_set_threads(fpout, n_threads);
+    if (htspool->pool)
+        hts_set_opt(fpout, HTS_OPT_THREAD_POOL, htspool);
 
     if (sam_hdr_write(fpout, hout) != 0) {
         print_error_errno(cmd, "failed to write header to \"%s\"", out);
@@ -2690,6 +2693,7 @@ int bam_sort_core_ext(SamOrder sam_order, char* sort_tag, int minimiser_kmer,
     const char *new_ss = NULL;
     buf_region *in_mem = NULL;
     khash_t(const_c2c) *lib_lookup = NULL;
+    htsThreadPool htspool = { NULL, 0 };
     int num_in_mem = 0;
     int large_pos = 0;
 
@@ -2832,11 +2836,14 @@ int bam_sort_core_ext(SamOrder sam_order, char* sort_tag, int minimiser_kmer,
         }
     }
 
-    // No gain to using the thread pool here as the flow of this code
-    // is such that we are *either* reading *or* sorting.  Hence a shared
-    // pool makes no real difference except to reduce the thread count a little.
-    if (n_threads > 1)
-        hts_set_threads(fp, n_threads);
+    if (n_threads > 1) {
+        htspool.pool = hts_tpool_init(n_threads);
+        if (!htspool.pool) {
+            print_error_errno("sort", "failed to set up thread pool");
+            goto err;
+        }
+        hts_set_opt(fp, HTS_OPT_THREAD_POOL, &htspool);
+    }
 
     if ((bam_mem = malloc(max_mem)) == NULL) {
         print_error("sort", "couldn't allocate memory for bam_mem");
@@ -2926,7 +2933,7 @@ int bam_sort_core_ext(SamOrder sam_order, char* sort_tag, int minimiser_kmer,
                 if (bam_merge_simple(g_sam_order, sort_by_tag, fns[n_files],
                                      large_pos ? "wzx1" : "wbx1", header,
                                      0, NULL, n_threads, in_mem, buf, keys,
-                                     lib_lookup, n_threads, "sort", NULL, NULL,
+                                     lib_lookup, &htspool, "sort", NULL, NULL,
                                      NULL, 1, 0) >= 0) {
                     merge_res = 0;
                     break;
@@ -2985,7 +2992,7 @@ int bam_sort_core_ext(SamOrder sam_order, char* sort_tag, int minimiser_kmer,
         char *sort_by_tag = (sam_order == TagQueryName || sam_order == TagCoordinate) ? sort_tag : NULL;
         if (bam_merge_simple(sam_order, sort_by_tag, fnout, modeout, header,
                              n_files, fns, num_in_mem, in_mem, buf, keys,
-                             lib_lookup, n_threads, "sort", in_fmt, out_fmt,
+                             lib_lookup, &htspool, "sort", in_fmt, out_fmt,
                              arg_list, no_pg, write_index) < 0) {
             // Propagate bam_merge_simple() failure; it has already emitted a
             // message explaining the failure, so no further message is needed.
@@ -3020,6 +3027,9 @@ int bam_sort_core_ext(SamOrder sam_order, char* sort_tag, int minimiser_kmer,
     lib_lookup_destroy(lib_lookup);
     sam_hdr_destroy(header);
     if (fp) sam_close(fp);
+    if (htspool.pool)
+        hts_tpool_destroy(htspool.pool);
+
     return ret;
 }
 
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -1775,6 +1775,8 @@ static int bam_merge_simple(SamOrder sam_order, char *sort_tag, const char *out,
                 goto fail;
             }
             hts_set_opt(fp[i], HTS_OPT_BLOCK_SIZE, BAM_BLOCK_SIZE);
+            if (htspool->pool)
+                hts_set_opt(fp[i], HTS_OPT_THREAD_POOL, htspool);
 
             // Read header ...
             hin = sam_hdr_read(fp[i]);
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -54,7 +54,8 @@ DEALINGS IN THE SOFTWARE.  */
 #include "bedidx.h"
 #include "bam.h"
 
-#define BAM_BLOCK_SIZE 1024*1024
+#define BAM_BLOCK_SIZE 2*1024*1024
+#define MAX_TMP_FILES 64
 
 // Struct which contains the sorting key for TemplateCoordinate sort.
 typedef struct {
@@ -2680,7 +2681,7 @@ int bam_sort_core_ext(SamOrder sam_order, char* sort_tag, int minimiser_kmer,
                       const htsFormat *in_fmt, const htsFormat *out_fmt,
                       char *arg_list, int no_pg, int write_index)
 {
-    int ret = -1, res, i, nref, n_files = 0;
+    int ret = -1, res, i, nref, n_files = 0, n_big_files = 0, fn_counter = 0;
     size_t max_k, k, max_mem, bam_mem_offset;
     sam_hdr_t *header = NULL;
     samFile *fp = NULL;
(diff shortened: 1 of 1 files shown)


jkbonfield:
I haven't assessed the code yet, just the effect.  Some stats from running it on small and big files.

Firstly, a whopper 158GB name sorted file, sorted to position:

```
-rw-r--r-- 2 jkb team117 168607956340 Aug 22 12:56 NA12878_S1.name.bam

@ [lustre/NA12878]; time ./samtools.dev sort --no-PG -@16 -m1G NA12878_S1.name.bam -o _dev.bam
[bam_sort_core] merging from 31 files and 16 in-memory blocks...
real    100m21.610s
user    349m23.107s
sys     24m13.316s

@ [lustre/NA12878]1; time ./samtools.PR1706 sort --no-PG  -@16 -m1G NA12878_S1.name.bam -o _PR1706.bam
[bam_sort_core] merging from 31 files and 16 in-memory blocks...
real    67m54.740s
user    394m44.315s
sys     27m40.874s
```

That's using 16GB ram across 16 threads.  Approx 13% more cpu, but 32% less real time.  Seems worth the trade off.  I'm somewhat surprised my copy of devel branch still only produced 31 temp files.

On a smaller (~6GB) file and using a deliberately low 10MB memory per thread to stress test lots of I/O (but note not stressing enough likely as it would have fitted in the machine buffer cache):

```
@ [lustre/NA12878]130; time ./samtools.dev sort --no-PG -n -@16 -m10M ~/scratch/data/9827_2#49.bam -o _dev.bam
[bam_sort_core] merging from 113 files and 16 in-memory blocks...
real    4m52.329s
user    17m28.418s
sys     0m42.740s

@ [lustre/NA12878]; time ./samtools.PR1706 sort --no-PG -n  -@16 -m10M ~/scratch/data/9827_2#49.bam -o _PR1706.bam
[bam_sort_core] merging from 17 files and 16 in-memory blocks...
real    3m41.882s
user    22m11.895s
sys     0m48.134s
```

Again it's an increase in CPU due to the multiple merge steps, but another 24% reduction in real time.  I'd expect that to increase substantially if running on slower disk and with larger files where buffer cache doesn't rescue us.  Given the small input, most of the saving in the above test will be coming from the asynchronous temp file decoding, rather than the merge to avoid disk thrashing.

I saw similar drops when doing pos -> name sorting.

jkbonfield:
It has a certain amount of O(N^2) complexity going on once we start repeatedly triggering the biggest merge I think.   We merge 64 files down to 1, and then start coalescing 1..N, until we have 1..65 which we then need to merge to a big file again.  It doesn't then have 2 huge files (the original one plus the next 64 combined), but merges back into the original.

This is really contrived however and I could only trigger this by a rediculously small amount of memory requested (1MB per thread on a 16GB (when uncompressed) file, giving 1138 temp files total in the current dev code).  Even then it came out the same wall clock time as before, although about 2x CPU, so not catastrophic.  In realistic data even on huge files I think you'd struggle to ever get more than a couple passes, making the complexity bound to something low.

So I'm OK with this as appropriate usage gives a clear benefit to performance on very large files (which is where it matters most).

ghuls:
Was this patch included in SAMtools 1.16.1?

Because the following command created 650 temporary files of 512MB.
If you have samtools sort e.g. 2G memory, shouldn't the chunks be 2G instead of 512M?

Also in case of sorting output from a mapping process, it would be useful to have a different number of threads for the initial sorting and the final merging step. During the mapping, sorting with 2 threads keeps up easily, but when the mapping finishes, merging all the 650 temp files takes ages with those 2 threads.


```bash
bwa-mem2 \
                -t 16 \
                -R "@RG\tID:${fastq_basename_prefix}\tSM:${short_bc_id}\tLB:${library_id}\tPL:ILLUMINA" \
                "${bwa_mem2_index}" \
                "${fastq_PE1}" \
                "${fastq_PE2}" \
  | samtools fixmate -u -m -O bam - - \
  | samtools sort \
                -@ 2 \
                -m 2G \
                -O bam \
                --write-index \
                -T "/tmp/${fastq_basename_prefix}.bwa.out.fixmate.possorted.TMP" \
                -o "${fastq_basename_prefix}.bwa.out.fixmate.possorted.bam##idx##${fastq_basename_prefix}.bwa.out.fixmate.possorted.bam.bai" \
                -
```

daviesrob:
No, 1.16.1 only included small bug fixes.  This PR will be included in the next release, which should be out soon.  Meanwhile you can get the upgraded sort by building the `develop` branch of samtools.

The temporary files will be smaller than the sort memory because they're compressed using zlib's quickest setting on writing.  This might not be so useful if you're in the happy position of owning a massive NVME drive, but should be beneficial for network filesystems and spinning-rust hard drives.  It also helps keep the temporary storage needed down to a reasonable level - otherwise you could get quite an expansion when, say, sorting a big cram file.

jkbonfield:
> No, 1.16.1 only included small bug fixes. This PR will be included in the next release, which should be out soon. Meanwhile you can get the upgraded sort by building the `develop` branch of samtools.

A small reminder that this also needs the develop branch of htslib.

Also I assume you mean libdeflate's quickest setting, which will be vastly better than zlib! :)  On my test file, libdeflate's level 1 is comparable in size to level 3-4 of zlib while being >2x quicker than zlib -1.

Did we ever benchmark the speed/size of using CRAM level1 for temporary files? In my admittedly small noddy test, it looks like `-O CRAM,level=1,no_ref` is approx 1/3rd smaller than the BAM,level=1 while being comparable on write CPU (probably slower on decode, although I/O may alter that).

ghuls:
@jkbonfield By any chance did you test the lastest libdeflate (>= 1.14) and saw more speedup compared with older versions)?

Should setting `-m 2G` not create temp files of 2G, or are the temp files always around 512MB?

jkbonfield:
That was my old git clone at 1.9 (which was a significant speed and size gain over 1.8).  With 1.14 it looks to be about 6% faster and no size difference, albeit on a small test.   At higher compression levels there's almost no change in speed.

The memory size is uncompressed in-memory representation, which is very similar to BAM, plus some extra allowance for additional data structures.  It's possible it may actually have memory for both sorted and unsorted copies while it's working, so that may impact the mem vs on-disk sizes.  @daviesrob could explain more.

jkbonfield:
Some benchmarks of 1 million NovaSeq record uncompressed BAM.

```
@ seq4-os0000002[ftp/compress...]; for i in libdeflate/programs/libdeflate-gzip 
zlib-1.2.13/minigzip zlib-cloudflare/minigzip zlib-intel/minigzip zlib-ng/minigzip;do echo === $i -1 ===;time $i -1 < /tmp/_ | wc -c;done
=== libdeflate/programs/libdeflate-gzip -1 ===
57597335

real	0m1.458s
user	0m1.366s
sys	0m0.092s
=== zlib-1.2.13/minigzip -1 ===
67646970

real	0m3.801s
user	0m3.676s
sys	0m0.137s
=== zlib-cloudflare/minigzip -1 ===
66406017

real	0m1.888s
user	0m1.772s
sys	0m0.144s
=== zlib-intel/minigzip -1 ===
86019933

real	0m1.160s
user	0m1.034s
sys	0m0.178s
=== zlib-ng/minigzip -1 ===
81201451

real	0m1.251s
user	0m1.096s
sys	0m0.150s
```

The Intel zlib is and older binary because the latest git pull wouldn't build for me. :/

You can see the Intel zlib and zlib-ng have a weaker level 1 which is fastest, but considerably larger.   However at level 2 they're both slower and larger than libdeflate's level 1:

```
@ seq4-os0000002[ftp/compress...]130; for i in libdeflate/programs/libdeflate-gzip zlib-1.2.13/minigzip zlib-cloudflare/minigzip zlib-intel/minigzip zlib-ng/minigzip;do echo === $i -2 ===;time $i -2 < /tmp/_ | wc -c;done
=== libdeflate/programs/libdeflate-gzip -2 ===
55810479

real	0m1.812s
user	0m1.739s
sys	0m0.087s
=== zlib-1.2.13/minigzip -2 ===
64852587

real	0m4.137s
user	0m4.061s
sys	0m0.114s
=== zlib-cloudflare/minigzip -2 ===
63568761

real	0m2.063s
user	0m1.934s
sys	0m0.156s
=== zlib-intel/minigzip -2 ===
66459719

real	0m2.315s
user	0m2.187s
sys	0m0.160s
=== zlib-ng/minigzip -2 ===
62909542

real	0m2.061s
user	0m1.905s
sys	0m0.166s
```


At level 5, libdeflate is still pulling ahead of the others:

```
=== libdeflate/programs/libdeflate-gzip -5 ===
52431629

real	0m2.756s
user	0m2.691s
sys	0m0.065s
=== zlib-1.2.13/minigzip -5 ===
52703671

real	0m7.309s
user	0m7.183s
sys	0m0.154s
=== zlib-cloudflare/minigzip -5 ===
52417282

real	0m3.482s
user	0m3.392s
sys	0m0.105s
=== zlib-intel/minigzip -5 ===
52790581

real	0m4.058s
user	0m3.937s
sys	0m0.156s
=== zlib-ng/minigzip -5 ===
51780341

real	0m3.545s
user	0m3.435s
sys	0m0.111s
```

Finally at -9
```
=== libdeflate/programs/libdeflate-gzip -9 ===
49503896

real	0m16.432s
user	0m16.361s
sys	0m0.059s
=== zlib-1.2.13/minigzip -9 ===
49264027

real	1m12.072s
user	1m11.880s
sys	0m0.202s
=== zlib-cloudflare/minigzip -9 ===
49498921

real	0m39.761s
user	0m39.618s
sys	0m0.182s
=== zlib-intel/minigzip -9 ===
49263774

real	1m1.196s
user	1m1.059s
sys	0m0.173s
=== zlib-ng/minigzip -9 ===
49078341

real	0m28.759s
user	0m28.569s
sys	0m0.173s
```

Zlib-ng is looking like it's coming on well and easily the best at compression ratio/speed vs the others, on this uBAM anyway.  I do wonder why distributions normally ship the vanilla zlib instead of zlib-ng.

Libdeflate doesn't have the same interface and it's not streaming, but for us it's perfect as we're doing block-based compression.  Indeed the lack of a streaming interface simplifies things as it's a trivial API (unlike Zlib's). 

Note libdeflate goes up to -12, but it gets much slower.  However at -10 it's still faster than vanilla zlib and much smaller:

```
@ seq4-os0000002[ftp/compress...]; time libdeflate/programs/libdeflate-gzip -10 < /tmp/_ | wc -c
47934445

real	0m52.836s
user	0m52.719s
sys	0m0.095s
```

ghuls:
If you want fastq streaming decompression (when the input is not blockcompressed and can't use libdeflate), igzip is a clear winner.

https://github.com/zlib-ng/zlib-ng/issues/986

Also in my experiments `bgzip -@ 4` is slower (libdeflate seems to be linked in the binary) than `pigz -t 4` when pigz uses zlib-ng.

jkbonfield:
It's Friday afternoon and you've got me on my favourite topic.  So... :-)

Other than level 1 for NG and Intel, there's just no contest vs libdeflate.

I also added ZStd just out of curiosity.  Note all of these tests were just compressing the entire file (~300Mb).  This is very different than the usual BAM usage which has random access and relatively small blocks for random access.  With random access and a comparable block size it'd be much closer to the other tools.  The same obviously applies to the other tools too.  Unless I wrote a dedicated block-based engine to validate things on I can't say how this would compare on a real BAM.

image

jkbonfield:
It's possible to build bgzip without libdeflate, and then with appropriate compiler flags you could get it to link against zlib-ng instead of vanilla zlib to evaluate performance.

ghuls:
I think I found the reason why pigz is faster for me.

libdeflate is indeed fast (as I knew already), but single threaded.
pigz with multiple threads (e.g 4) using zlib-ng, beats libdeflate for similar compressed filesize when looking at total wall time.
pigz with multiple threads (e.g.4) using zlib-ng beats bgzip with multiple threads using libdeflate. pigz somehow manages to keep each compression thread busy at 100% CPU usage, while for bgzip the CPU usage per compression thread is only 80% for 4 threads and 70% when using 8 threads.

```bash
# Best compression level of igzip (ISA-L).
❯  time igzip -3 -k -c test.fastq > test.igzip_l3.gz                                                                                                                                             
                                                                                                                                                                                       
real    0m36.028s                                                                                                                                                                      
user    0m32.820s                                                                                                                                                                      
sys     0m3.108s 

# libdeflate gzip 1.15 (single threaded).
❯  time libdeflate-gzip -k -c test.fastq > test.deflate.fastq.gz                                                                                                                       
                                                                                                                                                                                       
real    2m57.240s                                                                                                                                                                      
user    2m52.727s                                                                                                                                                                      
sys     0m4.098s   


# libdeflate gzip 1.15 (single threaded), compression level 1.
❯  time libdeflate-gzip -k -1 -c test.fastq > test.deflate1.fastq.gz

real    0m53.526s
user    0m49.259s
sys     0m4.139s


# bgzip with 8 threads using libdeflate 1.15. Each compression thread only uses around 70% CPU.
❯  time bgzip -@ 8 -k -c test.fastq > test.bgzip.fastq.gz                                                                                                                              
                                                                                                                                                                                       
real    0m57.516s                                                                                                                                                                      
user    5m20.633s                                                                                                                                                                      
sys     0m7.306s                                                                                                                                                                       
                                                                                                                                                                                       

# pigz with 8 threads using zlib-ng 2.0.6. Each compression thread only uses around 100% CPU.                               
❯  time pigz -p 8 -k -c test.fastq > test.pigz.fastq.gz                                                                                                                                
                                                                                                                                                                                       
real    0m33.319s                                                                                                                                                                      
user    4m24.218s                                                                                                                                                                      
sys     0m7.926s  

# bgzip with 4 threads using libdeflate 1.15. Each compression thread only uses around 80% CPU.
❯  time bgzip -@ 4 -k -c test.fastq > test.bgzip_t4.fastq.gz

real    1m33.813s
user    5m0.040s
sys     0m6.539s

# pigz with 4 threads using zlib-ng 2.0.6. Each compression thread only uses around 100% CPU.     
time pigz -p 4 -k -c test.fastq > test.pigz_t4.fastq.gz                                                                                                                               
                                                                                                                                                                                       
real    1m6.554s                                                                                                                                                                       
user    4m24.364s                                                                                                                                                                      
sys     0m9.129s


# pigz with 4 threads using zlib 1.2.13. Each compression thread only uses around 100% CPU.     
❯  time pigz -p 4 -k -c test.fastq > test.pigz_zlib_1.2_13.fastq.gz

real    4m3.101s
user    15m54.497s
sys     0m9.710s
```

File sizes are very similar after compression (pigz files are slightly smaller as they can compress the whole input vs only per block).
```
1683888451  test.bgzip.fastq.gz
1683888451  test.bgzip_t4.fastq.gz
1682775204  test.deflate.fastq.gz
1916988501  test.deflate_l1.fastq.gz
1833513871  test.deflate_l3.fastq.gz
9519954727  test.fastq
1990408694  test.igzip_l3.gz
1674850975  test.pigz.fastq.gz
1674851440  test.pigz_t4.fastq.gz
1711772278  test.pigz4_independend.fastq.gz
1678911156  test.pigz_zlib_1.2_13.fastq.gz

```

PR #1721, opened 2022-09-20
Work around PacBio HiFi quality induced biases.

jkbonfield:
With a homopoymer, eg in ATTTTC, one base in the run is likely given a low confidence value.  This is always the same end.  After reverse complementing during alignment, we get a mix of low qualities at the first and last base.  However the alignment algorithms typically do left-justification of gaps, so any slight strand bias will be exacerbated as high vs low confidence calls for that indel.

This method averages the first and last qualities together to get an amortised new quality.  It then averages the end two inner qualities together, and so on.

This has been seen to reduce indel consensus call errors considerably in some PB HiFi data sets.

### DIFF
--- a/bam_consensus.c
+++ b/bam_consensus.c
@@ -99,6 +99,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 // but 30T+ 20T- 18A+ 2A- seems like a consistent A miscall on one strand
 // only, while T is spread evenly across both strands.
 
+// TODO: Find lowest quality within STR or homopolymer if indel
+// So T insert to left end with last T being low qual, means low ins score
+
 #include <config.h>
 
 #include <stdio.h>
@@ -170,6 +173,7 @@ typedef struct {
     int incl_flags;
     int min_mqual;
     double P_het;
+    int homopoly_fix;
 
     // Internal state
     samFile *fp;
@@ -380,6 +384,49 @@ int nins(const bam1_t *b){
     return indel;
 }
 
+/*
+ * Some machines, including 454 and PacBio, store the quality values in
+ * homopolymers with the first or last base always being the low quality
+ * state.  This can cause problems when reverse-complementing and aligning,
+ * especially when we left-justify indels.
+ *
+ * Other platforms take the approach of having the middle bases high and
+ * the low confidence spread evenly to both start and end.  This means
+ * reverse-complementing doesn't introduce any strand bias.
+ *
+ * We redistribute qualities within homopolymers in this style to fix
+ * naive consensus or variant calling algorithms.
+ */
+void homopoly_qual_fix(bam1_t *b) {
+    static double ph2err[256] = {0};
+    if (!ph2err[0]) {
+        for (int i = 0; i < 256; i++)
+            ph2err[i] = pow(10, i/-10.0);
+    }
+    uint8_t *seq = bam_get_seq(b);
+    uint8_t *qual = bam_get_qual(b);
+    for (int i = 0; i < b->core.l_qseq; i++) {
+        int s = i; // start of homopoly
+        int base = bam_seqi(seq, i);
+        while (i+1 < b->core.l_qseq && bam_seqi(seq, i+1) == base)
+            i++;
+        // s..i inclusive is now homopolymer
+
+        if (s == i)
+            continue;
+
+        // Simplest:  reverse if end_qual < start_qual
+        // Next:      average outer-most two, then next two, etc
+        // Best:      fully redistribute so start/end lower qual than centre
+
+        // Middle route of averaging outer pairs is sufficient?
+        for (int j = s, k = i; j < k; j++,k--) {
+            double e = ph2err[qual[j]] + ph2err[qual[k]];
+            qual[j] = qual[k] = -fast_log2(e/2)*3.0104+.49;
+        }
+    }
+}
+
 // Return the local NM figure within halo (+/- HALO) of pos.
 // This local NM is used as a way to modify MAPQ to get a localised MAPQ
 // score via an adhoc fashion.
@@ -491,6 +538,11 @@ int nm_init(void *client_data, samFile *fp, sam_hdr_t *h, pileup_t *p) {
         }
 #endif
     }
+
+    // Fix e.g. PacBio homopolymer qualities
+    if (opts->homopoly_fix)
+        homopoly_qual_fix((bam1_t *)b);
+
 
     // Adjust local_nm array by the number of edits within
     // a defined region (pos +/- halo).
@@ -1405,6 +1457,7 @@ static void usage_exit(FILE *fp, int exit_status) {
     fprintf(fp, "      --high-MQ INT     Cap maximum mapping quality [60]\n");
     fprintf(fp, "      --P-het FLOAT     Probability of heterozygous site[%.1e]\n",
             P_HET);
+    fprintf(fp, "  -p, --homopoly-fix    Spread low-qual bases to both ends of homopolymers\n");
 
     fprintf(fp, "\nGlobal options:\n");
     sam_global_opt_help(fp, "-.---@-.");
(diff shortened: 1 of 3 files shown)


jkbonfield:
I did a further fix, which I haven't validated on Illumina yet (will do so, but unlikely to make any diff).

On PacBio it's easy enough to compare vs GIAB HG002 truth set and it's definitely better.  I need to dig out my consensus to VCF conversion script.  I know I wrote a noddy one somewhere.

jkbonfield:
Some stats of consensus on PB CCS data vs 1.16, also with the additional `-p` option:

```
PacBio CCS kHG002.GRCh38.PacBio_CCS_15Kb.bam

1-77M consensus116:
SNP          Q>0 /   Q>=30d
SNP   TP   85553 /   85262
SNP   FP    1160 /     932
SNP   GT     294 /     253
SNP   FN      88 /     379

InDel TP   14565 /   13922
InDel FP   35602 /   35156
InDel GT    2537 /    2457
InDel FN     404 /    1047

1-77M consensus: (min vs avg qual for gaps)
SNP          Q>0 /   Q>=30
SNP   TP   85559 /   85293
SNP   FP    1157 /     929 -3
SNP   GT     235 /     195 -58
SNP   FN      82 /     348 -31

InDel TP   14590 /   13943
InDel FP   22431 /   21982 -13174 !!
InDel GT    2352 /    2261 -196
InDel FN     379 /    1026 -21
```

Huge gains here.  This is due to large swings in confidence values with PBCCS vs Illumina which is relatively flat.  New model for gap confidence is obviously important.  Hence getting the confidence values for gaps correct is a big win.

```
1-77M consensus -p: (redistribution of quals in homopolymers)
SNP          Q>0 /   Q>=30
SNP   TP   85465 /   85227
SNP   FP    1128 /     891 -38
SNP   GT     197 /     154 -41
SNP   FN     176 /     414 +38

InDel TP   14492 /   13885
InDel FP    7116 /    7008 -14974 !!
InDel GT    2000 /    1930 -331
InDel FN     477 /    1084 +58
```

`-p` harms FN a little, but benefits FP and GT errors, especially again for indels where FP gain is once again enormous.  It's clearly something that justifies being an option.  This is optional and not suitable for blocking-based chemistries.  It's documented in the man page as being suitable for HiFi.

For reference, this is just a noddy fast consensus algorithm rather than a variant caller, which is why it doesn't have a VCF output (the above is from a hacked up pileup2vcf conversion tool).  However vs bcftools it's clearly behind.  I'm a bit disappointed in the big SNP FN change, so maybe that's something to research further.

```
SNP          Q>0 /   Q>=30
SNP   TP   85605 /   85569
SNP   FP    1208 /     864
SNP   GT     166 /     166
SNP   FN      36 /      72

InDel TP   14687 /   14366
InDel FP   10957 /    3465
InDel GT    1863 /    1857
InDel FN     282 /     603
```

================================================================================

The remaining question is the impact on Illumina data.  I used a chunk of SynDip chr1 with Illumina data, again 1.16 vs new (without -p, so just the gap quality change):

```
consensus116
SNP          Q>0 /   Q>=30
SNP   TP  153763 /  152845
SNP   FP    3517 /    2566
SNP   GT     614 /     446
SNP   FN    2315 /    3233

InDel TP   20324 /   20188
InDel FP    1229 /    1046
InDel GT    2305 /    2282
InDel FN    6522 /    6658

consensus new (gap score change)
SNP          Q>0 /   Q>=30
SNP   TP  153779 /  152871
SNP   FP    3577 /    2611 +45
SNP   GT     613 /     446   =
SNP   FN    2299 /    3207 -26

InDel TP   20319 /   20182
InDel FP    1219 /    1034 -12
InDel GT    2270 /    2247 -35
InDel FN    6527 /    6664  +6
```

Slight 1-2% shifts with Illumina in either direction and unclear which is better tool here.  On this data vs bcftools, it's less clear who wins for SNPs as it's neck and neck, but again bcftools is obviously vastly better on indels due to the realignment code:

```
SNP          Q>0 /   Q>=24 / Filtered
SNP   TP  153408 /  152841 /  152762
SNP   FP    4035 /    3127 /    2431
SNP   GT     445 /     399 /     362
SNP   FN    2670 /    3237 /    3316

InDel TP   22062 /   21826 /   21734
InDel FP     985 /     782 /     720
InDel GT     941 /     906 /     887
InDel FN    4784 /    5020 /    5112
```

bam_consensus.c
```diff
@@ -380,6 +384,49 @@ int nins(const bam1_t *b){
...
+
+        // Simplest:  reverse if end_qual < start_qual
+        // Next:      average outer-most two, then next two, etc
+        // Best:      fully redistribute so start/end lower qual than centre
+
+        // Middle route of averaging outer pairs is sufficient?
+        for (int j = s, k = i; j < k; j++,k--) {
+            double e = ph2err[qual[j]] + ph2err[qual[k]];
+            qual[j] = qual[k] = -fast_log2(e/2)*3.0104+.49;
+        }
```
daviesrob, line 426: The result of this transformation is that the new quality value is the minimum of the original two, plus a factor which varies between 0 and 3 depending on the absolute difference of the qualities.  So, if I've got this right the same result could be had without the floating-point manipulations like this:

```
uint8_t adjust[] = { 0, 0, 1, 1, 2, 2, 2, 2, 2, 2, 3, 3, /* 3's omitted ... */, 3 };
qual[j] = qual[k] = MIN(qual[j], qual[k]) + adjust(ABS(qual[j] - qual[k]));
```

Adding three is essentially halving the probability that the base call is incorrect (and arises from the `/2` in the original formulation).

Presumably if there's a big drop along the length of the run, the quality at the high end goes down by quite a lot.  Could this be the reason for the slight increase in false negatives you saw in your tests?

jkbonfield: Yes, it's basically +3 in the limit of large difference due to it being phred for 0.5.  I did think about doing that, but the above code was clearer and `fast_log2` wasn't a bottleneck that I could see.

Note this isn't a drop along the length of a run (that was one idea I had and is in the comments above), but rather swapping ends only.  However potentially there is benefit to a more complex analysis, e.g. when TAAAAAC is mutated to TAAGAAC.  It's low hanging fruit stuff though.  An alternative is sorting instead of averaging, e.g. left normalisatoin.  I haven't explored that though and my gut feeling is it may make problems elsewhere.

I did also wonder about a slight rescaling of the resulting qual values (for all cases obviously, not just homopolymers) so we have similar FN rates and perhaps rescue the lowest qual ones.  However none of this is particularly well calibrated in the first place and QUAL filtering is therefore somewhat arbitrary.  It may still be something to consider though.

jkbonfield: I should also add in the spirit of more smaller PRs instead of large ones, this is part 1.

Part 2 revises the consensus algorithm itself, instead of the quality values supplied to it.  It's based on the later platform-specific indel rates added to the gap5 consensus algorithm.  This has a significant impact on calling, but it still greatly benefits from the simple -p transformation here.

daviesrob:
I've been having a look at how well this works with some ONT test data, from [PRJEB51164](https://www.ebi.ac.uk/ena/browser/view/PRJEB51164?show=reads).  I tested  [ERR8958469](https://www.ebi.ac.uk/ena/browser/view/ERR8958469) ("Ecoli CFT073 isolate ONT sequenced with r10.4 using model HAC") and [ERR8958471](https://www.ebi.ac.uk/ena/browser/view/ERR8958471) ("Ecoli CFT073 isolate ONT sequenced with r10.4 using model DUPLEX").  ERR8958471 also used the slower, but higher quality, base caller.  For the analysis, I'm assuming that the CFT073 reference is the correct result.

The second commit was mostly an improvement for both inputs.  For ERR8958469, 14 changes whet towards the reference, and one against.  For ERR8958471, four changes went towards the reference, four others were difficult to judge due to the number of other differences in the region, and none obviously went against the reference.

The `-p` option had rather more mixed results.  For ERR8958469, 19 changes went towards the reference, 9 against and one was difficult to judge.  For ERR8958471, two changes went towards the reference, three against, and three were difficult to tell.  So it looks like it's more use with lower-quality ONT data, but definitely a your mileage may vary thing.

I think on the whole this is worth going with, given that `-p` does help in certain cases, and it has to be specifically enabled.

PR #1722, opened 2022-09-21
Reads wrongly being removed in hard clip mode.

### LINKED ISSUE #1717: Samtools ampliconclip hard clip not giving same results as soft clip
Hi all,

After stumbling on this post: https://bioinformatics.stackexchange.com/questions/16112/precisely-clipping-bam-file-to-bed-coordinates, I found out about the samtools ampliconclip package and wanted to use it for my own project. However, I'm running into some issues with using the samtools ampliconclip --soft-clip versus --hard-clip option.

Essentially, what I'm trying to do is clip certain reads of interest in a bam file down to a specific window that I specify using a bed file. However, when I run ampliconclip with --soft-clip, it gives me what I want, but when I run it with --hard-clip, it basically produces a blank bam file as output.

This is a picture of the unprocessed bam file of interest
Screen Shot 2022-09-13 at 8 18 57 PM

This is the complement bed file that I want to clip
```
9   0   33261120
9   33262737    149000000000
```

After running the code 
```
samtools ampliconclip --both-ends --soft-clip -b input.bed input.bam> output.bam
```
 I get the result I want, which is all my reads clipped besides a certain genomic interval
Screen Shot 2022-09-13 at 8 25 54 PM

However, what I really want is the hard clipped version of these reads. However, when running the same code 
```
samtools ampliconclip --both-ends --hard-clip -b input.bed input.bam> output.bam 
```
I basically get an empty file with no reads at all.
Screen Shot 2022-09-13 at 8 19 20 PM

From my understanding, using ampliconclip with --hard-clip versus --soft-clip should only change the CIGAR string and not how the intersection with the bed file should work, so I'm a little stumped as to why its giving me such different outputs. I have even tried --hard-clip with other bam files (from yeast) and it works as intended, but just not with these reads (from human). Any help is greatly appreciated!

whitwham:
The pre hard clip check was wrongly removing entire reads.  The large clip was bigger than the query length (calculated with the CIGAR string) but not the length against the reference (which is what was needed).

Changed to use the length against the reference.

Fixes #1717.


daviesrob:
It looks like this works, albeit at the expense of an extra CIGAR walk when hard clipping.  I wonder though if we could optimise that away?  Given that we have to walk the CIGAR data anyway, I think it should be possible to move the everything hard-clipped away handling to [here](https://github.com/samtools/samtools/blob/798a9776/bam_ampliconclip.c#L358) in `bam_trim_left()`, and [here](https://github.com/samtools/samtools/blob/798a9776/bam_ampliconclip.c#L503) in `bam_trim_right()`.  Getting to those places should be equivalent to the `bases >= bam_cigar2rlen(rec->core.n_cigar, bam_get_cigar(rec))` test, so we could just do `if (clipping == hard_clip) { /* remove seq/qual/cigar as before */ }` at those locations and get the same result.

I've tried it on the test data in this PR, and it seems to work OK.  What do you think?

We could also just remove the test completely, but that changes the output.  Instead of an unmapped read, you get one with a single CIGAR entry saying everything has been hard-clipped.  It's debatable which is better.

whitwham:
I think your idea is worth a try.  It works on the data provided by @KevinXu264 and my own test.

PR #1724, opened 2022-09-22
Fix SEGV in samtoools collate when no filename given.

jkbonfield:
"/dev/stdin" or "-" both work, but just reading from stdin crashed.


jkbonfield:
Correct, although I was bug fixing the existing code rather than making a decision on whether or not this command should support this syntax.

I haven#t checked if it ever did.  If not then it wouldn't be a regression to remove that ability and go with the refusal to read stdin without an explicit request.

PR #1726, opened 2022-09-27
Use POSIX `grep`

SoapGentoo:
* GNU grep 3.8 considers `egrep` and `fgrep` obsolescent and throws warnings: https://lists.gnu.org/archive/html/info-gnu/2022-09/msg00001.html


SoapGentoo:
@whitwham

whitwham:
Thank you @SoapGentoo.

PR #1733, opened 2022-10-13
Improvements to samtools consensus for platforms with higher indel error rates

jkbonfield:
- Make -p also account for homopolymer scoring in nm_local / qmin adjustments

- Add in Gap5 consensus algorithm improvements (--P-indel).

  Specifically the probability of indel related errors (both genuine over/under-calls but also alignment artifacts and reference bias). These hugely help platforms with homopolymer errors, such a ONT, Ultimate Genomics and PacBio HiFi.

  P-indel isn't really a probability, but a scale applied to various confidence values.  We may be able to model correctly if all input qualities were correctly calibrated, but they are not remotely close.  So this is more of a scaling heuristic.

- Improved the consensus_init scaling so we can pick between gap5's originally scoring functions and a pacbio-ccs optimised alternative.  Possibly we will need to tune these parameters per platform (see comment above about incorrect calibration) to get maximum performance.

- Extend -m MODE parameter to permit specification of bayesian_116 (samtools 1.16 compatible), as well as explicitly permit the different parameter sets (gap5 vs new bayesian tuning).

- Add a method to compute the consensus twice with different parameters, merging the results together.  This provides a better FN vs FP middle ground that covers more QUAL thresholds.  This is only used when --P-indel >= 0.01 is used as otherwise it has little impact.

- Add a consensus -X profile / --config profile option.

  At present it just understands "illumina" and "pacbio-ccs". Names are chosen to match bcftools mpileup -X.

  TODO: add configurations for more machine types as we tune to other data sets.


jkbonfield:
The consequence of adding more fields to sum over into the consensus algorithm is a small performance drop, but it's not tragic.

On 10MB region of PacBio CCS GIAB HG002 set, samtools develop with `-p` option took 31s.  This branch with same opt took 41s (that's the impact of separating subst from indel error distributions), while this branch with `-X pacbio-ccs` took 63s.  The latter shows the cost of calling the consensus algorithm twice. Using `-X pacbio-ccs -m bayesian_p` to call the consensus just once (favouring precision over recall) took 41s again as expected.

The benefit of these on the false-negative vs false-positive graph can be seen below:

image

image

Note all of these are pre-filtering.  Some tools (notably GATK) remove a very significant number of false positives through filtering, although obviously this cannot remote false-negatives.  In the above plots, "_7" is the default bayesian mode while "_7b" is the `bayesian_p`.  The most significant difference (other than speed) is the more precise method starts with a much higher baseline of false-negatives, and at that level of FN it also has fewer FP.

Note this isn't doing any realignment, so I'm not so concerned over accuracy in poor regions.  Rather the raison d'etre is to faithfully produce the consensus given the alignments presented in the input data, either for the purposes of iteratively realigning against them or for use as a better quality embedded reference for CRAM.

Future work needs more machine types assessing, but this is still one more machine type handled well than we had before!

jkbonfield:
Strictrly for local users wishing to evaluate, my command is:

```
cd ~jkb/lustre/GIAB/chr1
samtools consensus -X pacbio-ccs -@8 -C0 -A -r chr1:10M-20M --show-ins yes --show-del no -f pileup HG002.GRCh38.PacBio_CCS_15Kb.bam |  ./pileup2vcf2.pl > _.cons.vcf
FORMAT=gnuplot NORM=0 ../compare_vcf_simple.sh HG002_GRCh38_1_22_v4.2_benchmark.chr1.vcf.gz _.cons.vcf "" ../HG002_GRCh38_1_22_v4.2_benchmark.bed chr1:10M-20M
```

The pileup2vcf.pl is a simple perl script to process the pileup output from samtools consensus and generate a noddy vcf (I thought about adding it in to consensus itself, but it needs to know the reference and this feels out of place with the fundamental aim of the tool).  The compare_vcf_simple.sh script is a combination of norm, isec, and counting with QUAL >= x filtering rules to produce data suitable for plotting.

If people feel a vcf output mode should be added to consensus, then that's probably a new PR.  However given the caveats with no realignments etc (meaning it's less precise than other tools on e.g. illumina data) then I'm reticent incase it leads to a bunch of bug reports on "why didn't it spot this variant"?  This tool is more as a better quality replacement for e.g. ivar consensus.

jkbonfield:
PacBio-CCS, again from GIAB ftp site (HG002.GRCh38.PacBio_CCS_15Kb.bam).  There's not a huge difference between new PR and old PR, except for not quite going as far down to the low FN area.  However it's only using one consensus algorithm so is simpler and faster.  Without specifying the profile though, it's doing poorer than the old devel branch.

This is why I wonder if the default parameters could do with being tweaked somewhat (assuming it doesn't tragically break other data sets).

image

jkbonfield:
ONT R10.4 called with Guppy "super" mode.  Approx 19x depth (higher depth data coming soon).

image

The old PR didn't have a configuration for ONT, but I chose CCS mode as an example given the similarities in errors.  It's clear the new branch is a considerable improvement over devel (as was the old one), although it has a high FN rate.

jkbonfield:
Finally higher depth ONT R10 data (~70x).  This is, sadly, not working well yet.  More for me to do.  The old PR (using CCS as there's no ONT profile) is handsomely beating the new method.  However it's also using the "mixed" mode of calling twice and blending results.  I was hoping to try and avoid it, but it may still be the clincher for ONT.

image

jkbonfield:
Revised plots for ONT (plus noting my old plots had a bug where the max FN rate was 50% instead of 100% due to a bug in my plotting).  I switched out of logscale for this as it's easier to view the relevant portions.

Lower depth: 15-20x
image

Higher depth ~70x:
image

The plots show the FN vs FP rates against the truth set for the current develop, and this revised PR when using `-X hifi`, `-X r10.4_sup` (PR-ont-0.01: with default `--homopoly-redux 0.01`) and `-X r10.4_sup --homopoly-redux 1.0` (PR-ont-1.0).

The homopoly-redux code modifies the quality values based on the length of a homopolymer, to scale down the confidence as they get longer.  As expected, this removes a lot of false positives on platforms with great variability in homopolymer lengths, but also as a consequence removes some true variants.  The effect is minimal for SNPs, but can  be significant for indels.  However this PR leaves it set to 0.01 for all platforms for now as I'm wondering if we wish to advertise it as a feature given it's tricky to describe and the tradeoff isn't *always* justified if we favour recall over precision.

I note the HiFi calibration table is still beneficial for ONT data (over the ONT table) if you're after maximising the recall of heterozygous SNPs, at about half the false negative rate and a bit over double the false positive rate.  These tradeoffs are somewhat annoying in that they persist at all QUAL levels.  That was the intention for the mixed mode - that we could have low QUAL values favouring recall and higher QUAL favouring precision by calling twice and blending the two result sets together.  That code is still there too, but also not enabled by default in any standard profile.  Again, it's slow and very heuristic driven so hard to describe formally in the manual.

I think therefore we could just have some experimental features left in and working, but not publically advertised in usage or the man page.  That gives us a step forward in consensus quality while leaving us room to improve in the future.

TODO: I'd also like to reevaluate Illumina using the pacbio/ont parameters (but an Illumina calibration table).  The logic here is that in future this code should be capable of calling consensus on mixed data sets, applying calibration per read-group based on the PLatform tag, so global parameters don't work well there.

TODO2: Also evaluate the `@flat` qual-calibration per platform, so we can see what the effect would be when not specifying any machine specific profile.  This would be useful as it's a proxy for what happens when we have platforms we haven't evaluated yet, new chemistries, or just data that's subtly different in other ways.

bam_consensus.c
```diff
@@ -468,63 +958,69 @@ int nm_init(void *client_data, samFile *fp, sam_hdr_t *h, pileup_t *p) {
...
-        for (;i < qlen-qhalo; i++) {
-            int t = (qual[i]   + 5*qmin)/4; // good on 15x
-            local_nm[i] += t < qual[i] ? (qual[i]-t) : 0;
+            local_nm[i] += t < qual[i] ? qual[i]-t : 0;
+
+            // Brute force qminp in polyl to polyr range.
+            // TODO: optimise this with sliding window
+            qminp = qual[i];
+            for (int j = MAX(polyl,i-qhalop); j <= MIN(polyr,i+qhalop); j++)
+                if (qminp < qual[j])
```
daviesrob, line 1021: This is finding the max rather than the min.  Is that correct?

bam_consensus.c
```diff
@@ -536,16 +1032,36 @@ int nm_init(void *client_data, samFile *fp, sam_hdr_t *h, pileup_t *p) {
...
+        for (j = i+1; j < qlen; j++)
+            if (bam_seqi(seq, j) != base)
+                break;
+        //printf("%d x %d\n", base, j-i);
+
+        poly = j-i-1; if (poly > 100) poly = 100;
+        const int HALO=0;
+        for (k = i-HALO; k < j+HALO; k++)
+            if (k >= 0 && k < qlen)
+                local_nm[k] |= (MAX(poly,local_nm[k]>>24))<<24;
```
daviesrob, line 1061: This works for `HALO=0`, but the update would need to be done rather more carefully if it was bigger...

bam_consensus.c
```diff
@@ -1557,12 +2407,18 @@ int main_consensus(int argc, char **argv) {
         {"min-MQ",             required_argument, NULL, 13},
         {"min-BQ",             required_argument, NULL, 16},
         {"P-het",              required_argument, NULL, 15},
+        {"P-indel",            required_argument, NULL, 17},
+        {"het-scale",          required_argument, NULL, 19},
         {"mode",               required_argument, NULL, 'm'},
         {"homopoly-fix",       no_argument,       NULL, 'p'},
+        {"homopoly-score",     required_argument, NULL, 'p'+100},
+        {"homopoly-redux",     required_argument, NULL, 'p'+200},
```
daviesrob, line 2415: What does this do?  It doesn't appear to be documented anywhere.

bam_consensus.c
```diff
@@ -468,63 +958,69 @@ int nm_init(void *client_data, samFile *fp, sam_hdr_t *h, pileup_t *p) {
...
 
+    double poly_adj = opts->homopoly_fix ? opts->homopoly_fix : 1;
+
     if (opts->adj_qual) {
-#if 0
-        // Tweak by localised quality.
-        // Quality is reduced by a significant portion of the minimum quality
-        // in neighbouring bases, on the pretext that if the region is bad, then
-        // this base is bad even if it claims otherwise.
+        // Skew local NM by qual vs min-qual delta
```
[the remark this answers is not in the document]
jkbonfield: Agreed, although the comment wasn't a new addition - it's just a github display problem.

I've revised it, but the algorithm is definitely a bit "black magic" mixed with pragmatism to find something that works.

bam_consensus.c
```diff
@@ -468,63 +958,69 @@ int nm_init(void *client_data, samFile *fp, sam_hdr_t *h, pileup_t *p) {
...
         uint8_t *qual = bam_get_qual(b);
-        const int qhalo = 8; // 2?
-        int qmin = 50; // effectively caps PacBio qual too
+        uint8_t *seq = bam_get_seq(b);
+        const int qhalo = 8; // 4
+        int qmin = 99, qminp = 99;
+        int base = bam_seqi(seq, 0), polyl = 0, polyr=0; // pos, not len
```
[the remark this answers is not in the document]
jkbonfield: It's filling out the local_nm array, which gets assigned to `p->cd`.  Elements of that array are then returned by the `nm_local` function, returned via indices from `p->seq_offset` (with a `b->core.pos` added and then subtracted again as the interface is working in ref coords instead).  So I think it's correct to be working on the raw sequence line including soft-clips.

Besides, it's figuring out homopolymer lengths amongst other things, which may extend into the soft-clip.  The same applies for regions of low confidence.

bam_consensus.c
```diff
@@ -468,63 +958,69 @@ int nm_init(void *client_data, samFile *fp, sam_hdr_t *h, pileup_t *p) {
...
+        const int qhalop = 2;
+
+        for (i = 1; i < qlen; i++) {
+            if (bam_seqi(seq, i) != base)
+                break;
+            if (i < qhalop && qminp > qual[i])
+                qminp = qual[i];
+        }
+        polyr = i-1;
```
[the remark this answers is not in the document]
jkbonfield: Only if `opts->mode == MODE_BAYES_116`.

We assign `int pl = polyr-polyl` (homopolymer length), which then gets used in the calculation of `t`.  That's the minimum quality threshold such that lower values cause us to start aggregating `local_nm` qual adjustments, as evident in the `t < qual[i] ? qual[i]-t : 0` below.

Arguably some of the code could be behind conditionals to avoid computation when options mean the results won't be used, but I doubt it'll help much in general usage.

bam_consensus.c
jkbonfield: I could cull that, given it was a failed experiment that didn't help much, but I don't understand what's wrong with it.

Basically it changes a loop that goes from `i` to `j` (coords of the homopolymer) to `i-halo` to `j+halo` to include a few bases either side of the homopolymer.  This means we adjust `local_nm` not only for "within a homopolymer" but also "adjacent to a homopolymer".  It sounded like a reasonable thing to consider, but didn't really pan out.  I left it there incase different aligners and/or instruments demonstrated there are reasons to do this, in which case it'd become a parameter.

There is bounds checking in the code though, regardless of `HALO` value.

It gets set into the top bits of `p->cd` so the `poly_len` function can skew base quality.  The amount of skew is the `--homopoly-redux` option, which potentially has some benefit to ONT.  However it's quite a tradeoff of FP vs FN and for simplicity I decided to leave it set to a constant amount and left it out of the documentation and usage statements.  Basically it may help, but the jury is out and I didn't want to lose the code.

So for now it's just a fixed non-zero factor (0.01), which essentially just means decrementing qual by 1 if poly-len is > 2.
I could ifdef out all code involved if you prefer.

bam_consensus.c
jkbonfield: Deliberately so.  They're experimental and don't appear in the documentation yet.

bam_consensus.c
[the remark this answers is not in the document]
jkbonfield: Sorry I see what you mean now.  Polyr is used, but it's reassigned again after this assignment.  I'll see what can be culled.  The initial assignment of polyl is needed however.

It's possible qminp can be culled too from that initial loop, with a small change.

bam_consensus.c
jkbonfield: With `qminp` actually being the min value, it's working better mostly.

The first 140MB of chr1 for the HG002 PB Hifi sample:

image

image

image

A slight increase in FN for deletions, but a big drop in FP for insertions and a modest one for SNPs.  Overall a win.

bam_consensus.c
```diff
@@ -468,63 +958,69 @@ int nm_init(void *client_data, samFile *fp, sam_hdr_t *h, pileup_t *p) {
...
+        const int qhalo = 8; // 4
+        int qmin = 99, qminp = 99;
+        int base = bam_seqi(seq, 0), polyl = 0, polyr=0; // pos, not len
+        const int qhalop = 2;
+
+        for (i = 1; i < qlen; i++) {
+            if (bam_seqi(seq, i) != base)
+                break;
+            if (i < qhalop && qminp > qual[i])
+                qminp = qual[i];
```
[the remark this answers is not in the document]
jkbonfield: Corrected

bam_consensus.c
daviesrob: You're using bitwise-or to set the new value.  That's OK if the old one was zero, but if you allow `k` to extend outside the range `(i .. j]` then the top bits of `local_nm[k]` might have been set by an earlier pass through the loop.  If that happens, the bitwise-or may set the wrong value depending on exactly which bits are set.

bam_consensus.c
```diff
@@ -1466,7 +2315,18 @@ static void usage_exit(FILE *fp, int exit_status) {
...
     fprintf(fp, "      --P-het FLOAT     Probability of heterozygous site[%.1e]\n",
             P_HET);
+    fprintf(fp, "      --P-indel FLOAT   Probability of indel sites[%.1e]\n",
+            P_INDEL);
+    fprintf(fp, "      --het-scale FLOAT Heterozygous SNP probability multiplier[%.1e]\n",
+            P_HET_SCALE);
     fprintf(fp, "  -p, --homopoly-fix    Spread low-qual bases to both ends of homopolymers\n");
+    fprintf(fp, "      --homopoly-score FLOAT\n"
+                "                        Qual fraction adjustment for -p option [%g]\n", P_HOMOPOLY);
+    fprintf(fp, "  -t, --qual-calibration FILE / @config (see man page)\n");
```
daviesrob, line 2325: Before we bake this in, do we want to use `FILE / @config` or `@FILE / config`?  The latter is more consistent with the `@FILE` syntax used in [bcftool's expressions](https://www.htslib.org/doc/bcftools.html#expressions) (although I don't think it uses `@FILE` anywhere else).  Other tools that work this way include [curl](https://curl.se/docs/manpage.html#-d) and [javac](https://docs.oracle.com/javase/9/tools/javac.htm#GUID-AEEC9F07-CB49-4E96-8BC7-BCC2C7F725C9__STANDARDOPTIONSFORJAVAC-7D3D9CC2).  Unfortunately I haven't worked out a good way of searching for tools that might go the other way to find out how many there might be.

jkbonfield: Good question.  We could reverse the syntax I guess, but I'm not sure what the best approach is.

My thought was the primary option is a calibration file and this is what most people would do.  Given the inbuilt tables are basically only there to be used by the `-X profile` canned configurations, it's unlikely people wish to refer to those directly outside of a `-X` option.  They're more or less there just for completeness sake.  (Then again, I'm not that convinced people are going to specify external files either, and again it's more for development of new profiles than much else.)

Bcftools doesn't explain what `@filename` does.  Is it essentially the equivalent of an inline `` `cat filename` `` within an expression?  Similarly @ is working that way for curl and javac.  If so then it's not really the same as what we're doing here, as we're not specifying the calibration data in the command line itself and it's not a synonym for backquoting to load in data and paste in-line.

daviesrob: I'm not entirely clear about `bcftools`, but it looks like it allows you to create a set of rules that match a set of samples in a file.  It doesn't just read an expression from a file.  Curl uses it to indicate that the content for the given parameter should come from a file, e.g. `-F profile=@portrait.jpg` makes it do a POST with a file upload.  The java one, on the other hand, reads options from the file, much as if you'd put them on the command-line.  So it's not that standardised.

I guess we can run with the way this currently works, but you can handle any complaints that it doesn't work in the same way as some other tools 😄 .

jkbonfield: Well I could be persuaded the other way, but won't we then also get complaints that it's the only FILE argument that needs an @ symbol?

I think either way is a bit weird tbh.  Maybe it needs a different symbol?

jkbonfield: Thinking more, although there is absolutely no real standard between the tools you mentioned with what they do with an `@file`, the one common thread they have is that the `@` means "_do something with this file instead of the normal CLI parameter_".

Therefore I think there is a decision here.

1. Is the "normal" parameter the internal baked-in named calibration table with the file being the exception?  If so the usage needs to not state "FILE", and we can have `@file` as an exception.

2. Is the "normal" parameter the filename mode, in which case the exception shouldn't be using the `@` symbol as that's typically associated with filenames.  Maybe `#` or `=` or something else that's unlikely to start a filename.  I don't know of any commonly used mechanisms for that.

PR #1746, opened 2022-11-02
Consensus --mark-ins option

jkbonfield:
Permits the consensus output to include a markup indicating the next base is an insertion.  This is necessary as we need a way of outputting both consensus and also how that consensus marries up with the reference coordinates.  We could write out an additional file, but that's more effort than simply skipping the markup characters and building a map while reading the file.

Note this builds on top of #1733

### DIFF
--- a/bam_consensus.c
+++ b/bam_consensus.c
@@ -169,6 +169,7 @@ typedef struct {
     int all_bases;
     int show_del;
     int show_ins;
+    int mark_ins;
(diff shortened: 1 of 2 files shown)


jkbonfield:
Rebased it off develop with the earlier bit nuked so it's just this option, given the delays in other consensus improvements (and my local need for a tool with both is no longer present).

PR #1767, opened 2022-12-10
Reset

vasudeva8:
Implementation of reset functionality is available in this PR. Please have a review and share any suggestions / corrections required.

Reference data are discarded, RG/PG lines are option controlled. Flags are updated (proper pair removed, marked as unmapped and mate unmapped). Secondary and supplementary alignments discarded and reverse flag reset after reversing and complementing sequence and quality data.

A few aux tags are defined as default and other tags can be selected/deselected with keep / reject csv file.

PG entries can be kept as such or fully rejected. Selected entries can be discarded by passing them through csv file with option. PP tags pointing to removed PG entries will be removed.


Makefile
```diff
@@ -213,6 +213,7 @@ bam_markdup.o: bam_markdup.c config.h $(htslib_thread_pool_h) $(htslib_sam_h) $(
 tmp_file.o: tmp_file.c config.h $(tmp_file_h) $(htslib_sam_h)
 bam_ampliconclip.o: bam_ampliconclip.c config.h $(htslib_thread_pool_h) $(sam_opts_h) $(htslib_hts_h) $(htslib_hfile_h) $(htslib_kstring_h) $(htslib_sam_h) $(samtools_h) bam_ampliconclip.h
 bam_samples.o: bam_samples.c config.h $(htslib_hts_h) $(htslib_sam_h) $(htslib_faidx_h) $(htslib_khash_h) $(htslib_kseq_h) $(samtools_h)
+reset.o: reset.c $(samtools_h) $(htslib_hts_h)
```
jkbonfield, line 216: This needs more dependencies adding.  From the source I see:

    #include "samtools.h"
    #include "htslib/sam.h"
    #include "sam_opts.h"
    #include "htslib/khash.h"

bamtk.c
```diff
@@ -72,6 +72,7 @@ int main_import(int argc, char *argv[]);
 int main_samples(int argc, char *argv[]);
 int main_consensus(int argc, char *argv[]);
 int main_reference(int argc, char *argv[]);
+int main_reset(int argc, char *argv[]);     //reset facility implementation
```
jkbonfield, line 75: Can drop the comment as it's out of place compared to all others.

reset.c
```diff
@@ -0,0 +1,888 @@
+/*  reset.c --  removes aligner updates and reference data from input sam/bam/cram file and makes read data raw
+                for new processing
+
+    Copyright (C) 2022 Genome Research Ltd.
+
+    Author: Vasudea Sarma <vasudeva.sarma@sanger.ac.uk>
```
jkbonfield, line 6: - If we're going to line wrap, it's probably best to do so at the same
  column width used elsewhere in the file, which is (usually) 80
  chars.  Eg see the very first comment line for over-long wrapping.

- Typo in author name? Vasudea

reset.c
```diff
@@ -0,0 +1,888 @@
...
+
+#include "htslib/sam.h"
+#include "sam_opts.h"
+#include "htslib/thread_pool.h"
+#include "htslib/khash.h"
+
+KHASH_SET_INIT_INT(aux_set);        //set hold the aux tags to be retained
+KHASH_SET_INIT_STR(pg_set);         //set hold the pg tags to be removed
+
+#define TAGNUM(X) ((X)[0] << 8) | (X)[1]    //to create key for aux tags, like type key in htslib
```
jkbonfield, line 38: A style thing, but best to add surrounding parentheses around macros, incase they're ever used in complex expressions that change precedence, eg TAGNUM(x)+1.  (A moot point here though)

    #define TAGNUM(X) ((X)[0] << 8) | (X)[1]

vs

    #define TAGNUM(X) (((X)[0] << 8) | (X)[1])

reset.c
```diff
@@ -0,0 +1,888 @@
...
+
+typedef struct conf_data
+{
+    int keepRGs;                    //RG line handling
+    int keepPGs;                    //PG line handling
+    khash_t(aux_set) *auxtagsin;    //SET that holds the aux tags to be reatined
+    khash_t(pg_set) *pgout;         //SET that holds the pg tags to be removed
+    int tmpfield;                   //tmp field
```
jkbonfield, line 46: `tmpfield` isn't very descriptive.  Given the usage, maybe rename to `reject`?

reset.c
```diff
@@ -0,0 +1,888 @@
...
+KHASH_SET_INIT_INT(aux_set);        //set hold the aux tags to be retained
+KHASH_SET_INIT_STR(pg_set);         //set hold the pg tags to be removed
+
+#define TAGNUM(X) ((X)[0] << 8) | (X)[1]    //to create key for aux tags, like type key in htslib
+
+typedef struct conf_data
+{
+    int keepRGs;                    //RG line handling
+    int keepPGs;                    //PG line handling
+    khash_t(aux_set) *auxtagsin;    //SET that holds the aux tags to be reatined
```
jkbonfield, line 44: Type for "retained"

reset.c
```diff
@@ -0,0 +1,888 @@
...
+
+    return 0;
+}
+
+/// getcsvauxtags - aux tag parsed from csv input file, validate and add to internal set
+/** @param field - the aux tag parsed
+ *  @param data - pointer to conf_data which holds internal flags and aux_set
+returns -1 on failure and 0 on success
+*/
+int getcsvauxtags(const char *field, void *data)
```
jkbonfield, line 102: Why is data "void *"?  It's called from one function with type conf_data *, which it is cast into anyway.

reset.c
```diff
@@ -0,0 +1,888 @@
...
+    len = strlen(field);
+    if (len != 2)                                                       //invalid tag!
+        return -1;
+
+    if (!config->tmpfield)                                              //tmpfield says add or remove
+        kh_put(aux_set, config->auxtagsin, TAGNUM(field), &ret);        //add to set
+    else {                                                              //remove from set
+            khint_t iter = kh_get(aux_set, config->auxtagsin, TAGNUM(field));
+            iter == kh_end(config->auxtagsin) ? 0 : kh_del(aux_set, config->auxtagsin, iter);
+    }
```
jkbonfield, line 119: Inconsistent indentation.  Please keep to 4 spaces per level.

Also (personal preference), I dislike mixed bracing and non-bracing.  Ie I'd replace

```
if (expr)
    blah1;
else {
    blah2;
    blah3
}
```

with

```
if (expr) {
    blah1;
} else {
    blah2;
    blah3
}
```

We don't really have a style guide though, so I don't know how others think?  It's certainly not wrong, and we probably have lots of the former code style already present - so this is more for discussion than a request to change.

reset.c
```diff
@@ -0,0 +1,888 @@
...
+    if (!config->keepRGs) {
+        //RG lines not present, no point in having RG aux fields either, remove RG
+        khiter_t iter = kh_get(aux_set, config->auxtagsin, TAGNUM("RG"));
+        iter != kh_end(config->auxtagsin) ? kh_del(aux_set, config->auxtagsin, iter) : 0;
+    }
+
+    return 0;
+}
+
+int __get_aux_data(uint8_t* data, uint8_t *end, char *tag, char *type, void** ptr, int* len)
```
jkbonfield, line 346: What is this?  It looks unused.

Also generally we try to avoid starting functions with underscores as it's a namespace reserved for system use (although the chances of a clash are miniscule, so it's more of a style thing than something that's liable to cause breakage).

reset.c
```diff
@@ -0,0 +1,888 @@
...
+}
+
+/// parsecsvfile - parses the csv file and invokes the consumer function / callback
+/** @param field - the aux tag parsed
+ *  @param data - pointer to conf_data which holds internal flags and aux_set
+returns -1 on failure and 0 on success
+data expected in csv format terminated by '\n', parses with ',', callback is invoked to process the tokens
+rolls the buffer incase last char is not '\n' or ','. max supported tag is ~1023 long.
+*/
+int parsecsvfile(FILE *fp, int (*processdata)(const char *, void *), void *data)
```
jkbonfield, line 130: The comments above have `@param field`, but we have `processdata` instead.  I'm guessing this was from an earlier iteration.

Also, you may wish to consider looking into kgetline.  It returns a kstring of the whole line read and it grows automatically to fit it in.  This avoids the whole `partialread` shenanigans when using `fgets`, so it ought to make the code much simpler.

reset.c
```diff
@@ -0,0 +1,888 @@
...
+        return EXIT_FAILURE;
+
+    //set output file format based on name
+    sam_open_mode(outmode + 1, outname, NULL);
+
+    //open input and output files
+    infile = sam_open(inname, "r");
+    outfile = sam_open_format(outname, outmode, &ga.out);
+    if (!infile || !outfile) {
+        printf("Could not open %s%s%s", !infile ? inname : "", (!infile && !outfile)? ", " : "", !outfile ? outname : "");
```
jkbonfield, line 849: There are lots of `printf` calls for error messages.  These need to be on stderr and not stdout (see `-o file` vs stdout mentioned elsewhere.

However you may wish to look into `print_error` and `print_error_errno` calls instead.  The latter is best used in examples such as the `sam_open_format` where `errno` will have been set, so an appropriate system error message may be produced (eg "no such file or directory" or "permission denied").

reset.c
```diff
@@ -0,0 +1,888 @@
...
+    int c = 0, ret = 0;
+    char outmode[4] = "w", *args = NULL;
+    conf_data resetconf = {1, 1, NULL, NULL, 0};                //keep RGs and PGs by default
+
+    if (argc < 5)
+    {
+        //not enogh args
+        usage(stdout);
+        return EXIT_FAILURE;
+    }
```
jkbonfield, line 748: This has a number of problems.

Firstly, when using no arguments (`samtools reset`) it's almost always because the user is explicitly asking to get the usage statement, so that's fine to go to stdout, but if we *need* 5 arguments for correct usage and they specified 3 then that's a usage error and it should report to stderr.

However in this case I think it's incorrect anyway as you have a hard requirement on `-i file -o file`.  Generally we don't use `-i file` and just let `file` be the last thing on the command line.  Eg see `samtools sort` usage statement.

The use of `-o file` is good (thank you), but generally it should be optional.  As should the input filename.  When they're not specified the command should read and write to stdin / stderr, assuming it's not doing random access or needing multiple passes over the input stream.  This enables it to work cleanly on unix pipes.

Despite that, `-o file` is our preference over `> file` as it avoids the potential for mixing errors with the output file (especially when combined with the `nohup` command).

jkbonfield (commented):
A few comments from a first pass.  I haven't compiled and tested it yet, nor have I read the man page.  However here are a few things to be looking into meanwhile.

Thanks for the work.  It looks quite comprehensive.

Makefile
vasudeva8: Done.

bamtk.c
vasudeva8: Done.

reset.c
vasudeva8: :-) Done

reset.c
vasudeva8: Done

reset.c
vasudeva8: it is not really a configuration item but a flag to select different behaviour on invocation of same method (to add aux tag or to remove aux tag in aux tag callback function, it is not relevant for pg callback). As the signature of callback is same for all and the flag is not relevant for other cases, it was made part of conf_data structure and named as 'tmp' to make it generic and usable for any purpose. Updating its comments would help?

reset.c
whitwham: Is that a typo for typo? :smiley:

vasudeva8: Done

jkbonfield: Oops yes, it is a typo for typo.  How very meta!

reset.c
vasudeva8: Done.

reset.c
vasudeva8: Removed - missed during cleanup!.

reset.c
jkbonfield: I'm inclined to document the current usage how it works now (and name it as such), rather than be concerned about potential future reuse.  This isn't part of the public API as it's internal to samtools rather than htslib, so we can always rename it and generalise at a later date.

bamtk.c
```diff
@@ -181,6 +182,7 @@ static void usage(FILE *fp)
 "     fasta          converts a BAM to a FASTA\n"
 "     import         Converts FASTA or FASTQ files to SAM/BAM/CRAM\n"
 "     reference      Generates a reference from aligned data\n"
+"     reset          Remove alignment changes in sequences and make them as raw reads\n"
```
jkbonfield, line 185: This line is longer than all the others, meaning it wraps on an 80-char terminal.  While people often have larger terminals, maybe we can rephrase it as "Reverts aligned reads to unaligned" or similar?

reset.c
```diff
@@ -0,0 +1,849 @@
...
+/** @param fp pointer to the file / terminal to which usage to be dumped
+returns nothing
+*/
+static void usage(FILE *fp)
+{
+    fprintf(fp,
+            "Usage: samtools reset [options]\n\
+      -i <file path>        input file\n\
+      -o <file path>        output file\n\
+      -k <file path>        aux tags to be retained in csv file, optional\n\
```
jkbonfield, line 59: For sake of consistency, please see other subcommands for the description / style.

Generally we use `FILE` over `<file path>`.

We also usually have a `--no-PG` option to avoid adding the `@PG` line, and support `-@` as a short-hand form of `--threads`.

reset.c
jkbonfield: You may wish to look at how `samtools view --keep-tag` and `--remove-tag` options work.  We should probably use the same long and short option names too.

Samtools view usage:

```
  -x, --remove-tag STR
               Comma-separated read tags to strip (repeatable) [null]
      --keep-tag STR
               Comma-separated read tags to preserve (repeatable) [null].
               Equivalent to "-x ^STR"
```

The "^" bit is a negation operation.  (I think I copied that nomenclature from some of Petr's bcftools commands.)

You can copy the text out of the doc/samtools-view.1 man page as well.

reset.c
```diff
@@ -0,0 +1,849 @@
...
+    }
+
+    config->auxtagsin = kh_init(aux_set);    //initialize the new one
+
+    //add defaults, they may get overridden by reject option
+    kh_put(aux_set, config->auxtagsin, TAGNUM("BC"), &ret);   //barcode
+    kh_put(aux_set, config->auxtagsin, TAGNUM("CB"), &ret);   //cell identifier
+    //TODO add tags which are kept for backward compatibility?
+    kh_put(aux_set, config->auxtagsin, TAGNUM("LB"), &ret);   //library
+    kh_put(aux_set, config->auxtagsin, TAGNUM("MI"), &ret);   //molecular identifier
```
jkbonfield, line 92: These defaults need to be added to the usage statement.  The man page is also too wishy-washy when describing this.  It needs to be explicit stating the default list is exactly these 4 items and not implying there may be other unlisted ones too.

You can see examples of how we handle lists like this in the `samtools mpileup` usage:

```
  --ff, --excl-flags STR|INT  filter flags: skip reads with any of the mask bits set
                                            [UNMAP,SECONDARY,QCFAIL,DUP]
```

Note with `--excl-flags` it defaults to that list of 4 flags, and the user specifying a new list *replaces* the default rather than appending to it. Compare this to `samtools view --keep-tag` which doesn't replace as it appends to the list, but the list starts off empty.

I'm not sure how best to handle this.  There are both pros and cons to replacing vs appending, and there is also the option of doing both where the first usage is an explicit replace, and all subsequent are append.  That way it has a default list, but as soon as you specify any relevant command line options the default is void.  I'm open persuation either way.  Maybe @whitwham or @daviesrob have views on the best CLI for this?

reset.c
```diff
@@ -0,0 +1,849 @@
...
+        //update flags
+        uint16_t flags = bamdata->core.flag & ~BAM_FPROPER_PAIR;    //reset pair info
+        flags |= BAM_FUNMAP | BAM_FMUNMAP;                          //mark as unmapped
+        flags &= ~BAM_FMREVERSE;                                    //reset mate orientation
+
+        ks_resize(&querydata, bamdata->core.l_qseq);                //get enough space
+        ks_clear(&querydata);
+
+        ks_resize(&qualdata, bamdata->core.l_qseq);                 //get enough space
+        ks_clear(&qualdata);
```
jkbonfield, line 607: Both `querydata` and `qualdata` are a small memory leak.

jkbonfield (commented):
A few more comments to consider.

It's also slower than I expected: around 6x slower than samtools view (with both tools outputting as SAM).  The bulk of that appears to be the `ksprintf` calls when building up `querydata` and `qualdata`.  (I recommend "perf record CMD" and "perf report CMD" for doing quick profiles.)

I think this boils down to deconstructing the record and rebuilding it with `bam_set1`, which isn't a bad idea, however unfortunately it does mean conversion from encoded to string form and back again which is costly.  It may be better to edit in-situ to avoid this.  It's a lot messier though, so we can deal with this in another subsequent PR perhaps.

As an initial alternative, you could use `ks_resize` to ensure the string is large enough and just do byte by byte copying instead of `ksprintf` using `%c`.  Eg:

```
char *sp = ks_str(querydata);
char *qp = ks_str(qualdata);
for (i = 0; i < bamdata->core.l_qseq ; ++i)
    *sp++ = seq_nt16_str[bam_seqi(bam_get_seq(bamdata), i)];
memcpy(qp, bam_get_qual(bamdata)[i], bamdata->core.l_qseq);
```

It's possible tham `bam_get_seq` does more work than needed, but I'm unsure if the compiler can prove to itself that nothing in bamdata is changing.  Hopefully it's aware there are no possible overlaps between bamdata and querydata so it caches the pointer. If not, just take it outside the loop.

That may make it fast enough to not care about `bam_seq1` inefficiencies.

jkbonfield:
I decided to test the ksprintf avoidance, and on some long-read PacBio data, outputting to uncompressed BAM, this edit sped it up 9-fold.

```
diff --git a/reset.c b/reset.c
index bc65151..08f7629 100644
--- a/reset.c
+++ b/reset.c
@@ -645,20 +645,24 @@ int reset(samFile *infile, samFile *outfile, conf_data *config, char *args)
         ks_resize(&qualdata, bamdata->core.l_qseq);                 //get enough space
         ks_clear(&qualdata);
 
+        uint8_t *sp = (uint8_t *)ks_str(&querydata);
+        uint8_t *qp = (uint8_t *)ks_str(&qualdata);
+        uint8_t *seq = bam_get_seq(bamdata);
+        uint8_t *qual = bam_get_qual(bamdata);
         if (bamdata->core.flag & BAM_FREVERSE) {
             //sequence data ordered as reverse complemented, reorder/complement sequence and quality data as read and clear the flag
             for (i = bamdata->core.l_qseq - 1; i >= 0; --i) {
-                ksprintf(&querydata, "%c", "=TGKCYSBAWRDMHVN"[bam_seqi(bam_get_seq(bamdata), i)]);      //hack to get complemented data; TODO: this should be part of htslib?
-                ksprintf(&qualdata, "%c", bam_get_qual(bamdata)[i]);
+                *sp++ = "=TGKCYSBAWRDMHVN"[bam_seqi(seq, i)];
+                *qp++ = qual[i];
             }
             flags &= ~BAM_FREVERSE;                                 //reset flag as well
         }
         else {
             //data in read order itself
             for (i = 0; i < bamdata->core.l_qseq ; ++i) {
-                ksprintf(&querydata,"%c", seq_nt16_str[bam_seqi(bam_get_seq(bamdata), i)]);
-                ksprintf(&qualdata, "%c", bam_get_qual(bamdata)[i]);
+                *sp++ = seq_nt16_str[bam_seqi(seq, i)];
             }
+            memcpy(qp, qual, bamdata->core.l_qseq);
         }
 
         //remove aux tags and add bam data to output
```

It was only 30% slower than samtools view in this case.   The long-read nature of it probably meant the printfs became an even bigger bottleneck than on the short-read data I explored before.

reset.c
vasudeva8: Done.

bamtk.c
vasudeva8: Done.

reset.c
vasudeva8: Done.

reset.c
vasudeva8: Added ks_free

reset.c
vasudeva8: Done

reset.c
jkbonfield: The `--no-PG` option is serving a different purpose here, but it's a useful one that I didn't think about.  Your option is removing the existing PG lines, in a similar fashion to how `--no-RG` removes the RG lines.  That's a good option to have.

However the other tools use `--no-PG` as a way to avoid *adding* a new PG record.  I'm not sure what the best option naming would be here as there's a clash of functionality.  However minimally when specifying `--no-PG` we shouldn't be added a new PG record I think.  Ideally we'd have two options for control the two different types of PG editing here.

reset.c
jkbonfield: A further oddity about this is that it needs a newline to succeed.  If the last line isn't complete it is silently ignored.  Eg compare:

```
@ seq4c[samtools.../samtools]; ./samtools reset --no-RG -k <(perl -e 'print "NM\n"') -i ~/scratch/data/9827_2#49.1m.bam -o -|grep -c NM:i
99890
@ seq4c[samtools.../samtools]; ./samtools reset --no-RG -k <(perl -e 'print "NM"') -i ~/scratch/data/9827_2#49.1m.bam -o -|grep -c NM:i
0
```

I still think we should just copy the code from samtools view:  https://github.com/samtools/samtools/blob/develop/sam_view.c#L396-L421.  We can move that to utils.c or similar and share the same parsding and hash table functionality.

reset.c
```diff
@@ -0,0 +1,858 @@
...
+        return EXIT_FAILURE;
+    }
+
+    //get the aux tags to be retained
+    if (getauxtags(&resetconf, auxkeepname, auxrejname))
+        return EXIT_FAILURE;
+
+    //get the pg names to be discarded
+    if (getrejectedpgnames(&resetconf, pgrejname))
+        return EXIT_FAILURE;
```
jkbonfield, line 810: These can leak memory and so need a call to cleanup too.  To ensure cleanup doesn't try to do access memory it hasn't yet allocated, extra guards need adding there.

```
diff --git a/reset.c b/reset.c
index 6e6a8ce..85c2274 100644
--- a/reset.c
+++ b/reset.c
@@ -668,11 +668,13 @@ void cleanup(conf_data *config)
     khiter_t iter = 0;
     const char *tmp = NULL;
     kh_destroy(aux_set, config->auxtagsin);
-    for (iter = kh_begin(config->pgout); iter != kh_end(config->pgout); ++iter) {
-        tmp = (kh_exist(config->pgout,iter)) ? kh_key(config->pgout, iter) : NULL;
-        if (tmp)
-            free((void*)tmp);
-        tmp = NULL;
+    if (config->pgout) {
+        for (iter = kh_begin(config->pgout); iter != kh_end(config->pgout); ++iter) {
+            tmp = (kh_exist(config->pgout,iter)) ? kh_key(config->pgout, iter) : NULL;
+            if (tmp)
+                free((void*)tmp);
+            tmp = NULL;
+        }
     }
     kh_destroy(pg_set, config->pgout);
 }
@@ -802,12 +804,16 @@ int main_reset(int argc, char *argv[])
     }
 
     //get the aux tags to be retained
-    if (getauxtags(&resetconf, auxkeepname, auxrejname))
+    if (getauxtags(&resetconf, auxkeepname, auxrejname)) {
+        cleanup(&resetconf);
         return EXIT_FAILURE;
+    }
 
     //get the pg names to be discarded
-    if (getrejectedpgnames(&resetconf, pgrejname))
+    if (getrejectedpgnames(&resetconf, pgrejname)) {
+        cleanup(&resetconf);
         return EXIT_FAILURE;
+    }
 
     //set output file format based on name
     sam_open_mode(outmode + 1, outname, NULL);
```

reset.c
jkbonfield: Ah I see it was originally named `--reject-PG`.  When I asked for `--no-PG` I was referring to the same functionality offered in other subcommands: to not add a new PG record.  So renaming it back to the original name is probably good.

I don't see the purpose however of selecting which individual PG records to keep in a CSV.  It feels complex.  I could perhaps see a specific single PG name which we discard and all subsequent in the chain.  Eg if our headers are:

```
@PG	ID:SCS	PN:HiSeq Control Software	...
@PG	ID:basecalling	PN:RTA	PP:SCS	...
@PG	ID:Illumina2bam	PN:Illumina2bam	PP:basecalling	...
@PG	ID:BamAdapterFinder	PN:BamAdapterFinder	PP:Illumina2bam	...
@PG	ID:BamIndexDecoder	PN:BamIndexDecoder	PP:BamAdapterFinder	...
@PG	ID:pb_cal	PN:predictor_pu	PP:BamIndexDecoder	...
@PG	ID:spf	PN:spatial_filter	PP:pb_cal	...
@PG	ID:bwa	PN:bwa	PP:spf	...
@PG	ID:BamMerger	PN:BamMerger	PP:bwa	...
@PG	ID:SplitBamByReadGroup	PN:SplitBamByReadGroup	PP:BamMerger	...
@PG	ID:bwa_aln	PN:bwa	PP:SplitBamByReadGroup	...
@PG	ID:bwa_aln_1	PN:bwa	PP:bwa_aln	...
@PG	ID:bwa_sam	PN:bwa	PP:bwa_aln_1	...
@PG	ID:Picard_SamFormatConverter	PN:SamFormatConverter	PP:bwa_sam	...
@PG	ID:samtools_fixmate	PN:samtools	PP:Picard_SamFormatConverter	...
@PG	ID:BamMerger_1	PN:BamMerger	PP:samtools_fixmate	...
@PG	ID:AlignmentFilter	PN:AlignmentFilter	PP:BamMerger_1	...
@PG	ID:samtools_sort	PN:samtools	PP:AlignmentFilter	...
@PG	ID:picard_markduplicates	PN:Picard	PP:samtools_sort	...
@PG	ID:ChangeBamHeader	PN:ChangeBamHeader	PP:picard_markduplicates	...
@PG	ID:BamTagStripper	PN:BamTagStripper	PP:ChangeBamHeader	...
@PG	ID:samtools	PN:samtools	PP:BamTagStripper	...
```

Then potentially we'd say `--reject-PG bwa` and everything from that point onwards get thrown away, while retaining the earlier processing that lead to the FASTQ file we input to bwa.

jkbonfield: I should also note that man page states it's using the PN field to match on.  The code looks to match both ID and PN which is perhaps OK, but I note PN is not unique so it would be unclear what "bwa" meant in my above example (it'd match ID:bwa, ID_bwa_aln, ID:bwa_aln_1 and ID:bwa_sam).  We could just say whatever is first I guess, as all descendents should get rejected anyway.

reset.c
```diff
@@ -0,0 +1,858 @@
...
+        else {
+            printf ("Could not open aux reject file!\n");
+            return 1;
+        }
+    }
+
+    if (!config->keepRGs) {
+        //RG lines not present, no point in having RG aux fields either, remove RG
+        khiter_t iter = kh_get(aux_set, config->auxtagsin, TAGNUM("RG"));
+        iter != kh_end(config->auxtagsin) ? kh_del(aux_set, config->auxtagsin, iter) : 0;
```
jkbonfield, line 341: Although valid, I'm rather surprised this doesn't give a warning about "value computed is not used" or similar.  I'm sure compilers used to warn about such things.

It's just `if (iter != kh_end(config->auxtagsin)) kh_del(aux_set, config->auxtagsin, iter);` unless I've misread it.

reset.c
vasudeva8: Done; removes PGs from the 1st match to end on match of PN tag to given option (expects a name instead of earlier csv file).

reset.c
```diff
@@ -0,0 +1,858 @@
...
+*/
+int getcsvpgnames(const char *field, conf_data *config)
+{
+    int ret = 0;
+
+    if (!field || !config || !config->pgout)
+        return -1;
+
+    //add to internal set
+    kh_put(pg_set, config->pgout, strdup(field), &ret);
```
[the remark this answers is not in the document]
vasudeva8: Yes, it is required.

reset.c
vasudeva8: Modified to use command line aux tags as in view instead of csv files. Defaults are not in use.

reset.c
vasudeva8: -i is removed and last parameter is considered as input. if -o is not given, "-" will be used for stdout.

reset.c
```diff
@@ -0,0 +1,593 @@
...
+    }
+
+    //set output file format based on name
+    sam_open_mode(outmode + 1, outname, NULL);
+
+    //open input and output files
+    infile = sam_open(inname, "r");
+    outfile = sam_open_format(outname, outmode, &ga.out);
+    if (!infile || !outfile) {
+        fprintf(stderr, "Could not open %s%s%s", !infile ? inname : "", (!infile && !outfile)? ", " : "", !outfile ? outname : "");
```
jkbonfield, line 554: Missing newline on the end of the fprintf statement.

reset.c
```diff
@@ -0,0 +1,593 @@
...
+        return EXIT_FAILURE;
+    }
+
+    //write output header
+    if (sam_hdr_write(outfile, out_samhdr)) {
+        print_error_errno("reset", "Output header write failed (%d)!\n", errno);
+        sam_hdr_destroy(in_samhdr);
+        sam_hdr_destroy(out_samhdr);
+        return EXIT_FAILURE;
+    }
```
jkbonfield, line 299: Don't feel frightened to use `goto` for error handling.  We're not proud! :-)

While it's questionable in other cases, it can greatly reduce code duplication for cases like this.  Eg move `ret_r` declaratrion to the top of the function and initialise it to `EXIT_FAILURE`, add an `error:` label just before the final cleanup section in this function, and then all the error recovery can be a self-documenting `goto error;` statement.

Obviously this needs to make sure that all memory being freed is initialised to NULL at the start and if appropriate checked for before freeing.  See the plethora of gotos in https://github.com/samtools/samtools/blob/develop/bam_sort.c#L2719-L3101 as an example.

That said - there's nothing at all wrong with what you've written.  Just note we won't complain at use of goto in C when appropriate. :)  (I know it has a bad rap generally!)

reset.c
```diff
@@ -0,0 +1,593 @@
...
+        fprintf(stderr, "Invalid parameters in getPGlines!\n");
+        return 1;
+    }
+
+    if (-1 == (count = sam_hdr_count_lines(in_samhdr, pg))) {
+        fprintf(stderr, "Failed to get PG count!\n");
+        return 1;
+    }
+
+    if (config->pgname && config->pgname[0]) {            //when reject-PG is givern, and is not empty, remove given pg onwards
```
jkbonfield, line 192: Typo: "givern" -> "given"

reset.c
```diff
@@ -0,0 +1,593 @@
...
+
+        // read data
+        cnt++;
+        if (bamdata->core.flag & BAM_FSECONDARY || bamdata->core.flag & BAM_FSUPPLEMENTARY) {
+            continue;
+        }
+
+        //update flags
+        uint16_t flags = bamdata->core.flag & ~BAM_FPROPER_PAIR;    //reset pair info
+        flags |= BAM_FUNMAP | BAM_FMUNMAP;                          //mark as unmapped
```
jkbonfield, line 335: This shouldn't set `BAM_FMUNMAP` if the data was originally not paired (eg single-ended ONT data).

jkbonfield (commented):
Thanks for the review updates.  Much improved.  Only minor comments, and some points for group discussion maybe (stylistic).

It multi-threads well too:
                         
```                                                                      
$ time samtools reset --remove-tag "^RG" ~/scratch/data/novaseq.10m.bam -o /tmp/_.bam
real    0m10.350s
user    0m10.065s 
sys     0m0.272s

$time samtools reset --remove-tag "^RG" ~/scratch/data/novaseq.10m.bam -o /tmp/_.bam -@8
real    0m17.667s
user    2m11.169s
sys     0m1.906s
```

jkbonfield:
One additional thing, which is very much a home goal!

```
samtools view --keep-tag xyz in.bam
main_samview: Error parsing option, auxiliary tags should be exactly two characters long.
... [usage statement cut]...
Direct leak of 40 byte(s) in 1 object(s) allocated from:
    #0 0x7ff150bf07f7 in __interceptor_calloc ../../../../gcc-12.1.0/libsanitizer/asan/asan_malloc_linux.cpp:77
    #1 0x557a96 in kh_init_aux_exists /nfs/users/nfs_j/jkb/work/samtools_master/samtools/sam_utils.h:33
    #2 0x557a96 in parse_aux_list /nfs/users/nfs_j/jkb/work/samtools_master/samtools/sam_utils.c:165=
```

The above is samtools view (hence the home goal comment), but you copied that code to here and it has the same problem.
We need to `kh_destroy` the hash table created when we exit based on command line usage error.  Pretty harmless and no one else spotted it yet in view, but now we know about it we may as well fix it here too.

reset.c
```diff
@@ -0,0 +1,593 @@
...
+
+        //remove aux tags and add bam data to output TODO
+        removeauxtags(bamdata, config);
+        if (0 > (ret_w = bam_set1(outdata, bamdata->core.l_qname - bamdata->core.l_extranul - 1, bam_get_qname(bamdata), flags, -1, -1, 0, 0, NULL, -1, -1, 0, bamdata->core.l_qseq, querydata.s, qualdata.s, bam_get_l_aux(bamdata)))) {
+            print_error_errno("reset", "Failed to set output data (%d)!\n", errno);
+            break;
+        }
+
+        memcpy(bam_get_aux(outdata), bam_get_aux(bamdata), bam_get_l_aux(bamdata));
+        outdata->l_data += bam_get_l_aux(bamdata);  //TODO htslib doesnt have an interface for this?
```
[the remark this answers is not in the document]
vasudeva8: Done

reset.c
```diff
@@ -0,0 +1,593 @@
...
+        }
+        else {
+            //data in read order itself
+            for (i = 0; i < bamdata->core.l_qseq ; ++i) {
+                *sp++ = seq_nt16_str[bam_seqi(bamquery, i)];
+            }
+            memcpy(qp, bam_get_qual(bamdata), bamdata->core.l_qseq);
+        }
+
+        //remove aux tags and add bam data to output TODO
```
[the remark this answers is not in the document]
vasudeva8: Done

reset.c
```diff
@@ -0,0 +1,593 @@
...
+        uint16_t flags = bamdata->core.flag & ~BAM_FPROPER_PAIR;    //reset pair info
+        flags |= BAM_FUNMAP | BAM_FMUNMAP;                          //mark as unmapped
+        flags &= ~BAM_FMREVERSE;                                    //reset mate orientation
+
+        if (0 > ks_resize(&querydata, bamdata->core.l_qseq) ||
+            0 > ks_resize(&qualdata, bamdata->core.l_qseq)) {
+                fprintf(stderr, "Failed to get allocate memory!\n");
+                ret_r = -3;
+                break;
+            }
```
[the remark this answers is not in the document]
vasudeva8: Done.

reset.c
vasudeva8: Corrected.

reset.c
vasudeva8: Done

reset.c
vasudeva8: Updated.

reset.c
vasudeva8: Done.

reset.c
```diff
@@ -0,0 +1,593 @@
...
+            if (ga.nthreads != 0) {
+                usage(stderr);
+                return EXIT_FAILURE;
+            }
+            ga.nthreads = atoi(optarg);
+        }
+        break;
+        // handle standard samtool options like thread count, verbosity...
+        default:
+            if (parse_sam_global_opt(c, optarg, lopts, &ga) == 0)
```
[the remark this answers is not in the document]
vasudeva8: Added -O and free for ga

reset.c
```diff
@@ -0,0 +1,593 @@
...
+        break;
+        case 'p':                           //--reject-PG=<name>
+        {
+            if (resetconf.pgname) {
+                usage(stderr);              //already given!
+                return EXIT_FAILURE;
+            }
+            resetconf.pgname = optarg;
+        }
+        break;
```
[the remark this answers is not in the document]
vasudeva8: Done.

reset.c
```diff
@@ -0,0 +1,593 @@
...
+{
+    fprintf(fp, "Usage: samtools reset [options]\n\
+  -o FILE      output file, optional\n\
+  -x, --remove-tag STR\n\
+               aux tags to be removed, optional\n\
+      --keep-tag STR\n\
+               aux tags to be retained, optional. Equivalent to -x ^STR\n\
+      --reject-PG <name>\n\
+               removes PG line with PN matching to input and succeeding PG lines, optional\n\
+      --no-RG  to have RG lines or not, optional\n");
```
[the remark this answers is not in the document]
vasudeva8: Updated with caps to make it in sync with global options.

reset.c
```diff
@@ -0,0 +1,593 @@
...
+}
+
+/// main_reset - starts the reset of data
+/** @param argc - count of arguments
+ *  @param argv - pointer to array of arguments
+returns 1 on failure 0 on success
+*/
+int main_reset(int argc, char *argv[])
+{
+    static const struct option lopts[] = {
```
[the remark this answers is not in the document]
vasudeva8: Added --no-PG

sam_utils.h
```diff
@@ -0,0 +1,44 @@
+/*  sam_utils.c -- to hold utility functions and types
```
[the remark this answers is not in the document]
vasudeva8: Moved relevant declarations from samtools.h to sam_utils.h and included it in samtools.h

doc/samtools.1
```diff
@@ -813,6 +815,25 @@ samtools samples
...
+samtools reset
+.B -o 
+.I output.sam 
+.RB [ -x / --remove-tag
+.I aux_tags_to_be_removed ]
+.RB [ --keep-tag 
+.I aux_tags_to_be_reatained ]
+.RB [  --reject-PG
+.I pg_name ]
+.RB [ --no-RG ]
```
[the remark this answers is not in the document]
vasudeva8: Done.

reset.c
```diff
@@ -0,0 +1,593 @@
...
+    char *sp = NULL, *qp = NULL;
+    uint8_t *bamquery = NULL, *bamqual = NULL;
+    //get bam data, make updates and dump to output
+    while (0 <= (ret_r = sam_read1(infile, in_samhdr, bamdata)))
+    {
+        sp = NULL; qp = NULL;
+        bamquery = NULL; bamqual = NULL;
+
+        // read data
+        cnt++;
```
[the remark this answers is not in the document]
vasudeva8: Done.

vasudeva8:
commit [4eec82c](https://github.com/samtools/samtools/pull/1767/commits/4eec82ce12b3fe4e1616189dd878ae03e42ead84) have the change for sam_view.c for parse_aux_list leak fix.

Issue #1775, opened 2023-01-04
Reading cram files becomes extremely slow when no reference file is provided. 


y9c:
I wonder if samtools `view` (*also other commands*) automatically use reference file path in the header when reading CRAM files without `--reference` arg? I tried to view a CRAM with a simple command `samtools view test.cram` and it works. But it takes a very long time (~4mins) to finish for the first time. But if I specified `--reference` arg, it will be as fast as 0.2 seconds. 

What is the difference? Is it because samtools are creating some cache for reference file for the first time? But why does this not happen when `--reference` arg is given. 

I am also curious why we need to parse the reference for every command. If I just want to count records number, say `samtools view -c`, why the program still try to parse reference first?

jkbonfield:
Please see http://www.htslib.org/doc/samtools.html#REFERENCE_SEQUENCES for the details on how the reference sequences work for htslib and CRAM.

To answer the question - if you give samtools a `--reference` file then it believes you're not trying to do something silly and it uses that as the reference, therefore it has it already and it can start decoding instantly.  If it didn't supply a reference, then it needs to find it somehow.  It tries a variety of methods, but if none of the ways of finding one locally work then it goes to the internet and downloads it from the EBI via the refget API.  It should cache this locally so the next time it's fast again.

The fastest way of handling references however is to take your fasta files and process them with the `seq_cache_populate.pl` perl script.  See http://www.htslib.org/doc/samtools.html#ENVIRONMENT_VARIABLES for more details.  This produces one file per reference sequence, named after its md5sum.  CRAM will then match the SQ M5 tag to this.  The speed up here over traditional fasta is that all the whitespace has been removed already and we can simply do direct random access into the file (usually via the `mmap` system call, to make it appear to be memory).

For the last question, the CRAM format offers too much flexibility.  The earliest CRAMs produced by cramtools.jar and some versions of scramble mixed multiple types of data together in a CRAM block.   An example in SAM could be "I want to know POS field, but POS and MQUAL" are always stored together so I have to read both.   CRAM is totally flexible in what gets stored where, so even just counting the number of records may mean decoding any part of CRAM, especially given `view -c` can be used with filtering options.  Technically it could just ignore the reference and decode the wrong bases, as we know the count will be throwing them away again after decode, but the library isn't written that way.  It can't decode something wrongly - it's either correct or not at all.  Hence the reference is needed for all operations as there is no way to know ahead of time whether or not it is needed.

[...]

PR #1776, opened 2023-01-04
Add CRAM_OPT_REQUIRED_FIELDS option for view -c.

jkbonfield:
This is a big speed up for CRAM (maybe 5-fold), but it depends on which filtering options are being used.

See comments in #1775.

### DIFF
--- a/sam_view.c
+++ b/sam_view.c
@@ -1253,6 +1253,36 @@ int main_samview(int argc, char *argv[])
         }
     }
 
+    if (settings.is_count) {
+        int rf = SAM_FLAG; // don't want 0, and this is quick
+        if (settings.filter)      rf = INT_MAX; // no easy way to know what is needed
+        if (settings.bed || regs) rf |= SAM_POS | SAM_RNAME | SAM_CIGAR;
+        if (settings.rghash)      rf |= SAM_RGAUX;
+        if (settings.library)     rf |= SAM_RGAUX;
+        if (settings.rnhash)      rf |= SAM_QNAME;
+        if (settings.subsam_frac>0) rf |= SAM_QNAME; // -s is via qname hash
+        if (settings.min_mapQ)    rf |= SAM_MAPQ;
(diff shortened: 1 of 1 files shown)


jkbonfield:
A possible future improvement:  when we're not doing any filtering or range queries, we could in theory just add up the number of sequences listed in the cram container structure.

An example, via the experiment "cram_size" command:

```
$ time samtools view -c ~/scratch/data/9827_2#49.cram
56463236

real	2m6.704s
user	1m45.008s
sys	0m11.274s

$ time samtools cram_size ~/scratch/data/9827_2#49.cram | grep sequences
Number of sequences             56463236

real	0m0.769s
user	0m0.189s
sys	0m0.577s
```

In reality it'd be a lot slower due to I/O.  This file is small and in cache due to the previous operation.  On an earlier copy of the same file it took 28s real time.  It's possible that could be sped up with we added a `cram_skip_block(cram_fd *)` API instead of the existing `cram_read_block()`, or a `cram_next_container` function that used the container size field to work out where to seek to.  Certainly room for improvement, albeit for the most naive and simplistic of counts only.

That same could apply to BAM if we have an index, as it's also part of `idxstats`:

```
$ time samtools idxstats ~/scratch/data/9827_2#49.bam | awk '{a+=$3+$4} END {print a}'
56463236

real	0m0.022s
user	0m0.018s
sys	0m0.004s
```

It's debatable whether `view -c` should be doing a count(view) style operation as that's what you asked, or whether it is justifiable for it to use other methods available.  We could argue that it's useful to have a second more fundamental way to count for checking purposes, but that's no different than a `view|wc -l` anyway.

jkbonfield:
Good spot on sanitize.  I didn't think about the way that worked.  I can redistribute the required fields params around, although tbh it wouldn't have made me do anything different as I did work through all the opts and it was just a failure of me to think about -z that caused me not to add it.

jkbonfield:
Made it work via samview_settings_t and distributed the updates to the options.  Not as neat IMO, but done.

PR #1779, opened 2023-01-16
Speed up optical duplicate checking.

### LINKED ISSUE #1771: Performance issues with high coverage regions in markdup
Using the latest version available:
```
$ samtools version
samtools 1.16.1
Using htslib 1.16
```

```
$ uname -sr
Linux 3.10.0-1160.76.1.el7.x86_64
```
```
$ uname -m
x86_64
```
```
$ gcc --version
gcc (GCC) 12.1.0
```

I'm just trying to run markdup on a bam file, with not many fancy flags:
`samtools markdup -d 2500 -@ 16 in.bam out.bam
`
The command probably works fine and gives good results. But not in a timely manner. It looks like that the time goes really up when it hits regions with a few thousands reads of coverage.
I've been doing some benchmarks and it runs at around 50Kreads/s in my test system, but in regions with high coverage it slows down to 0.3Kreads/s. This happens when most of the positions in the region have between 1000 and 10000 reads starting there.
I suspect that the optical duplicate marking involves a quadratic computation (didn't checked the code), so not sure how much optimization margin there exist. But I know that the processing of the entire bam file did not finish in 14 days. And I'm not even sure if it finished the first chromosome.
It would be nice to have some optimizations there, or at least, an option to set the biggest group to disambiguate, so in case of hitting big duplicate groups, only check for optical duplicates if the size is below the threshold. Not sure how to set the DT tag informing that the duplicate is neither a PCR nor an optical one.

whitwham:
Speed up the optical duplicate checks when there are a lot of duplicates (compared to the number of "originals").

The man page still needs fleshing out but that will have to wait for another PR.

Hopefully should fix #1771, or at least help ameliorate it.

### DIFF
--- a/bam_markdup.c
+++ b/bam_markdup.c
@@ -1,7 +1,7 @@
 /*  bam_markdup.c -- Mark duplicates from a coord sorted file that has gone
                      through fixmates with the mate scoring option on.
 
-    Copyright (C) 2017-2022 Genome Research Ltd.
+    Copyright (C) 2017-2023 Genome Research Ltd.
 
     Author: Andrew Whitwham <aw7@sanger.ac.uk>
 
@@ -95,6 +95,7 @@ typedef struct read_queue_s {
     key_data_t single_key;
     bam1_t *b;
     struct read_queue_s *duplicate;
+    struct read_queue_s *original;
     hts_pos_t pos;
     int dup_checked;
     int read_group;
@@ -1147,6 +1148,8 @@ static int check_chain_against_original(md_param_t *param, khash_t(duplicates) *
     if (!ret && coord_fail)
         ret = coord_fail;
 
+    ori->dup_checked = 1;
+
     return ret;
 }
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
While reviewing this I discovered `-u` doesn't work as `sam_open_mode` overwrites that `0`.  That can be easily fixed with:

```
diff --git a/bam_markdup.c b/bam_markdup.c
index 33aa9a5..fc333c4 100644
--- a/bam_markdup.c
+++ b/bam_markdup.c
@@ -2254,7 +2254,7 @@ static int markdup_usage(void) {
 
 int bam_markdup(int argc, char **argv) {
     int c, ret, bc_name = 0;
-    char wmode[4] = {'w', 'b', 0, 0};
+    char wmode[4] = {'w', 0, 0, 0};
     sam_global_args ga = SAM_GLOBAL_ARGS_INIT;
     htsThreadPool p = {NULL, 0};
     kstring_t tmpprefix = {0, 0, NULL};
@@ -2303,7 +2303,7 @@ int bam_markdup(int argc, char **argv) {
                 }
 
                 break;
-            case 'u': wmode[2] = '0'; break;
+            case 'u': wmode[1] = '0'; break;
             case 1001: param.include_fails = 1; break;
             case 1002: param.no_pg = 1; break;
             case 1003: param.check_chain = 0; break;
@@ -2409,7 +2409,8 @@ int bam_markdup(int argc, char **argv) {
         return 1;
     }
 
-    sam_open_mode(wmode + 1, argv[optind + 1], NULL);
+    strcat(wmode, "b"); // default if unknown suffix
+    sam_open_mode(wmode + strlen(wmode)-1, argv[optind + 1], NULL);
     param.out = sam_open_format(argv[optind + 1], wmode, &ga.out);
 
     if (!param.out) {
```

I've got some data now created by deliberately multiplying up 1 in 10ish reads by 1000fold with similarly generated names, forming a huge number of optical duplicates.  I can't really see any difference in speed though, so I'm unsure which case is being sped up.

whitwham:
Okay, I've added your patch.

PR #1782, opened 2023-01-16
Use TMPDIR environment variable, when looking for a temporary folder.

### LINKED ISSUE #1172: samtools collate should respect TMPDIR in generate_prefix()
samtools 1.10

linux

run samtools collate with TMPDIR set
expected result: temporary files end up in $TMPDIR if set
actual result: generate_prefix() is hardcoded to use /tmp

jkbonfield:
Fixes #1172, replaces #1178.

My changes to 1178 are minimal, to improve corner cases when on systems with larger PID values.

### DIFF
--- a/bamshuf.c
+++ b/bamshuf.c
@@ -567,13 +567,18 @@ char * generate_prefix() {
     snprintf(prefix + ret, PREFIX_LEN - ret, "\\%x", pid);
     return prefix;
 #else
-#  define PREFIX_LEN 64
-    prefix = malloc(PREFIX_LEN);
+    char *tmp_env = getenv("TMPDIR");
+    if (!tmp_env)
+        tmp_env = "/tmp";
+
+    size_t prefix_len = strlen(tmp_env)+20;
+    prefix = malloc(prefix_len);
     if (!prefix) {
         perror("collate");
         return NULL;
     }
-    snprintf(prefix, PREFIX_LEN, "/tmp/collate%x", pid);
+    snprintf(prefix, prefix_len, "%s/collate%x", tmp_env, pid);
+
     return prefix;
 #endif
 }
--- a/bamshuf.c
+++ b/bamshuf.c
@@ -1,7 +1,7 @@
 /*  bamshuf.c -- collate subcommand.
 
     Copyright (C) 2012 Broad Institute.
-    Copyright (C) 2013, 2015-2019 Genome Research Ltd.
+    Copyright (C) 2013, 2015-2019,2023 Genome Research Ltd.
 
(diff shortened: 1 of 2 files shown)


jkbonfield:
I'm expanding this after further testing.

- `samtools sort -o /foo/bar.bam` uses /foo for the temp files, while `collate` does not - it'll still output to /tmp (unless explicitly given a prefix obviously).  This inconsistency should be fixed IMO.  I suspect this was always the intended behaviour given using `-o` is enough to bypass the prefix-is-mandatory check, despite it changing nothing.

- The restriction of having prefix as a mandatory option seems excessive, given we now have support for TMPDIR.

jkbonfield:
I thought I'd made it so it did work.  Maybe I somehow lost that change.  Odd. I'll investigate.

jkbonfield:
Ah sorry I misread what it's doing as the syntax of this command is so alien compared to all others.

If we want stdout we must explicitly specify stdout with `-O`.  This is... odd.  However it's a decision that was made in the past that we cannot undo.  Therefore no `-o` nor `-O` is an error (and always was).

It's tempting to think that prefix has no role here, but it does because pre 934c4931 the -o option didn't exist and prefix was the way of specifying the output filename.  So it's now the combination of all 3 lacking that is an error.

We could argue that `-O` is redundant and it should default to stdout  when no output is specified, but that's a different topic.

bamshuf.c
```diff
@@ -573,7 +587,20 @@ char * generate_prefix() {
         perror("collate");
         return NULL;
     }
-    snprintf(prefix, PREFIX_LEN, "/tmp/collate%x", pid);
+
+    int use_default = 1;
+    char *tmp_env = getenv("TMPDIR");
```
daviesrob, line 592: On my Mac, `TMPDIR` is 49 characters long.  It does fit in PREFIX_LEN, but only just.  Possibly it would be better to measure the length of `TMPDIR` and then allocate a sufficiently long buffer?  Plus, it would make the no TMPDIR easier to handle, by just doing:
```
    if (!tmp_env)
        tmp_env = "/tmp";
    size_t prefix_len = strlen(tmp_env) + 16;
    // ...
```

jkbonfield: Agreed, although what on earth is the Mac doing with such a rediculously long temp dir?

Most of that old cold was in earlier commits, but I've pushed a newer one that can cope with uint32_t max length pid (they may get that high on some clusters perhaps).  No need to check for snprintf return as it can no longer overflow, and even if it does it'll have enough bytes of random prefix by then.

PR #1798, opened 2023-02-21
Take variable initialisation out of for statement.

whitwham:
This is just to appease older versions of C compilers.  Changed some of the variable names to avoid potential shadowing.

### DIFF
--- a/bam_consensus.c
+++ b/bam_consensus.c
@@ -1492,8 +1492,8 @@ int calculate_consensus_gap5(hts_pos_t pos, int flags, int depth,
             continue;
 
 //        fprintf(stdout, "%ld %d %2d %2d\n", pos, k, poly_dist[0][k], poly_dist[1][k]);
-        d1 += (i+1)*poly_dist[0][k];
-        d2 += (i+1)*poly_dist[1][k];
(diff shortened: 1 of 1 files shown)


bam_consensus.c
```diff
@@ -1483,15 +1486,16 @@ int calculate_consensus_gap5(hts_pos_t pos, int flags, int depth,
...
             continue;
 
-//        fprintf(stdout, "%ld %d %2d %2d\n", pos, i, poly_dist[0][i], poly_dist[1][i]);
-        d1 += (i+1)*poly_dist[0][i];
-        d2 += (i+1)*poly_dist[1][i];
-        nd1 += poly_dist[0][i];
-        nd2 += poly_dist[1][i];
+//        fprintf(stdout, "%ld %d %2d %2d\n", pos, k, poly_dist[0][k], poly_dist[1][k]);
+        d1 += (i+1)*poly_dist[0][k];
+        d2 += (i+1)*poly_dist[1][k];
```
jkbonfield, line 1496: `i` was on this line twice: `i+1` and `[i]`.  I don't think the variable needed renaming though as I can't see any shadowing.

PR #1802, opened 2023-03-07
Expand the mpileup man page default filtering.

### LINKED ISSUE #1801: samtools 1.14 mpileup excludes duplicates - not documented behavior
Hi,
samtools 1.14 mpileup excludes 'duplicate' reads, but this is not documented in the official documentation.

In my opinion, this should not be the correct behavior. The correct behavior should be to include all reads, and then the -rf and -ff flags control which reads are considered.

jkbonfield:
Fixes #1801

### DIFF
--- a/doc/samtools-mpileup.1
+++ b/doc/samtools-mpileup.1
@@ -76,6 +76,11 @@ could be specified using \fB-r 20 -l chr20.bed\fR, meaning that the
 index is used to find chromosome 20 and then it is filtered for the
 regions listed in the bed file.
 
+By default secondary alignments, QC failures and duplicate reads will
+be omitted, along with low quality bases and some reads in high depth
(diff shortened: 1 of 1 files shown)


doc/samtools-mpileup.1
```diff
@@ -76,6 +76,11 @@ could be specified using \fB-r 20 -l chr20.bed\fR, meaning that the
 index is used to find chromosome 20 and then it is filtered for the
 regions listed in the bed file.
 
+By default secondary alignments, QC failures and duplicate reads will
+be omitted, along with low quality bases and some reads in high depth
+regions.  See the \fB--ff\fB, \fB-Q\fR and \fB-d\fR options for
```
jmarshall, line 81:
```suggestion
regions.  See the \fB--ff\fR, \fB-Q\fR and \fB-d\fR options for
```

PR #1803, opened 2023-03-08
Allow sorting using minimisers with fwd strand only

armintoepfer:
This feature enables avoiding downstream tools to have special logic for reverse unaligned records. It's in particular complicated when methylation tags or custom tags in native orientation are used. The compression is not as great as using both strands, but acceptable. We prefer ease of use and have this feature as part of the official `samtools sort`. On average, for PacBio Revio HiFi datasets with MM/ML, we reduce file size by 13.8% only using fwd strand and 14.0% using both strands.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -2243,6 +2243,7 @@ typedef struct {
     int error;
     int large_pos;
     int minimiser_kmer;
+    bool try_rev;
 } worker_t;
 
 // Returns 0 for success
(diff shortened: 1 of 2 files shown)


jkbonfield:
The code changes look good to me.

Could you please update the doc/samtools-sort.1 man page too?

armintoepfer:
Done @jkbonfield

jkbonfield:
Rebased, a minor tweak to man page, and merged as c80ab28

PR #1816, opened 2023-03-25
add --duplicate-count option in 'samtools markdup' subcommand to record the original read duplication count(include itself) in a 'dc' tag.

wulj2:
add --duplicate-count option in 'samtools markdup' subcommand to record the original read duplication count(include itself) in a 'dc' tag.

for example, one read b has 9 other reads as its duplicates, then with --duplicate-count turned on, the output of read b will be added an tag dc:i:10 to record that this read is from a cluster of 10 duplication cluster.

this function is useful in some situation that it is important to evaluate the quality of non-duplication reads by it's duplication cluster reads count. for example a read with dc=100 might show more importance with a read with dc=1 in super-high depth sequencing as the dc=1 read might comes from some sequence error rather than a real duplicate.

pleas consider put this option in the samtools markdup subcommand, I carefully check the modification and found it has no effect on the existing computation results and it needs really little extra memory. I also check about 30 BAMS to check the dc value rightness against the do tag and found not any error. however please help me check it again whether I have put things right anyway. thank you very much.


whitwham:
Okay, I will take a look at this properly in the next couple of days.  It will need the man page updating though.

whitwham:
Just to clarify, a dc=1 means the read is not a duplicate and has no duplicates?

wulj2:
> Just to clarify, a dc=1 means the read is not a duplicate and has no duplicates?

yes it is, dc=1 means the read is not a duplicate and has no duplicate reads

wulj2:
> Just to clarify, a dc=1 means the read is not a duplicate and has no duplicates?

what's more, if a read is a duplicate of some original read, then this read will have no dc tag.

whitwham:
Okay, this look like something we can add.

There are a few things.  The option needs adding to the man page.  Maybe with a mention that this only counts primary reads.

The **bam_aux_update_int()** calls are not checking for errors.  I know checks in bam_markdup on the bam_aux_update functions are inconsistent (and I will fix them) but I don't want to make it worse.

You could also add a little test, just so future changes don't break it.

wulj2:
> Okay, this look like something we can add.
> 
> There are a few things. The option needs adding to the man page. Maybe with a mention that this only counts primary reads.
> 
> The **bam_aux_update_int()** calls are not checking for errors. I know checks in bam_markdup on the bam_aux_update functions are inconsistent (and I will fix them) but I don't want to make it worse.
> 
> You could also add a little test, just so future changes don't break it.

Okay，Thank you very much. I have add the test input data and its expected output, the command to generate the expected result is in the commit message. I also update the markdup document items in terms of this new functionality. Please check whether I have done it properly.

whitwham:
Thanks for adding those.

Did you miss committing test.pl?  Otherwise the test will not run.

The data in the tests looks human.  Is this publicly available data?  We don't want to include private data by accident.

wulj2:
> Thanks for adding those.
> 
> Did you miss committing test.pl? Otherwise the test will not run.
> 
> The data in the tests looks human. Is this publicly available data? We don't want to include private data by accident.

Yes the original test data I provide is from human, however it is not some personal sequence data, It is sequencd from the public available Hela cell line by my partner. However, I still modify the test data and use sequence generated by wgsim, hide all genome information in the updated test data. Please recheck it.

PR #1828, opened 2023-04-06
Issue templates: Try to direct conda-related issues to Bioconda

jmarshall:
Use an [issue template config file](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser) to add a “Conda issues” link alongside samtools's issue templates on the “New issue” form. This may help users to report such issues in a more appropriate repository and stop spamming maintainers who are uninterested in conda.

You can see how this looks at https://github.com/jmarshall/samtools/issues/new/choose :—

New issue dialog on jmarshall's repository

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,4 @@
+contact_links:
+  - name:  Conda issues
+    url:   https://github.com/bioconda/bioconda-recipes
+    about: Please report issues specific to the samtools conda package to Bioconda.


jmarshall:
(Nonetheless, considering how many users install samtools via conda, it might be worth one of the maintainers developing an interest in the samtools et al bioconda packages.)

jkbonfield:
Agreed with this change, but will let Rob decide on policy things.

As for bioconda being supported directly by us, I'll personally pass on that one.  I've never had good experiences with conda, just frustration. Lots and lots of frustration!  Maybe someone with more patience to wade through the mire will be willing.  If we are to support it, we could probably do with something in README or INSTALL regarding conda.  Even if it's just a "how to get started with conda" thing, as that in itself is a total minefield (conda, bioconda, anaconda, mamba, miniconda, ...  novice users, including myself, are almost certainly going to start off heading in the wrong direction given the complexity and many pitfalls for the unwary).

jkbonfield:
Shrug.  Maybe it's just me then!  I've occasionally had to totally blow away conda installs and start again.  The last time I had problems conda install was taking *ages*.  I completely forgot about it and discovered it over 24 hours later... still "resolving".  Lol.  Seriously?  How can it be so unbelievably broken?

I've never had this with any other package management system, and I've used a lot.

jmarshall:
I would say you should add similar to the htslib and bcftools repositories, but I see you're not using issue templates over there yet.

whitwham:
No, we were talking about adding them yesterday.  Which, I think, we will.

Issue #1470, opened 2021-07-22
Add option "-g" to samtools view to keep reads with *any* of the specified bitflags


eboyden:
Samtools view -f (require all), -F (remove any), and -G (remove all) are useful, but there is no option to "require any" without resorting to expressions.  This would be useful if e.g. I want to keep all FR and RF read pairs based on the bitflags, and discard all FF and RR read pairs; I need bitflag 16 OR bitflag 32, but not both and not neither.  RR pairs could be removed with "-G 48", but there is no means to remove FF pairs.  This could be enabled with "-g 48 -G 48" to require either 16 or 32 but disallow both.

jkbonfield:
Right now we have:

```
-f  keep if all
-F keep if none
-G exclude if all
```

The obvious missing one is `-g exlude if none` (which I believe is the same as "keep if any"), but note sadly we're also swapping case here as f/F is all/none while we'd have g/G being none/all.  Argh!  (A classic "I wouldn't start from here" scenario!) 

However rather than keep adding to an already lengthy list of options and risk confusion over the meaning of upper/lower case, is there a specific reason why you don't wish to use the filter expression language?

eboyden:
Yeah agreed, though you could choose to look at it as:
```
-f keep if all
-g keep if any
-F exclude if any (opposite of -f)
-G exclude if all (opposite of -g)
```
The manual says that -G is the opposite of -f, but I don't really agree, I think logically -f/-F are opposites, as would be -g/-G.  Revising the descriptions might also help minimize confusion.

No specific reason not to use the filter expression language, other than that it's complicated to explain to others when providing advice on how to run their pipelines.  Ideally expression-based filtering should be reserved for complicated needs, not simple needs for which an option doesn't currently exist (but could be added, hopefully).

valeriuo:
All the filtering options have two parts: the matching part (**all** | **none**) and an action part (**keep** | **exclude**).
`-G` is the opposite of `-f` in the action part (exclude vs keep), but they both use the same matching rule (all).

jmarshall:
James pointed at the discussion in https://github.com/samtools/samtools/pull/1442#issuecomment-871244137 and the rest of that PR. As the table in that comment shows, `mpileup` and `coverage` use `--rf`/`--incl-flags` for the proposed filtering operation.

[...]

PR #1842, opened 2023-05-06
Consistent, more comprehensive flag filtering for samtools fasta/fastq

DevangThakkar:
This PR is an extension of https://github.com/samtools/samtools/pull/1508,  for the `fasta`/`fastq` command - similar to https://github.com/samtools/samtools/pull/1718 for `depth`. This should allow exporting reads that match any/either of the bits present in the FLAG to a fastq file. I have followed the conversation in https://github.com/samtools/samtools/issues/1470 and https://github.com/samtools/samtools/pull/1442#issuecomment-871244137, and followed the recommendations for variable naming.

Summary of changes:
- Added the `--rf/incl[ude]-flags` option
- Allowed for `--rf/incl[ude]-flags` to be a longoption-only argument using the LONGOPT macro
- Added aliases for the existing `-F` option (`--excl[ude]-flags`)
- Refactored the assignment of the value for `-F` for conciseness and better readability
- Updated documentation for `fasta`/`fastq`

### DIFF
--- a/bam_fastq.c
+++ b/bam_fastq.c
@@ -64,8 +64,12 @@ static void bam2fq_usage(FILE *to, const char *command)
 "  -o FILE      write reads designated READ1 or READ2 to FILE\n"
 "               note: if a singleton file is specified with -s, only\n"
 "               paired reads will be written to the -1 and -2 files.\n"
-"  -f INT       only include reads with all  of the FLAGs in INT present [0]\n"       //   F&x == x
-"  -F INT       only include reads with none of the FLAGS in INT present [0x900]\n"       //   F&x == 0
+"  -f, --require-flags INT\n"
+"               only include reads with all  of the FLAGs in INT present [0]\n"       //   F&x == x
+"  -F, --excl[ude]-flags INT\n"
+"               only include reads with none of the FLAGs in INT present [0x900]\n"   //   F&x == 0
+"      --rf, --incl[ude]-flags INT\n"
+"               only include reads with any  of the FLAGs in INT present [0]\n"       // !(F&x == 0)
 "  -G INT       only EXCLUDE reads with all  of the FLAGs in INT present [0]\n"       // !(F&x == x)
 "  -n           don't append /1 and /2 to the read name\n"
 "  -N           always append /1 and /2 to the read name\n",
@@ -132,7 +136,7 @@ typedef struct bam2fq_opts {
     char *fnr[3];
     char *fn_input; // pointer to input filename in argv do not free
     bool has12, has12always, use_oq, copy_tags, illumina_tag;
-    int flag_on, flag_off, flag_alloff;
+    int flag_on, flag_off, flag_alloff, flag_anyon;
     sam_global_args ga;
     fastfile filetype;
     int def_qual;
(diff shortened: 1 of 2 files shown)


bam_fastq.c
```diff
@@ -64,8 +64,12 @@ static void bam2fq_usage(FILE *to, const char *command)
 "  -o FILE      write reads designated READ1 or READ2 to FILE\n"
 "               note: if a singleton file is specified with -s, only\n"
 "               paired reads will be written to the -1 and -2 files.\n"
-"  -f INT       only include reads with all  of the FLAGs in INT present [0]\n"       //   F&x == x
-"  -F INT       only include reads with none of the FLAGS in INT present [0x900]\n"       //   F&x == 0
```
DevangThakkar, line 68: added longoption arguments for filtering flags for clarity, consistent with [samtools view](https://github.com/samtools/samtools/blob/develop/sam_view.c#L1471).

bam_fastq.c
```diff
@@ -64,8 +64,12 @@ static void bam2fq_usage(FILE *to, const char *command)
...
+"  -f, --require-flags INT\n"
+"               only include reads with all  of the FLAGs in INT present [0]\n"       //   F&x == x
+"  -F, --excl[ude]-flags INT\n"
+"               only include reads with none of the FLAGs in INT present [0x900]\n"   //   F&x == 0
+"      --rf, --incl[ude]-flags INT\n"
+"               only include reads with any  of the FLAGs in INT present [0]\n"       // !(F&x == 0)
```
DevangThakkar, line 72: Added --rf/--incl[ude]-flags arguments

bam_fastq.c
```diff
@@ -176,6 +180,9 @@ static void free_opts(bam2fq_opts_t *opts)
     free(opts);
 }
 
+// Make mnemonic distinct values for longoption-only options
+#define LONGOPT(c)  ((c) + 128)
+
```
DevangThakkar, line 185: Added LONGOPT since --rf does not have a user facing single character option

bam_fastq.c
```diff
@@ -223,14 +237,11 @@ static bool parse_opts(int argc, char *argv[], bam2fq_opts_t** opts_out)
             case '2': opts->fnr[2] = optarg; break;
             case 'o': opts->fnr[1] = optarg; opts->fnr[2] = optarg; break;
             case 'f': opts->flag_on |= strtol(optarg, 0, 0); break;
-            case 'F':
-                if (!flag_off_set) {
-                    flag_off_set = 1;
-                    opts->flag_off = 0;
-                }
-                opts->flag_off |= strtol(optarg, 0, 0);
-                break;
```
DevangThakkar, line 232: I believe this chunk is unnecessary and can be replaced by a simpler statement as shown below for better readability. `flag_off_set` has a default of zero so the `if` loop is always entered and `opts->flag_off` set to 0. This can be replaced by the line below.

This change is not essential to the objective of this PR.

bam_fastq.c
```diff
@@ -580,6 +592,7 @@ static inline bool filter_it_out(const bam1_t *b, const bam2fq_state_t *state)
 {
     return ((b->core.flag&(state->flag_on)) != state->flag_on // or reads indicated by filter flags
         ||  (b->core.flag&(state->flag_off)) != 0
+        ||  (((b->core.flag&(state->flag_anyon)) == 0) && (state->flag_anyon != 0))
```
DevangThakkar, line 595: `--rf` would filter out reads if none of the bits match, given that there is a nonzero value for `state->flag_anyon`.

PR #1855, opened 2023-06-01
Apply `samtools fastq --input-fmt-option ...` settings

jmarshall:
The `samtools fastq` usage claims it supports `--input-fmt-option`, but any options specified are not actually applied to the input file. This one-line PR fixes that. Possibly worthy of a _NEWS_ entry.

(I have not audited other commands to see if any others have a similar problem.)

### DIFF
--- a/bam_fastq.c
+++ b/bam_fastq.c
@@ -423,7 +423,7 @@ static bool init_state(const bam2fq_opts_t* opts, bam2fq_state_t** state_out)
     state->hstdout = NULL;
     state->compression_level = opts->compression_level;
 
-    state->fp = sam_open(opts->fn_input, "r");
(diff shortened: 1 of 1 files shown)


jmarshall:
I have now audited the other commands: the only other problematic one is `cram-size`. Its `SAM_OPT_GLOBAL_OPTIONS` declares that it enables only `--input-fmt-option`, but nothing is done with any input format settings. I'm not sure whether it actually needs any; it may be best just to remove the `sam_global_args` stuff from this one altogether.

jkbonfield:
Indeed I think cram-size wouldn't need the input-fmt-options.  It was probably a cut and paste error to have that in the options.

PR #1864, opened 2023-06-27
sample name for samtools stats

lindenb:
Hi all,
this is a proposal to add `sample name` in the section `SN` of samtools stats. My motivation is to help tools like `multiqc` to find the sample name(s) in the output.

```
./samtools stats in.bam | grep "sample name" -C1
# Summary Numbers. Use `grep ^SN | cut -f 2-` to extract this part.
SN	sample name:	BruceWayne
SN	raw total sequences:	892	# excluding supplementary and secondary reads
```

(no read group)
```
$ samtools view -h in.bam | grep -v '@RG'  | ./samtools stats |  grep "sample name" -C1
# Summary Numbers. Use `grep ^SN | cut -f 2-` to extract this part.
SN	sample name:	.	# no sample was found in read groups.
SN	raw total sequences:	1998	# excluding supplementary and secondary reads
```

if it's ok with you, I'll later fix the tests and the doc.

Thanks !

### DIFF
--- a/stats.c
+++ b/stats.c
@@ -240,6 +240,7 @@ typedef struct
     // Auxiliary data
     double sum_qual;                // For calculating average quality value
     void *rg_hash;                  // Read groups to include, the array is null-terminated
+    void* samples_hash;             // Samples name in the BAM, the array is null-terminated
 
     // Split
     char* split_name;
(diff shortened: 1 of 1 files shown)


daviesrob:
This looks like a good idea in principle, although I'm not sure if the information really belongs in the `SN` section which currently only reports numeric data.  Also, having multiple lines with the same key may confuse tools that expect the keys to be unique (e.g. MultiQC's stats parser appears to [put the data in a map based on the key](https://github.com/ewels/MultiQC/blob/7129bd10420149c032e380bef57f9c9ca0044d33/multiqc/modules/samtools/stats.py#L31)).  Possibly the list should go in the comment section at the start, or maybe it's own section with a new prefix.

It's possible to filter the samples contributing to the output by using the `-I` and/or `-S RG` command-line options.  It might be good if the reported list reflected these choices, so it's easy to see which samples were actually included.

PR #1868, opened 2023-06-28
Fix samtools reheader CRAM output version.

### LINKED ISSUE #1866: Non-in-place reheadering a CRAM v2.1 file produces a broken v3.0 file

Tested with both current **develop** and the 1.17 release.

Primarily tested on macOS.

Doing a non-in-place `samtools reheader` of a CRAM v2.1 file produces a CRAM file that cannot be indexed (here demonstrated using the test files from `test_reheader()`):

```
$ samtools reheader test/reheader/hdr.sam test/dat/view.001.tmp.v21.cram > bork.cram
$ samtools index bork.cram
[E::cram_index_container] CRAM slice offset 582 does not match landmark 0 in container header (578)
samtools index: failed to create index for "bork.cram"
```

Perhaps relatedly, it turns out that _bork.cram_ identifies itself as a v3.0 CRAM file — which is presumably incorrect:

```
$ htsfile test/dat/view.001.tmp.v21.cram bork.cram
test/dat/view.001.tmp.v21.cram:	CRAM version 2.1 compressed sequence data
bork.cram:	CRAM version 3.0 compressed sequence data
```

v2.1 files are fairly obsolete, so if it would be difficult to fix this so that out-of-place `samtools reheader` of a v2.1 CRAM file instead produces a valid indexable v2.1 CRAM file, a sufficient alternative might be for out-of-place reheadering to fail with an error message when given a v2.1 file.

jkbonfield:
For non-3.0 CRAMs we need to remember to set the output version in the header.  This also changes the container structures for CRAM 2.1.

Fixes #1866

### DIFF
--- a/bam_reheader.c
+++ b/bam_reheader.c
@@ -127,6 +127,11 @@ int cram_reheader(cram_fd *in, sam_hdr_t *h, const char *arg_list, int no_pg)
     if (!h)
         return ret;
 
+    // Match output version number with input file.
+    char vers[99];
+    sprintf(vers, "%d.%d", cram_major_vers(in), cram_minor_vers(in));
+    cram_set_option(out, CRAM_OPT_VERSION, vers);
+
     // Attempt to fill out a cram->refs[] array from @SQ headers
     sam_hdr_t *cram_h = sam_hdr_dup(h);
     if (!cram_h)


jmarshall:
Thanks — confirmed that this fixes the problem on the test data I was using.

jkbonfield:
Thanks for the validation.

PR #1873, opened 2023-06-30
Documentation formatting etc fixes [minor]

jmarshall:
Add `.bgz` to fasta/fastq's list of recognised compressed file extensions, as per `sam_open_z()`.
Format `CMD` as a metasyntactic variable.
Fix typo.


jkbonfield:
Thanks John

PR #1877, opened 2023-07-12
fix various typos caught by lintian parsers

emollier:
This patch addresses a couple of typos, plus a few malformations in manual pages, caught by lintian while packaging the latest samtools version of the day for Debian.

I was overdue to forward this upstream.
Thanks John Marshall for the reminder!


emollier:
The second commit was actually a patch carried over two years.  Better late than never they say.

whitwham:
Thank you, lintian sounds like a useful little tool.

PR #1909, opened 2023-08-21
Check fclose(stdout) at the end of main()

jmarshall:
Companion pull request to samtools/htslib#1665. If that PR's approach is accepted, then `hts_open("-", "w")` / `hts_close()` no longer actually closes stdout. Close it explicitly at the end of `main()` so there is an opportunity to detect I/O errors in previously-uncommitted writes.

Ignore EBADF as other code may have already closed stdout, e.g., either particular subcommands or when (dynamically) linked against an older version of HTSlib.

### DIFF
--- a/bamtk.c
+++ b/bamtk.c
@@ -303,5 +303,14 @@ int main(int argc, char *argv[])
         fprintf(stderr, "[main] unrecognized command '%s'\n", argv[1]);
         return 1;
     }
+
+    // For subcommands that may have produced substantial output on stdout,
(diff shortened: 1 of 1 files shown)


jkbonfield:
Thanks.  Looks good to me and I approve of the placement in `main` as a catch all for all sub-commands.

The `EBADF` check while mentioning prior closure in other code, also covers using samtools against an earlier, compatible, htslib release (eg this PR + htslib-1.18)

PR #1913, opened 2023-09-12
mpileup manual updated with samples

vasudeva8:
Updated man page for mpileup to showcase a few usages, related to flags, explicit index and on getting most of the alignments on mpileup output. (#1801)

### DIFF
--- a/doc/samtools-mpileup.1
+++ b/doc/samtools-mpileup.1
@@ -76,6 +76,7 @@ could be specified using \fB-r 20 -l chr20.bed\fR, meaning that the
 index is used to find chromosome 20 and then it is filtered for the
 regions listed in the bed file.
 
+Unmapped reads are not considered and are always discarded.
 By default secondary alignments, QC failures and duplicate reads will
 be omitted, along with low quality bases and some reads in high depth
 regions.  See the \fB--ff\fR, \fB-Q\fR and \fB-d\fR options for
@@ -283,8 +284,8 @@ none of the mask bits set.  Hence this does not override the
 .TP
 .BI --ff,\ --excl-flags \ STR|INT
 Filter flags: skip reads with any of the mask bits set.  This defaults
-to UNMAP,SECONDARY,QCFAIL,DUP.  The option is not accumulative, so
-specifying e.g. \fB--ff UNMAP,QCFAIL\fR will reenable output of
+to SECONDARY,QCFAIL,DUP.  The option is not accumulative, so
+specifying e.g. \fB--ff QCFAIL\fR will reenable output of
 secondary and duplicate alignments.  Note this does not override the
 \fB--incl-flags\fR option.
 .TP
@@ -482,7 +483,42 @@ The
 .B -E
 option can be used to make it ignore the contents of the BQ:Z tag and
 force it to recalculate the BAQ scores by making a new alignment.
+.PP
+.SH EXAMPLES
+Using range:
+With implicit index files in1.bam.<ext> and in2.sam.gz.<ext>,
+.EX 2
+samtools mpileup in1.bam in2.sam.gz -r chr10:100000-200000
+.EE
+With explicit index files,
+.EX 2
+samtools mpileup in1.bam in2.sam.gz idx/in1.csi idx/in2.csi -X -r chr10:100000-200000
+.EE
+With fofn being a file of input file names, and implicit index files present with inputs,
+.EX 2
+samtools mpileup -b fofn -r chr10:100000-200000
+.EE
+Using flags:
+To get reads with flags READ2 or REVERSE and not having any of SECONDARY,QCFAIL,DUP,
+.EX 2
+samtools mpileup --rf READ2,REVERSE in.sam
+.EE
+or
+.EX 2
+samtools mpileup --rf 144 in.sam
+.EE
+To get reads with flag SECONDARY,
+.EX 2
+samtools mpileup --rf SECONDARY --ff QCFAIL,DUP in.sam
+.EE
+Using all possible alignmentes:
+To show all possible alignments, either of below two equivalent commands may be used,
+.EX 2
+samtools mpileup --count-orphans --no-BAQ --max-depth 0 --fasta-ref ref_file.fa \\
+--min-BQ 0 --excl-flags 0 --disable-overlap-removal in.sam
 
+samtools mpileup -A -B -d 0 -f ref_file.fa -Q 0 --ff 0 -x in.sam
+.EE
 .SH AUTHOR
 .PP
 Written by Heng Li from the Sanger Institute.


jkbonfield:
Rather curious that the old man page had "See EXAMPLES section for sample of usage" under the `-X` option, but no examples.  

I see this vanished with 9de488aaf4c959b3a8cfefb4c4cfa8282f910020 when we culled the VCF output from mpileup, but one of the examples there was:

```
       o Include customized index file as a part of arguments.

           samtools mpileup [options] -X /data_folder/in1.bam [/data_folder/in2.bam [...]] /index_folder/index1.bai [/index_folder/index2.bai [...]]
```

This is what the `-X` was referring to, and I was apparently over-zealous in removing now pointless examples.

Thank you for adding an example of `-X` back.  It's not the same one, but is sufficient.  I have some comments on the text, but I'll do that inline in the review.

doc/samtools-mpileup.1
```diff
@@ -482,7 +483,43 @@ The
...
+.B samtools mpileup --rf READ2,REVERSE in.sam
+.R or
+.B samtools mpileup --rf 144 in.sam
+.PP
+to get reads with flag SECONDARY,
+.IP
+.B samtoos mpileup --rf SECONDARY --ff QCFAIL,DUP in.sam
+.R or
+.B samtools mpileup --rf 256 --ff 1536 in.sam
+.PP
```
jkbonfield, line 516: This applies to all other examples too, but just cherry picking one to discuss.

All the examples in other man pages, including the old examples here before we removed them, use `.EX` and `.EE` to mark text as example, rather than displaying in bold.  This also makes it fixed-width font on the web. See http://www.htslib.org/doc/samtools-view.html#EXAMPLES for examples.

We don't appear to have consistent indenting, but it looks like we're mostly doing `.EX 2` or `.EX 4`.  The old mpileup man was `.EX 2` so maybe just stick with that.  It's sufficient either way.  

I also fixed typo of "samtoos" above, removed the second "or" example for -rf 256 etc as we don't need to labour the point of using symbolic strings or numerics twice and less to read is better IMO.  Finally I prefer new paragraphs starting with a capital letter.  

So for example:

```
.PP
Using flags:
to get reads with flags READ2 or REVERSE and not having any of SECONDARY,QCFAIL,DUP,
.EX 2
samtools mpileup --rf READ2,REVERSE in.sam
.EE
or
.EX 2
samtools mpileup --rf 144 in.sam
.EE
To get reads with flag SECONDARY,
.EX 2
samtools mpileup --rf SECONDARY --ff QCFAIL,DUP in.sam
.EE
```

This renders in the terminal as:

```
       Using  flags:  to  get  reads  with  flags  READ2  or  REVERSE  and  not having any of SEC-
       ONDARY,QCFAIL,DUP,

         samtools mpileup --rf READ2,REVERSE in.sam

       or

         samtools mpileup --rf 144 in.sam

       To get reads with flag SECONDARY,

         samtools mpileup --rf SECONDARY --ff QCFAIL,DUP in.sam
```

*Some* other man pages have bullet point sections (eg http://www.htslib.org/doc/samtools-view.html#EXAMPLES mentioned already) while others do not (http://www.htslib.org/doc/samtools-fasta.html#EXAMPLES).   Given we don't really have a specific style, I'll leave that up to you, but if you feel it helps with the structure then take a look at samtools-view.1 for how this is done.  IMO it looks a bit ugly in the text terminal, but can still be helpful to break up lots of separate examples and delineate where one finishes and the next starts if we have multiple paragraphs.

doc/samtools-mpileup.1
```diff
@@ -482,7 +483,43 @@ The
...
+with implicit index files in1.bam.<ext> and in2.sam.gz.<ext>,
+.IP
+.B samtools mpileup in1.bam in2.sam.gz -r <name>:<range>
+.PP
+with explicit index files,
+.IP
+.B samtools mpileup in1.bam in2.sam.gz <in1 index path> <in2 index path> -X -r <name>:<range>
+.PP
+with fofn having input file names, with implicit index files present along with inputs,
```
jkbonfield, line 499: Nit-picky: maybe be more explicit with "fofn" by adding "file of", eg "with fofn being an file of input filenames", just to reinforce what the acronym stands for.

doc/samtools-mpileup.1
```diff
@@ -482,7 +483,43 @@ The
...
+.PP
+to get reads with flag SECONDARY,
+.IP
+.B samtoos mpileup --rf SECONDARY --ff QCFAIL,DUP in.sam
+.R or
+.B samtools mpileup --rf 256 --ff 1536 in.sam
+.PP
+Using all possible alignmentes:
+.PP
+i.e. counting orphans, avoiding BAQ calculation, using max depth, min BAQ as 0, exclude flag as 0 and overlap removal disabled
```
jkbonfield, line 519: BAQ is somewhat orthogonal to showing all alignments, but I think it is a reasonable thing to include as typically you're wanting all data because you'd like it in the rawest possible form, which also means not munging qualities.

Note though `-Q 0` is "using a minimum base quality of 0", not "min BAQ as 0".  That's a small distinction, but we've just stated with `-B` that we're avoiding BAQ so people may think the `-Q` then has no effect when phrased as-is.

Also as this is a proper man page we ought to use "minimum" and "maximum" instead of "min" and "max".

Similarly "exclude flag as 0" could be "exclude flag as 0 (none excluded)".
Finally, given this is a long list of things, maybe the Oxford comma here helps clarify, so ", and overlap removal disabled".

The alternative though to describing the command line and then showing the command line, would be to just use the long-opts, maybe with the short one underneath it.  Eg:

```
To show all possibile alignments, these two equivalent commands may be used.
These also disables BAQ so the quality values are presented in their original form.
.EX 2
samtools mpileup --count-orphans --no-BAQ --max-depth 0 --excl-flags 0 \\
    --disable-overlap-removal --min-BQ 0 in.bam

samtools mpileup -A -B -d 0 --ff 0 -x -Q 0 in.bam
.EE
```

doc/samtools-mpileup.1
```diff
@@ -482,7 +483,43 @@ The
...
 force it to recalculate the BAQ scores by making a new alignment.
-
+.PP
+.SH EXAMPLES
+.PP
+Using range:
+.PP
+with implicit index files in1.bam.<ext> and in2.sam.gz.<ext>,
+.IP
+.B samtools mpileup in1.bam in2.sam.gz -r <name>:<range>
```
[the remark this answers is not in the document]
vasudeva8: done.

doc/samtools-mpileup.1
```diff
@@ -482,7 +483,43 @@ The
...
+.PP
+with explicit index files,
+.IP
+.B samtools mpileup in1.bam in2.sam.gz <in1 index path> <in2 index path> -X -r <name>:<range>
```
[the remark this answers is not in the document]
vasudeva8: done.

doc/samtools-mpileup.1
vasudeva8: done.

doc/samtools-mpileup.1
vasudeva8: done.

PR #1917, opened 2023-09-20
Use the spec-defined PL values in man page example

jmarshall:
As noted by @zaeleus on Slack, an example in samtools ought to be using the correct specification-defined PL values.

Moreover the Perl command shown interpolates `@RG` as an undefined array and produces lines starting `<tab>ID:…`; instead simplify by using `printf(1)` (unfortunately… if only `echo` could portably express this with ease…) to avoid any such issues.


jkbonfield:
Thanks both

Issue #1824, opened 2023-04-04
Feature request: read length plot for samtools stat


mp15:
Previously we have not included a plot that gives us a histogram of read lengths because most of our reads have been Illumina reads with a fixed cycle length per run. With 3rd generation platforms read lengths now vary considerably, plotting this as part of samtools stats would be helpful. This feature should be able to be switched on or off manually but if neither of these is supplied it should detect known 3rd gen platforms in the RG tag.

jkbonfield:
I like this idea.  If it was always on it wouldn't take up much room for fixed-cycle instruments anyway, and it ought to be fast enough.   This sort of thing is normally done at the fastq level by e.g. FastQC, but with more instruments giving uBAM or even fully aligned BAM/CRAM as output format then a tool to run off BAMs makes sense.

I assume it'd be for primary alignments only and full unclipped length regardless of CIGAR, so it reports the instrument / library data rather than alignment statistics.  For completeness we'd probably want to support read-pair protocols too and separate READ1 from READ2 (or possibly the sum of the two for protocols where the pair is implemented by library prep and an embedded adapter to remove post single-ended sequencing).

jkbonfield:
I'd think we need some way of adjusting the bin size so small sizes go in increments of 1 and larger sizes start to be binned in bigger widths, as we may have reads 100Kb+ long.  That would also mean potentially columns showing the total count within that bin and the bin-averaged count, so a 100 wide bin with 2730 reads in it would be an average of 27.3 per length.  That makes it easy to plot still.  I think so anyway.  Would have to mock something up to see...

An ONT example without any binning (logscale x):

![image](https://github.com/samtools/samtools/assets/2210525/3eea79c6-12c2-4513-8d2f-8bc0b6e11235)

An ONT example with log-binning via

```
awk '$1>1 {b=int(log($1)/log(10)*10);bin[b]++} END {for (i in bin) {st=10**(i/10);en=10**((i+1)/10);print bin[i]/(en-st),bin[i],st,en}}' _l|sort -k 3n

gnuplot: plot "_b" using 3:1 with boxes
```

![image](https://github.com/samtools/samtools/assets/2210525/3e38597c-59f3-4e22-af1f-ef31357a8533)

The spike at length 62 vanished though.  I guess a clever algorithm could look for outliers and make specific boxes for that size?

erboone:
[...]

PR #1922, opened 2023-09-27
Bamstats read length histogram plot feature.

erboone:
Implementation of feature suggested in #1824. After feedback on the presentation of the data, I've cleaned up the implementation and produced this version.  Let me know if this looks satisfactory. Attached below are two examples of output.
read_length2
read_length

### DIFF
--- a/misc/plot-bamstats
+++ b/misc/plot-bamstats
@@ -25,6 +25,8 @@
 use strict;
 use warnings;
 use Carp;
+use POSIX qw/floor/;
+use List::Util qw(max);
 
 my $opts = parse_params();
 parse_bamcheck($opts);
@@ -39,6 +41,7 @@ plot_coverage($opts);
 plot_mismatches_per_cycle($opts);
 plot_indel_dist($opts);
 plot_indel_cycles($opts);
+plot_read_len_hist($opts);
 create_html($opts);
 
 exit;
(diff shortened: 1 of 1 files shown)


misc/plot-bamstats
```diff
@@ -1204,6 +1207,119 @@ sub plot_indel_cycles
...
+    # data into a single GNUplot file.
+    open(my $rd_fh, '>', "read_len_rawdata.txt");
+    open(my $hd_fh, '>', "read_len_histdata.txt");
+    
+    # Create gnuplot command file options
+    my $min_x = $rl[0][0];
+    my $max_x = $rl[-1][0];
+    # Gnuplot commands for normalized binned data
+    print $fh qq[
+        set y2tic
```
jkbonfield, line 1243: We could do with `set ytics nomirror` here too as the right Y axis has two sets of tics combined on it.

misc/plot-bamstats
```diff
@@ -1204,6 +1207,119 @@ sub plot_indel_cycles
...
 
+sub plot_read_len_hist
+{
+    my ($opts) = @_;
+
+    my @rl = get_values($opts, 'RL');
+    if ( !@rl ) { return; }
+
+    # TODO: Add modified version for paired end reads.
+    my $is_paired = get_value($opts,'SN','reads paired:');
```
jkbonfield, line 1218: Understood that this is a TODO, and OK with that, but maybe we should comment out the `$is_paired` assignment too as it's unused (for now).

misc/plot-bamstats
```diff
@@ -1204,6 +1207,119 @@ sub plot_indel_cycles
...
+        set autoscale x
+        set logscale x
+        set xrange[$min_x: $max_x]
+        set xtics rotate by -45 scale 0
+        set xlabel "Read Length"
+        set boxwidth .8 relative
+        set title "$$args{title}" noenhanced 
+        set title offset -15,-.5 
+        set key at graph 1, 1.15
+        plot "read_len_rawdata.txt" using 1:2 lt rgb "orange" with boxes fs solid .7 title 'unbinned', "read_len_histdata.txt" using 4:3 lt rgb "blue" with boxes title 'normalized bins(count/bin width)' axis x1y2
```
jkbonfield, line 1259: It took a while before I understood why we had two plots, and I've read the discussion :-).  Then I remembered it's because we have the normalised-by-bin-size and unnormalised versions.  I see these in the key, so there is a hint if you look, but it may not be that obvious to many.  

Without the background knowledge I'd simply be asking myself "normalised how?"!
Sadly it appears plot-bamstats has no man page or documentation, so people have to figure it out.  This isn't your fault - it's just where we're starting from.

However perhaps we can get a comment in the source code so if people read the code to understand what these two plots are and what the thing we're normalising over is, then it'll help.

jkbonfield (commented):
I like the general layout of the plot.  It's a marked improvement on the rest of plot-bamstats which has horrendous problems with long sequences.  I'm guessing it (and `samtools stats`) were both written in an era when 75bp reads was the normal.  Hence it uses terms like "cycle" and it makes assumptions that recording data for every observed length is sane, and that we can also plot them without producing a smear of values on the axis.

None of that is your problem, but thank you for being a step above what has come before.  Fixing the rest of stats and plot-bamstats is a far bigger project for another day.

I have a few very minor comments that should be simple enough to address hopefully.

misc/plot-bamstats
```diff
@@ -1204,6 +1207,119 @@ sub plot_indel_cycles
...
+    };
+    my $inv_log_bin = sub {
+        my $x = $_[0];
+        return floor(10**($x/10));
+    };
+
+    # Open files to store data: this is more elegant than copying four sets of 
+    # data into a single GNUplot file.
+    open(my $rd_fh, '>', "read_len_rawdata.txt");
+    open(my $hd_fh, '>', "read_len_histdata.txt");
```
jkbonfield, line 1236: Maybe these two files should be using `$$opts{prefix}`.  I know they're temporary files, but having fixed names means they overwrite each other if we're attempting to plot multiple stats files simultaneously.  Plus it's possible our current working directory isn't somewhere we have write access too (unlike presumably the prefix location).

misc/plot-bamstats
erboone: I've added some explanation to both of the graphs to explain how they are derived, what they show, and the logic behind why they were included. The lack of documentation on plot-bamstats is disappointing. Do you think that writing some documentation for it would be useful?

misc/plot-bamstats
erboone: Done

misc/plot-bamstats
erboone: Good catch: modified it.

jkbonfield:
Thanks.  I took the liberty of tidying up the axis slightly more, then squashed and pushed back prior to merging (easiest done that way).

misc/plot-bamstats
jkbonfield: I went ahead and merged as-is.  Documentation would clearly be useful, but there is quite a lot that's missing from plot-bamstats and a considerable amount to document.  If you wished to write documentation for it then we'd be grateful, but it's different to this issue and PR and it shouldn't delay the merging of the new feature.  So if you wanted to do this, please let us know so we can discuss formats, and then create a new PR.  Totally understood if you have other things to do too.  (We do which is why it's never been done!)

Either way, thanks.

PR #1934, opened 2023-11-06
Add samtools split -d TAG option

### LINKED ISSUE #1758: split: use custom tag
Currently `samtools split` only splits by tag `RG`, however there may be situations where you'd want to split by a different tag. Example: after soft-demultiplexing an experiment, you align all the reads using STAR, which overwrites tag `RG`.

Add a new argument `-t ##` to split by the value of the specified tag. Support only tags with integer or string type.

Alternatively: Add a new argument to split by an expression `-e`. In this formulation, `samtools split -e '[XX]'` would be equivalent to `samtools split -t XX`, but using expressions allows for greater flexibility. `samtools split -e '[XX] + "_" + [YY]'` would concatenate the values of tags XX and YY, and use that as the split key.

daviesrob:
Rebased, reworked and extended version of #1222.

Adds a new option to the `samtools split` command, which allows the user to split reads into distinct files by a custom read tag (with a string type value). E.g. `samtools split -d hc align.sam`, where some of the reads in `align.sam` have a tag in the format `hc:Z:<string value>`.

This version also adds a `-M NUM` option to limit the number of output files created, if the tag being split on has unexpectedly high cardinality.  At the moment, the records that would not have been split are treated as unaccounted - so an error unless you use `-u` to specify a file to write them into.  I do wonder if there should be something like a special exit code in the latter case though, as otherwise the incomplete split is a bit difficult to spot...

Running without `-d` splits only on read groups in the header, exactly as before.  Using `-d RG` also does this, but relaxes the rule that all read groups must be in the header.  If it finds a record with a new read group, it will open a new split with a minimal `@RG` header line.  This allows files with slightly broken `RG` data to be processed more easily.

Some variable and struct member names are changed in `bam_split.c`, as they're no longer specifically used for read groups.

Closes #1222
Fixes #1758

### DIFF
--- a/bam_split.c
+++ b/bam_split.c
@@ -48,6 +48,7 @@ struct parsed_opts {
     const char *unaccounted_header_name;
     const char *unaccounted_name;
     const char *output_format_string;
+    const char *tag;
     bool verbose;
     int no_pg;
     sam_global_args ga;
@@ -61,11 +62,11 @@ struct state {
     samFile* unaccounted_file;
     sam_hdr_t* unaccounted_header;
     size_t output_count;
-    char** rg_id;
+    char **rg_id;
     char **rg_index_file_name;
     char **rg_output_file_name;
-    samFile** rg_output_file;
-    sam_hdr_t** rg_output_header;
+    samFile **rg_output_file;
+    sam_hdr_t **rg_output_header;
     kh_c2i_t* rg_hash;
     htsThreadPool p;
     int write_index;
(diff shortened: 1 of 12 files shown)


whitwham:
I'm getting a failure when using the **--write-index** option.

> samtools split --write-index -u other.bam -d MC rg_test_md.sam
> 
> samtools split: Creating index file for file "rg_test_md_30M.bam" failed: No such file or directory
> samtools split: Could not create output file for tag "MC:30M"

File **rg_test_md_30M.bam** exists but seems to be corrupted.

> samtools view rg_test_md_30M.bam
> [main_samview] fail to read the header from "rg_test_md_30M.bam"

The index works with the RG tag though.

bam_split.c
```diff
@@ -275,6 +300,151 @@ static int header_compatible(sam_hdr_t *hdr1, sam_hdr_t *hdr2)
...
+            print_error_errno("split", "Creating index file for file \"%s\" failed", new_file_name);
+            goto fail;
+        }
+    }
+
+    if (sam_hdr_write(new_sam_file, new_hdr) != 0) {
+        print_error_errno("split", "Couldn't write header to \"%s\"",
+                          new_file_name);
+        goto fail;
+    }
```
whitwham, line 420: Move the sam_hdr_write code to before the write index code to get it to work.

daviesrob: Now swapped, plus I fixed a couple of other issues I spotted on the way.

PR #1939, opened 2023-11-14
Make samtools mpileup -aa work on empty files.

jkbonfield:
See comment https://github.com/samtools/samtools/issues/496#issuecomment-1787904643

### DIFF
--- a/bam_plcmd.c
+++ b/bam_plcmd.c
@@ -570,9 +570,11 @@ static int mpileup(mplp_conf_t *conf, int n, char **fn, char **fn_idx)
     int ret;
     int last_tid = -1;
     hts_pos_t last_pos = -1;
+    int one_seq = 0;
 
     // begin pileup
     while ( (ret=bam_mplp64_auto(iter, &tid, &pos, n_plp, plp)) > 0) {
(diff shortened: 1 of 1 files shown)


bam_plcmd.c
```diff
@@ -811,8 +813,11 @@ static int mpileup(mplp_conf_t *conf, int n, char **fn, char **fn_idx)
             last_tid = tid0;
             last_pos = beg0-1;
             mplp_get_ref(data[0], tid0, &ref, &ref_len);
+        } else if (last_tid < 0 && !one_seq && conf->all > 1) {
+            last_tid = 0; // --aa on a blank file
+            mplp_get_ref(data[0], tid0, &ref, &ref_len);
         }
-       while (last_tid >= 0 && last_tid < sam_hdr_nref(h)) {
+        while (last_tid >= 0 && last_tid < sam_hdr_nref(h)) {
```
vasudeva8, line 820: on run it was observed that the data from 1st reference sequence is reused for all subsequent ones.

jkbonfield: Fixed by moving the mplp_get_ref into the loop.  Thanks

PR #1946, opened 2023-11-24
News for 1.19

jkbonfield:
Also renamed it to NEWS.md so it's rendered nicely in GitHub and we can use that as visual checking for formatting prior to migration to the Release Notes (or at least partially can, as it still hides the hash-number links).


daviesrob:
Merging progress so far to build release candidates.

PR #1952, opened 2023-12-11
Speed up optical duplicate marking.

### LINKED ISSUE #1949: Slow optical duplicate marking with absurdly deep data.
(run `samtools --version`)

Cutting edge development version:
samtools 1.18-23-g3ba299b-dirty
Using htslib 1.18-54-gf1b5470

- compiler (run `gcc --version` or `clang --version`)-

Linux 4.15.0-214-generic
x86_64
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

This issue is made for us to easily track.

Very slow optical duplicate marking in very deep data.  About two days to process a 22GB cram file.  File supplied and problem identified.  PR to follow.

whitwham:
Checking optical duplicates in very deep data was very slow.

By adding more sort criteria we can reduce the total amount of work that needs to be done and so speed up the processing.  Sped of the processing of a 22GB file from two days to 70 minutes.

Fixes #1949.

### DIFF
--- a/bam_markdup.c
+++ b/bam_markdup.c
@@ -1195,7 +1195,7 @@ static int check_duplicate_chain(md_param_t *param, khash_t(duplicates) *dup_has
         while (++end_name_match < list->length) {
             check_t *chk = &list->c[end_name_match];
 
-            if (memcmp(base_name + base->beg, bam_get_qname(chk->b) + chk->beg, base->len) != 0)
+            if ((base->len == chk->len) && memcmp(base_name + base->beg, bam_get_qname(chk->b) + chk->beg, base->len) != 0)
                 break;
         }
 
--- a/bam_markdup.c
+++ b/bam_markdup.c
@@ -122,6 +122,7 @@ typedef struct {
     int opt;
     int beg;
     int end;
+    int len;
 } check_t;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


bam_markdup.c
```diff
@@ -1170,106 +1184,113 @@ static int check_duplicate_chain(md_param_t *param, khash_t(duplicates) *dup_has
...
-        while (++count < list->length && (list->c[count].x - current->x <= param->opt_dist)) {
-            // while close enough along the x coordinate
-            check_t *chk = &list->c[count];
+        // find the end of the matching name parts
+        while (++end_name_match < list->length) {
+            check_t *chk = &list->c[end_name_match];
 
-            if (current->opt && chk->opt)
-                continue;
+            if (memcmp(base_name + base->beg, bam_get_qname(chk->b) + chk->beg, base->len) != 0)
```
daviesrob, line 1198: There should be another length check here, or you could over-count if `base_name + base->beg` is a prefix of `bam_get_qname(chk->b) + chk->beg`.  Due to the sort order, `base` will always be the shorter of the two, or the same length.  (This is only likely to cause a problem for an unfortunate combination of names).

whitwham: Done.

Issue #1954, opened 2023-12-18
how to understand the MPC in samtools stats output


litun-fkby:
Hi，
I want to know the columns mean about the MPC in the samtools stats output, it shows:
![image](https://github.com/samtools/samtools/assets/82378273/1f6da5bb-74ac-4f80-9f67-1e33f90aaf96)

The first columns is the cycle number ,second is the number of N's and the rest is the number of mismatches , but it also see  columns correspond to qualities , it's make me confused ,can you give me a detail for the each columns?

Thanks!

pd3:
Not sure how to explain it better, will this make it more clear:

We count the number of mismatches in aligned reads at each position, across all reads. Rows correspond to read cycles (i.e. position within the read) and columns to qualities, the values are the number of bases for that read cycle and quality. However, there are two extra columns. The first is the read cycle number, the second is the number of N's (for those we do not collect base qualities).

If still unclear, check also the code
https://github.com/samtools/samtools/blob/a4c73e4d41670957be6c68ff964a0310f58be9ab/stats.c#L1617-L1628

litun-fkby:
you mean the value in 1st row and 3rd column mean's the mismatch number of 1st read cycle and the base quality 0 , as follows:
![image](https://github.com/samtools/samtools/assets/82378273/1d6e9884-d01b-441a-9629-d04693679ca7)

can samtools stats output the total base in each cycle , and samtools output the cycle value consider  the read1 , read2 , forward alignment and reverse alignment ?

pd3:
> you mean the value in 1st row and 3rd column mean's the mismatch number of 1st read cycle and the base quality 0

Yes, there are no mismatching bases with BQ=0, at any cycle; these would be turned into N's by the base caller I am assuming.


> can samtools stats output the total base in each cycle , and samtools output the cycle value consider the read1 , read2 , forward alignment and reverse alignment ?

Not in one go. But you can run the `stats` command twice, each time pre-filtering the reads you want.

litun-fkby:
_Not in one go. But you can run the stats command twice, each time pre-filtering the reads you want._
I can pre-filtering the read1 and read2 to get the MPC result，but how to get the total base per cycle if i want to get the mismatch rate per cycle fastly.

pd3:
Well, you sum up numbers in the columns.

jkbonfield:
> > you mean the value in 1st row and 3rd column mean's the mismatch number of 1st read cycle and the base quality 0
> 
[...]

PR #1963, opened 2024-01-10
Add MPC to the samtools stats man page

jkbonfield:
This is a bit convoluted when it comes to what is a match and what is not.  For example N in seq and N in reference is added to the "N" column, as that is all sequence Ns irrespective of reference.

Conversley "R" in the reference is not counted as mismatch, even if the sequence is "C" (R is A or G). Matches are case insensitive.

See discussion in #1954

### DIFF
--- a/doc/samtools-stats.1
+++ b/doc/samtools-stats.1
@@ -74,6 +74,7 @@ BCC	ACGT content per cycle for BC barcode
 CRC	ACGT content per cycle for CR barcode
 OXC	ACGT content per cycle for OX barcode
 RXC	ACGT content per cycle for RX barcode
+MPC	Mismatch distribution per cycle
 QTQ	Quality distribution for BC barcode
 CYQ	Quality distribution for CR barcode
 BZQ	Quality distribution for OX barcode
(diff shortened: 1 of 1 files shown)


pd3 (approved):
Thinking if it's worth explaining what "cycle number" is. It happens that people occasionally don't know what it is

jkbonfield:
Possibly, but I used it because cycle is already in use elsewhere in the man page (about 20 times).  It felt like an easy fix.  If we want to replace cycle, then it ought to be its own PR as it affects the entire document rather than one field.

I'm not really sure what the right phrase is anyway, as every technology differs a little bit, and sometimes it doesn't even make sense.  Eg 454 (and now Ultima Genomics) use flows.  A flow may give any number of bases (including zero).   It just doesn't have cycle at all.  ONT and PacBio are unclocked, so streaming signal and determining the base boundaries.  Cycle doesn't really make snese there either.

I think when stats was written Illumina was the only option and so cycles became synonymous with base positions.  I guess "Nth base from 5' end" is what it means, but it's a mouthful.  I'd say do it as a separate PR, if we can work out a better term to use.

pd3:
I meant it as a small addition, e.g. `.. cycle number (base position within the read)` or something like that.

PR #1987, opened 2024-02-07
fix error message in faidx.c

nickzoic:
very minor fix, just changes an error message when loading fai index


daviesrob:
Thanks, minor fix now applied.

PR #2005, opened 2024-03-12
Allow splitting by integer tag value

daviesrob:
* Extends the `samtools split` `-d` option so that it works on integer tags as well as strings.
* Fixes a bug where the `-d TAG` option would override a `-f FORMAT` that came before it on the command line.
* Adds a `-p` option to allow numeric values to be zero-padded.
* Enhances `samtools split` format string and functional tests.


whitwham:
Could you add the new option to the usage?

PR #2009, opened 2024-03-20
Update CRAM versions

adthrasher:
CRAM version 3.1 is missing from the list of valid CRAM version numbers in the man page.


daviesrob:
Thanks for the update.

PR #2012, opened 2024-03-26
Make name sort put primary reads before secondary reads.

### LINKED ISSUE #2010: Sorting by name does not guarantee primary alignment sorts before secondary and supplementary
Apologies for missing data - this was already confirmed by @jkbonfield .

`samtools sort -n` sorts on name. Its secondary sort does not guarantee that within-group (the same read with multiple alignments) the primary alignment sorts before secondary and supplementary alignments. This affects stream processing as the primary alignment has the read sequence.

jkbonfield:
For completeness, we consider non-secondary supplementaries as before secondaries, and non-supplementary secondaries before secondary+supplementary.

This is because a (non-2ndary) supplementary read is a continuation of the primary read and should be grouped before alternate alignments for this read.

We still have all READ1 data before all READ2 data, so the primary alignment has not changed ordering.  This is simply a better way of resolving ties than keeping the original ordering.

Fixes #2010

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -240,8 +240,12 @@ static inline int heap_lt(const heap1_t a, const heap1_t b)
         case QueryName:
             t = strnum_cmp(bam_get_qname(a.entry.bam_record), bam_get_qname(b.entry.bam_record));
             if (t != 0) return t > 0;
-            fa = a.entry.bam_record->core.flag & 0xc0;
-            fb = b.entry.bam_record->core.flag & 0xc0;
+            fa = a.entry.bam_record->core.flag;
+            fb = b.entry.bam_record->core.flag;
+            // Sort order is READ1, READ2, (PRIMARY), SUPPLEMENTARY, SECONDARY
+            // Get the bits in this order so sort is a natural a-b
+            fa = ((fa&0xc0)<<8)|((fa&0x100)<<3)|((fa&0x800)>>3);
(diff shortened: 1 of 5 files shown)


jkbonfield:
For validation, I've just been doing some perl one-liner hackery to summarise the samtools view outputs. Eg:

```
samtools sort -Osam -n test.sam | perl -ane 'next if /^@/; next if $F[1]&1;$F[1]&=~0x10;$flags=`samtools flag $F[1]`;$flags=~s/.*\t//;printf("%s\t%03x\t%s",$F[0],$F[1],$flags)'

a	040	READ1
a	044	UNMAP,READ1
a	840	READ1,SUPPLEMENTARY
a	140	READ1,SECONDARY
a	940	READ1,SECONDARY,SUPPLEMENTARY
a	080	READ2
a	880	READ2,SUPPLEMENTARY
a	180	READ2,SECONDARY
a	980	READ2,SECONDARY,SUPPLEMENTARY
b	044	UNMAP,READ1
b	040	READ1
b	840	READ1,SUPPLEMENTARY
b	140	READ1,SECONDARY
b	940	READ1,SECONDARY,SUPPLEMENTARY
b	080	READ2
b	880	READ2,SUPPLEMENTARY
b	180	READ2,SECONDARY
b	980	READ2,SECONDARY,SUPPLEMENTARY
```

PR #2014, opened 2024-03-27
Updates to manpages.

whitwham:
Mostly just clarifications and expanding explanations.

### DIFF
--- a/doc/samtools-ampliconclip.1
+++ b/doc/samtools-ampliconclip.1
@@ -77,7 +77,9 @@ BED file.  While this tool was originally written for clipping read alignment
 positions which correspond to amplicon primer locations it can also be used in
 other contexts.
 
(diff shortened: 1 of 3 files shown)


doc/samtools-ampliconclip.1
```diff
@@ -113,7 +115,8 @@ Soft clip reads (default).
 Hard clip reads.
 .TP
 .B --both-ends
-Clip at both the 5' and the 3' ends where regions match.
+Clip at both the 5' and the 3' ends where regions match.  Negates the --strand
```
daviesrob, line 118: Is "negates" the right term to use here?  Maybe say "\fB--strand\fR is ignored when using this option"?

whitwham: Changed the wording.

PR #2020, opened 2024-04-04
Add NEWS ready for 1.20 release

daviesrob:
NEWS updates

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -38,6 +38,10 @@ Release a.b
   tags with integer values.
   (PR #2005, feature request #1956.  Requested by Alex Leonard)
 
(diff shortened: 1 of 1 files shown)


whitwham:
Is not including #1969 deliberate?  We have added non user-visible changes before.

jkbonfield:
> Is not including #1969 deliberate? We have added non user-visible changes before.

My view is it's just part of the issue we fixed which is documented in htslib.

Besides I'd personally be more selective and editorial for news and non-visible things shouldn't appear, as too much detail just means fewer people read it, or even those that do skim over things and the important bits get lost in the noise.  (Those who are interested in detail can just do a git log for the raw details anyway or sort PRs by date on github.)

PR #2025, opened 2024-04-10
bedcov: CLI option to output a header row

kdm9:
Hello,

After spending an hour trying to debug a strange result, I realized that I'd accidentally used the output of `samtools bedcov -d 1 ...` where `samtools bedcov -c` was expected. I've added functionality to output a (commented) header line to the bed-like output, so that it's immediately obvious what the meaning of each column is. This is disabled by default, enabled with -H.

I also added a hopefully more explicit description of the column order to the man page.

Thanks,
Kevin


vasudeva8:
Thanks, the PR is now merged.

Issue #1884, opened 2023-07-19
samtools quickcheck fails to detect missing EOF in a bam file


bsigurjonsson:
Running samtools 1.17
OS:  Darwin 22.4.0
arm64

gcc --version
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.4.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

**Steps to reproduce**

1. DL a bam & bai file from http://hgdownload.cse.ucsc.edu/goldenPath/hg19/encodeDCC/wgEncodeUwRepliSeq/ (wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam)
2. ls -lh to see it is 64K size
3. Truncate it: `truncate -s 60K wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam`
4. `samtools quickcheck wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam` and there is no error
5. `samtools view wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam | wc -l` and you get [W::bam_hdr_read] EOF marker is absent

jmarshall:
I've attempted to reproduce this on macOS (arm64 Darwin 22.5.0, much like the OP's environment).

I'm a bit confused because the _wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam_ I downloaded from the specified location is 92M, not 64K.

But nonetheless I truncated it with `truncate -s 64k` and `truncate -s 60k` and attempted to reproduce the problem. However for me `quickcheck` produced the expected messages:

```
$ ls -lh wg*.bam
-rw-r--r--  1 johnm  staff    60K 20 Jul 21:36 wgEncode60K.bam
-rw-r--r--  1 johnm  staff    64K 20 Jul 21:34 wgEncode64K.bam
-rw-r--r--  1 johnm  staff    92M 18 Feb  2011 wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam
$ samtools-1.17 quickcheck wg*.bam
wgEncode60K.bam was missing EOF block when one should be present.
wgEncode64K.bam was missing EOF block when one should be present.
$ samtools quickcheck wg*.bam     # 1.17-36-g03124db+htslib-1.17-59-ga809db7
wgEncode60K.bam was missing EOF block when one should be present.
wgEncode64K.bam was missing EOF block when one should be present.
```

[...]

PR #2032, opened 2024-04-17
Avoid tview buffer overflow for positions with ≥14 digits

jmarshall:
As reported by @jmunoz94 in bioconda/bioconda-recipes#47137, the following crashes on some platforms, e.g., Linux:

```
$ samtools tview -dT   -p 17:99999999999999999999 test/mpileup/mpileup.1.bam test/mpileup/mpileup.ref.fa
malloc(): invalid next size (unsorted)
Aborted

$ samtools tview -dT -p chr1:99999999999999999999 wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam GRCh38.fa
malloc(): corrupted top size
Aborted
```

(I happened to test this with [wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam](http://hgdownload.cse.ucsc.edu/goldenPath/hg19/encodeDCC/wgEncodeUwRepliSeq/wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam) lying around from previously looking into #1884. This particular BAM file demonstrates some additional problems as described below.)

Using AddressSanitizer this can be quickly tracked down to a buffer overflow within _bam_tview.c_ due to the very large position. This buffer's size was not revised when the 64-bit `hts_pos_t` was introduced in #1117. This PR rewrites it using `faidx_fetch_seq64()` instead to avoid needing to build region strings at all.

Testing this further with fewer 9s I noticed some additional problems:

```
$ samtools tview -d T -p chr1:999999 wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam GRCh38-chr1.fa
  1000001   1000011   1000021   1000031   1000041   1000051   1000061           
GGGTGGAGCGCGCCGCCACGGACCACGGGCGNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN
.....C.CA..CT.A..CA...AAG.ACCT.                                                 
.T.G.C.CA..CT.A..CA...AAG.ACCT.     
```

Notice that the reference line reverts to NNNNNN at positions past the reads. This turns out to be a buglet that has existed since 67392b892147cff4bfb8718a819bab86d577b4e3 in 2009! I've also fixed this in this commit: Fix `tv->ref` guard (compare the similar code in `tv_pl_func()`) so that areas with no reads show the reference rather than ...NNNNNN...

Additionally with this data file samtools goes into an infinite(?) loop when given two fewer 9s than in the original report:

```
$ samtools tview -d T -p chr1:999999999999999999 wgEncodeUwRepliSeqBg02esG1bAlnRep1.bam GRCh38-chr1.fa
[runs for a long time]
```

This can be traced to the `do … while` loop in `hts_itr_query()`, which loops perhaps forever through negative bin numbers. Possibly this code should check for a `beg` value massively beyond the maximum position covered by the index's bins and sidestep most of its processing. (I haven't addressed that in this PR and can raise it as an HTSlib issue if you wish.)

Finally these positions specified on the command line are well beyond 2<sup>63</sup> so are ingested as unrelated values within the range of `hts_pos_t` (or even negative) because `hts_parse_decimal()` does no overflow checking.

Looking back at PR #1171 there's surprisingly little discussion and no mention of overflow handling. It didn't occur to me that people would specify positions unrelated to any reality-based meaningful chromosome positions, so it didn't occur to me that people might specify numbers of such magnitude, beyond 2<sup>63</sup>. This could be left as is (garbage in, garbage out! Or alternatively: Doctor, it hurts when I do this…) on the basis that such input deserves what it gets, or overflow detection could be added to `hts_parse_decimal()` and reported in some way.

### DIFF
--- a/bam_tview.c
+++ b/bam_tview.c
@@ -364,16 +364,11 @@ int base_draw_aln(tview_t *tv, int tid, hts_pos_t pos)
     tv->ccol = 0;
     // print ref and consensus
     if (tv->fai) {
-        char *str;
         if (tv->ref) free(tv->ref);
         assert(tv->curr_tid>=0);
 
         const char *ref_name = sam_hdr_tid2name(tv->header, tv->curr_tid);
-        str = (char*)calloc(strlen(ref_name) + 30, 1);
-        assert(str!=NULL);
-        sprintf(str, "%s:%"PRIhts_pos"-%"PRIhts_pos, ref_name, tv->left_pos + 1, tv->left_pos + tv->mcol);
-        tv->ref = fai_fetch64(tv->fai, str, &tv->l_ref);
-        free(str);
+        tv->ref = faidx_fetch_seq64(tv->fai, ref_name, tv->left_pos, tv->left_pos + tv->mcol - 1, &tv->l_ref);
         if ( !tv->ref )
         {
             fprintf(stderr,"Could not read the reference sequence. Is it seekable (plain text or compressed + .gzi indexed with bgzip)?\n");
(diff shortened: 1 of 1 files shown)


jkbonfield:
Good catch - thanks.

> This can be traced to the do … while loop in hts_itr_query(), which loops perhaps forever through negative bin numbers. Possibly this code should check for a beg value massively beyond the maximum position covered by the index's bins and sidestep most of its processing. (I haven't addressed that in this PR and can raise it as an HTSlib issue if you wish.)

It's probably worth fixing as it's another potential denial of service to web servers and the like, which could happen with anything doing bin lookups.

I noted this hangs
```
samtools tview -dT   -p 17:316659272122369 test/mpileup/mpileup.1.bam test/mpileup/mpileup.ref.fa
```

While ...68 does not.

I assume this is where it goes negative, but the complexities of the bin computation means it's an unexpectedly odd number  (`11ffffb6e0000` in hex is the last valid value).

PR #2042, opened 2024-04-28
plot-bamstats: Add option for alternative method for determining average base quality

### LINKED ISSUE #2017: plot-bamstats: Issue with accuracy of quality score plot
1.19.2+htslib-1.19.1

- OS: Linux 4.18.0-372.32.1.el8_6.x86_64
- compiler: gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-10)

1. ONT ubams converted to fastq
`samtools bam2fq -T 1 --reference (hg38.reference.genome) {input} > {output}`

2. minimap2 alignment
`minimap2 -y -ax map-ont {input} {output}`

3. samtools bam merging, sorting and indexing

4. `samtools stats {input} > {output}`

5. `plot-bamstats -p {prefix} {input} > {output}` 

#### Description of the problem

The quality score distribution given by plot-bamstats does not seem commensurate with other QC metrics that have been computed on this data.

For this data, other QC scoring algorithms (i.e. NanoPlot) tell me, for this specific dataset, I have a median read length of 2049, and a median read quality score of 16.5, with a small number of reads greater than 100 kb in length.

image

These plots would imply that the average quality score for most reads should be about 30, which is not commensurate with other QC metrics available to me. For comparison, here is the plot generated from NanoPlot using the same data:

image

The NanoPlot QC plot is computed directly from the ubams, and not from the aligned bam file

Or maybe I am misinterpreting the results of the plot?

I wonder if there is an issue with the decoding of quality scores from the bam files for this data type?

TBradley27:
Fixes #2017

It helps to illustrate the utility of this PR using an example:

This is a quality score plot generated by plot-bamstats without the `--alt-qual-avg-method` option being set:

PAQ60330_main-quals2

The is the same type of plot generated by plot-bamstats (for the same  dataset) when the `--alt-qual-avg-method` option is applied:

PAQ60330_alt-quals2

For the bottom plot, I think there is special utility to the users of plotting both median and mean base quality averages (using the non-default method) on the same plot, in assessing the relative proportion of bases of different quality scores within a dataset.

## A note on the example dataset used

In the example plots above, the results derive from one sequencing run on an Oxford Nanopore Technologies sequencing platform (_i.e._ a PromethION 24). It is an atypical dataset in the sense that due to reasons currently unknown, it is enriched for low quality bases (within reads and within base positions) compared to a typical sequencing dataset of this type. However, I chose this dataset as an example, as it conveys the ability of this alternative base quality averaging method to detect the influence of more extreme base error probabilities.

## A note on the choice of user interface with this specific option

As discussed in #2017, I now believe the alternative method of taking the mean base quality to be more meaningful than taking the mean of base quality scores. However, it was also discussed in #2017, that due to user's prior expectations and previous experience with this software, allowing users to compute average base quality scores in both ways would be helpful. This is also helpful for allowing users to compare the relative utility of both averaging approaches to each other. To be extra conservative, the previous method is applied as default. This is similar to approaches taken in popular sequencing data QC software which in part performs a similar function (see: https://github.com/s-andrews/FastQC/issues/110).


jkbonfield:
I notice this is only apply to FFQ and not LFQ.  That's OK for ONT data, but for consistency the new `-q` option should apply to both.

I've made the edits myself as it's easier than going to and fro and will merge soon.  Thanks for adding this option.

TBradley27:
Thank you, and apologies for the oversight

jkbonfield:
Rebased and merged as 6b0cc4a0e047f9c309bdefc531cdc117dd6e09d1 with an additional commit for LFQ to.  Thanks

PR #2047, opened 2024-05-09
update to exclude Duplicates from reset output

vasudeva8:
Reset doesn't remote the duplicate alignments during its operation. This was reported by Kevin Levis.
It removes supplementary and secondary alignments. This PR adds duplicate as well to this list as default.
Also added an option to override the default using flags.

To avoid the update of existing tests results, the exclude flag was added to get old behaviour on reset command.


whitwham:
For samtools reset you need to clear the duplicate flag but keep the read.  Unlike supplementary or secondary alignments, duplicate sequences are real reads that came out of the sequencer.  The original fastq files would have those sequences in.

vasudeva8:
The duplicate flag is cleared by default and an override to have it as it was - the initial behaviour - provided.
Test cases updated to use this override, --dupflag, to avoid output file update.

PR #2067, opened 2024-06-05
faidx with bgzf compressed output

### LINKED ISSUE #2055: bgzip'ed output from faidx
For the sake of convenience, it would be nice if the `fasta` output from `samtools faidx` could be bgzip'ed.

A command line argument that would enable the `fasta` output from `samtools faidx` (and potentially other tools with a `fasta` output) to be bgzip'ed.

vasudeva8:
fixes #2055
fasta/fastq output of faidx/fqidx will be bgzf compressed if the output file has an extension of .gz/.bgz/.bgzf.
stdout will not be compressed.
also added extended option output-fmt-option to set compression levels and threads for bgzf compression.

### DIFF
--- a/doc/samtools-faidx.1
+++ b/doc/samtools-faidx.1
@@ -3,7 +3,7 @@
 .SH NAME
 samtools faidx \- indexes or queries regions from a fasta file
 .\"
-.\" Copyright (C) 2008-2011, 2013-2018, 2020 Genome Research Ltd.
+.\" Copyright (C) 2008-2011, 2013-2018, 2020, 2024 Genome Research Ltd.
 .\" Portions copyright (C) 2010, 2011 Broad Institute.
 .\"
 .\" Author: Heng Li <lh3@sanger.ac.uk>
@@ -56,9 +56,9 @@ will index the file and create
 on the disk. If regions are specified, the subsequences will be
 retrieved and printed to stdout in the FASTA format.
 
-The input file can be compressed in the
+The input and output can be files compressed in the
 .B BGZF
-format.
+format. When output is compressed, the default compression level is 4.
 
 The sequences in the input file should all have different names.
 If they do not, indexing will emit a warning about duplicate sequences and
@@ -73,7 +73,10 @@ any extracted subsequence will be in FASTA format.
 
 .TP 8
 .BI "-o, --output " FILE
-Write FASTA to file rather than to stdout.
+Write FASTA to file rather than to stdout. With .gz, .bgz, .bgzf file extensions
+the output will be
+.B BGZF
+compressed.
 .TP
 .BI "-n, --length " INT
 Length for FASTA sequence line wrapping.  If zero, this means do not
@@ -124,6 +127,12 @@ Read/Write to specified compressed file index (used with .gz files).
 .TP
 .B -h, --help
 Print help message and exit.
+.TP
+.BI --output-fmt-option\  OPT=VAL
+Set the output format options, level=0..9 for compression level 0 to 9.
+.TP
+.BI "-@, --threads " N
+Set the number of extra threads for operations on compressed files.
 
 .SH AUTHOR
 .PP
--- a/faidx.c
+++ b/faidx.c
@@ -1,6 +1,6 @@
 /*  faidx.c -- faidx subcommand.
 
-    Copyright (C) 2008, 2009, 2013, 2016, 2018-2020, 2022 Genome Research Ltd.
+    Copyright (C) 2008, 2009, 2013, 2016, 2018-2020, 2022, 2024 Genome Research Ltd.
     Portions copyright (C) 2011 Broad Institute.
 
     Author: Heng Li <lh3@sanger.ac.uk>
@@ -44,6 +44,9 @@ History:
 #include <htslib/hts.h>
 #include <htslib/hfile.h>
 #include <htslib/kstring.h>
+#include <htslib/bgzf.h>
+#include <htslib/thread_pool.h>
+#include "sam_opts.h"
 #include "samtools.h"
 
 // Negative indicates the same as input data
@@ -52,6 +55,15 @@ History:
 #ifndef ABS
 #   define ABS(x) ((x)>=0?(x):-(x))
 #endif
+
+//new params required for output creation
+typedef struct output {
+    int isbgzip;                //is bgzip or uncompressed file
+    FILE *fp;                   //uncompressed file pointer
+    BGZF *bgzf_fp;              //bgzf file pointer
+    sam_global_args *gopt;      //options
+    kstring_t buffer;
+} output;
 
 static unsigned char comp_base[256] = {
   0,   1,   2,   3,   4,   5,   6,   7,   8,   9,  10,  11,  12,  13,  14,  15,
@@ -98,8 +110,19 @@ static void reverse(char *str, const hts_pos_t len) {
     }
 }
 
+/// wrappedwrite  - wraps the fwrite and bgzf_write
+/** @param out    - pointer to data required to write output
+*   @param buffer - data to write
+*   @param length - data length
+* returns error or length written on success
+*/
+static inline size_t wrappedwrite(output *out, const char *buffer, size_t length)
+{
+    return out->isbgzip ? bgzf_write(out->bgzf_fp, buffer, length) :
+        fwrite(buffer, 1, length, out->fp);
+}
 
-static int write_line(faidx_t *faid, FILE *file, const char *line, const char *name,
+static int write_line(faidx_t *faid, output *out, const char *line, const char *name,
                       const int ignore, const hts_pos_t length, const hts_pos_t seq_len) {
     int id;
     hts_pos_t beg, end;
@@ -124,8 +147,8 @@ static int write_line(faidx_t *faid, FILE *file, const char *line, const char *n
     for (i = 0; i < seq_sz; i += length)
     {
         hts_pos_t len = i + length < seq_sz ? length : seq_sz - i;
-        if (fwrite(line + i, 1, len, file) < len ||
-            fputc('\n', file) == EOF) {
+        if (wrappedwrite(out, line + i, len) < len ||
+              wrappedwrite(out, "\n", 1) < 1) {
             print_error_errno("faidx", "failed to write output");
             return EXIT_FAILURE;
         }
@@ -135,58 +158,66 @@ static int write_line(faidx_t *faid, FILE *file, const char *line, const char *n
 }
 
 
-static int write_output(faidx_t *faid, FILE *file, const char *name, const int ignore,
+static int write_output(faidx_t *faid, output *out, const char *name, const int ignore,
                         const hts_pos_t length, const int rev,
                         const char *pos_strand_name, const char *neg_strand_name,
                         enum fai_format_options format) {
-    hts_pos_t seq_len, wrap_len = length;
+    hts_pos_t seq_len, wrap_len = length, len = 0;
+    char *seq =  NULL, *qual = NULL;
+    int ret = EXIT_FAILURE;
+
     if (wrap_len < 0)
         wrap_len = fai_line_length(faid, name);
     if (wrap_len <= 0)
         wrap_len = HTS_POS_MAX;
-    char *seq = fai_fetch64(faid, name, &seq_len);
 
-    if (format == FAI_FASTA) {
-        fprintf(file, ">%s%s\n", name, rev ? neg_strand_name : pos_strand_name);
-    } else {
-        fprintf(file, "@%s%s\n", name, rev ? neg_strand_name : pos_strand_name);
-    }
-
+    seq = fai_fetch64(faid, name, &seq_len);
     if (rev && seq_len > 0) {
         reverse_complement(seq, seq_len);
     }
-
-    if (write_line(faid, file, seq, name, ignore, wrap_len, seq_len)
-        == EXIT_FAILURE) {
-        free(seq);
-        return EXIT_FAILURE;
+    //write the name
+    len = ksprintf(&out->buffer, "%c%s%s\n", format == FAI_FASTA ? '>' : '@', name, rev ? neg_strand_name : pos_strand_name);
+    if (wrappedwrite(out, out->buffer.s, out->buffer.l) < len) {
+        fprintf(stderr,"[faidx] Failed to write buffer\n");
+        goto exit;
+    }
+    ks_clear(&out->buffer);
+    //write bases
+    if ((ret = write_line(faid, out, seq, name, ignore, wrap_len, seq_len) == EXIT_FAILURE)) {
+        goto exit;
     }
 
-    free(seq);
-
     if (format == FAI_FASTQ) {
-        fprintf(file, "+\n");
-
-        char *qual = fai_fetchqual64(faid, name, &seq_len);
-
+        //write quality
+        qual = fai_fetchqual64(faid, name, &seq_len);
         if (rev && seq_len > 0) {
             reverse(qual, seq_len);
         }
 
-        if (write_line(faid, file, qual, name, ignore, wrap_len, seq_len)
-            == EXIT_FAILURE) {
-            free(qual);
-            return EXIT_FAILURE;
+        len = ksprintf(&out->buffer, "+\n");
+        if (wrappedwrite(out, out->buffer.s, out->buffer.l) < len) {
+            fprintf(stderr,"[faidx] Failed to write buffer\n");
+            goto exit;
         }
+        ks_clear(&out->buffer);
+        if ((ret = write_line(faid, out, qual, name, ignore, wrap_len, seq_len) == EXIT_FAILURE)) {
+            goto exit;
+        }
+    }
+    ret = EXIT_SUCCESS;
 
+exit:
+    if (seq) {
+        free(seq);
+    }
+    if (qual) {
         free(qual);
     }
-
-    return EXIT_SUCCESS;
+    return ret;
 }
 
 
-static int read_regions_from_file(faidx_t *faid, hFILE *in_file, FILE *file, const int ignore,
+static int read_regions_from_file(faidx_t *faid, hFILE *in_file, output *out, const int ignore,
                                   const hts_pos_t length, const int rev,
                                   const char *pos_strand_name,
                                   const char *neg_strand_name,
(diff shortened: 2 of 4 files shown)


faidx.c
```diff
@@ -124,69 +152,85 @@ static int write_line(faidx_t *faid, FILE *file, const char *line, const char *n
...
+    //make buffer to format the name
+    namelen = 3 + strlen(name) + strlen(rev ? neg_strand_name : pos_strand_name);   // >/@ + namelen + strand len + \n + null
+    if (!out->buffer || out->buff_len < namelen) {
+        out->buffer = realloc(out->buffer, namelen);
+        out->buff_len = namelen;
+    }
+    if (!out->buffer) {
+        fprintf(stderr,"[faidx] Couldn't create buffer\n");
+        goto exit;
     }
```
jkbonfield, line 188: I'm wondering if we should just create a `bgzf_putline` to be analogous the the `bgzf_getline` function so we can use kstring instead of handling our own dynamic reallocation buffer.

Or if not, then kstring here directly instead.  That would mean the existing fprintfs just become ksprintfs and the code change is relatively lightweight.

jkbonfield (commented):
I have more review to do still, but this was a very rapid first pass.

jkbonfield:
One thing that is noticable on fasta data is the compressability of DNA sequence is just poor with deflate.

On human chromosome 1 lidbeflate maximum compression gets 1.94 bits per base.  It's barely better than a naive 2-bit encoding.  Zlib's best is only 2.2bpb.

I plotted both zlib (top) and libdeflate (bottom) together.  Each point is a compression level from 1 to 9.  We see level 4 on libdeflate is a sharp corner and beyond that it's really not worth while trying harder.  With zlib (which ideally no one would use) it's somewhere between 4, 5 or 6.  I think therefore the default compression level should perhaps be different here.

Also, I noticed saying `--output-format-option level=0` is equivalent to `level=6` rather than uncompressed.  Maybe this is generic and happens elsewhere too?  It doesn't with `samtools view` but that could well be writing the hfile arguments to work around the problem.

image

jkbonfield:
Curiously, if you drive zlib correctly then this also turns things on its head.  zlib can be better than libdeflate which can be better than zstd!  I expect zstd -19 would win on size, and there are lots of options in zstd to tweak too.  Note also this is whole file compression rather than block based, but it's just an easy test to implement.

The key is just accepting that LZ matches are hard to find in fasta files so there's little benefit (unless we're printing up many entries and they're all related).  So RLE mode is usually best bang-for-buck.  Zlib lets us set this directly, and ironically it offers far better compression ratios than the default strategy, even at maximum compression level.

```
      		       	    size       user time
zlib Z_DEFAULT_STRATEGY	-1  79259417    4.27
zlib Z_DEFAULT_STRATEGY	-6  68323107   29.91
zlib Z_DEFAULT_STRATEGY	-9  65413881  170.36
zlib Z_HUFFMAN_ONLY	    67423830    3.37
zlib Z_RLE		    64770965	3.94 <<<
libdeflate gzip -1	    73115740	1.92
libdeflate gzip -3	    65588478	5.79
libdeflate gzip -4	    62213062	8.32
libdeflate gzip -6	    61686376   16.35
libdeflate gzip -9	    61470656   70.74
libdeflate gzip -11	    60058576   67.89
zstd -1     		    71152620    1.67
zstd -3	    		    70563644	2.15
zstd -5	    		    69284690	4.95
zstd -9	    		    67189974   15.10
zstd -12		    65468432   39.34
zstd -15		    61001117  130.88
```

I'm not sure what we should do with this information though.  Probably just ignore it.  Our recommended approach is libdeflate anyway and that basically means just sticking with a modest compression level as the default to hit the sweet spot.

doc/samtools-faidx.1
```diff
@@ -124,6 +127,12 @@ Read/Write to specified compressed file index (used with .gz files).
 .TP
 .B -h, --help
 Print help message and exit.
+.TP
+.BI --output-fmt-option\  OPT=VAL
+Set the output format options, level=0..9 for compression level 0 to 9.
```
jkbonfield, line 132: Change this to 1 to 9, as level 0 is parsed as default compression (level 6 atm).  We could change that to be uncompressed, but it's pointless as you could just not ask for .gz in the filename then.

faidx.c
```diff
@@ -314,17 +380,26 @@ int faidx_core(int argc, char *argv[], enum fai_format_options format)
...
                     fprintf(stderr, "[faidx] Unknown --mark-strand option \"%s\"\n", optarg);
-                    return usage(stderr, format, EXIT_FAILURE);
+                    exit_status = usage(stderr, format, EXIT_FAILURE);
+                    goto exit2;
                 }
                 break;
             case 1001: fai_name = optarg; break;
             case 1002: gzi_name = optarg; break;
-            default:  break;
+            // handle standard samtool options like thread count, compression level...
```
jkbonfield, line 389: Very minor nit-pick!  But "samtools".

faidx.c
```diff
@@ -360,55 +435,105 @@ int faidx_core(int argc, char *argv[], enum fai_format_options format)
...
     if( output_file != NULL ) {
         if( strcmp( output_file, argv[optind] ) == 0 ) {
             fprintf(stderr,"[faidx] Same input/output : %s\n", output_file);
-            return EXIT_FAILURE;
+            goto exit2;
+        }
+        if (!out.isbgzip) {
+            out.fp = fopen( output_file, "w" );
+        }
+        else {
```
jkbonfield, line 450: I know we don't have a style guide, but checking the source I see 21

```
}
else {
```

style clauses (6 of which are in reset.c) and 408

```
} else {
```

So basically 95% of the time we're keeping both braces on the same line.  Maybe we should go with the flow and try to keep a consistent bracing style where possible.


Don't even get me started on

```
if (expr)
{
```

Lol.  (It's mostly stats.c in samtools and VCF world in htslib, but it's needless whitespace in my book!)

faidx.c
```diff
@@ -360,55 +435,105 @@ int faidx_core(int argc, char *argv[], enum fai_format_options format)
...
             fprintf(stderr,"[faidx] Cannot open \"%s\" for writing :%s.\n", output_file, strerror(errno) );
-            return EXIT_FAILURE;
+            goto exit2;
+        }
+        //set threading if set
+        if (out.isbgzip && out.gopt->nthreads > 0) {
+            if (bgzf_mt(out.bgzf_fp, out.gopt->nthreads, 256/*unused*/) < 0) {
+                fprintf(stderr,"[faidx] Failed to set threads\n");
+                goto exit1;
+            }
```
jkbonfield, line 476: This made me recall that faidx does support reading bgzf files too, but it's never had the option to multi-thread the decoder.

However `perf record` implies libdeflate takes about 30% of the CPU time up, so I doubt it's really relevant.  If we're also compressing, then it'll be an even lower percentage of the CPU so attempting to share a thread pool probably isn't particularly useful.

So after musing a bit, I think this is probably the best way to handle this already.

faidx.c
```diff
@@ -360,55 +435,105 @@ int faidx_core(int argc, char *argv[], enum fai_format_options format)
...
+            print_error_errno("faidx", "Failed to flush output\n");
+            exit_status = EXIT_FAILURE;
+        }
+    }
+    else {
+        if (bgzf_flush(out.bgzf_fp) == EOF) {
+            print_error_errno("faidx", "Failed to flush output\n");
+            exit_status = EXIT_FAILURE;
+        }
+    }
```
jkbonfield, line 514: Could be more succint and remove code duplication.  Eg

```
int flushed = out.isbgzip
    ? fflush(out.fp)
    : bgzf_flush(out.bgzf_fp);
if (flushed == EOF) {
    print_error_errno("faidx", "Failed to flush output\n");
    exit_status = EXIT_FAILURE;
}
```

vasudeva8: Done.

faidx.c
```diff
@@ -98,8 +110,24 @@ static void reverse(char *str, const hts_pos_t len) {
     }
 }
 
+/// _wrappedwrite - wraps the fprintf and bgzf_write
+/** @param output - pointer to data required to write output
+*   @param buffer - data to write
+*   @param length - data length
+* returns error or length written on success
+*/
+static int _wrappedwrite(output *out, const char *buffer, hts_pos_t length)
```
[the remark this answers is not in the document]
vasudeva8: Changed.

faidx.c
vasudeva8: changed to use kstring.

faidx.c
```diff
@@ -124,69 +152,85 @@ static int write_line(faidx_t *faid, FILE *file, const char *line, const char *n
...
-    char *seq = fai_fetch64(faid, name, &seq_len);
 
-    if (format == FAI_FASTA) {
-        fprintf(file, ">%s%s\n", name, rev ? neg_strand_name : pos_strand_name);
-    } else {
-        fprintf(file, "@%s%s\n", name, rev ? neg_strand_name : pos_strand_name);
+    //make buffer to format the name
+    namelen = 3 + strlen(name) + strlen(rev ? neg_strand_name : pos_strand_name);   // >/@ + namelen + strand len + \n + null
+    if (!out->buffer || out->buff_len < namelen) {
+        out->buffer = realloc(out->buffer, namelen);
```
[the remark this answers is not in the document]
vasudeva8: :-), noted the error.
Updated to use kstring.

doc/samtools-faidx.1
vasudeva8: there was the error of skipping 0 which is now fixed.
retaining 0..9 as it is the allowed range and as the same is supported by other samtools commands.
with 0, it will be uncompressed data in bgzf format. 
default continues to be set by underlying compression library.

faidx.c
```diff
@@ -360,55 +435,105 @@ int faidx_core(int argc, char *argv[], enum fai_format_options format)
...
+            char mode[13] = "w";
+            int level = 0;
+            while (opts) {
+                if (opts->opt == HTS_OPT_COMPRESSION_LEVEL) {   //compression level
+                    level = opts->val.i;
+                    break;
+                }
+                opts = opts->next;
+            }
+            if (level > 0) {
```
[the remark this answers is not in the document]
vasudeva8: updated to accept 0 as well.
with 0, output will be uncompressed in bgzf format, as other commands.

faidx.c
vasudeva8: updated to use a thread pool in reading as well as writing compressed files.

jkbonfield:
It looks good, but I still think the default compression level ought to be 4 rather than 6.  The default is just a really bad trade-off in CPU vs size for fasta data.

```
@ seq4[samtools.../samtools]130; for l in `seq 1 9`;do echo -e -n "$l\t";/usr/bin/time -f '%U' samtools faidx --output-fmt-option level=$l -@4 $HREF38 chr1 -o /tmp/_.gz 2>&1 | tr '\012' '\011';wc -c < /tmp/_.gz;done
1	2.23	73528585
2	3.27	71830032
3	5.16	65682361
4	8.91	62372818
5	16.09	61960402
6	35.22	62021583
7	51.57	62072212
8	53.30	61391026
9	88.69	61043614

@ seq4[samtools.../samtools]; /usr/bin/time -f '%U' samtools faidx  -@4 $HREF38 chr1 -o /tmp/_.gz 2>&1 | tr '\012' '\011';wc -c < /tmp/_.gz
35.08	62021583
```

vasudeva8:
Set the default level as 4 instead of -1/zlib default.
Man page updated for the same.

jkbonfield:
There's some trailing white space in the man page that I was about to just force push up prior to merge (we don't need a discussion on it) and one minor slight rewording about default compression levels to avoid an implication that the default output style is compressed.

However I also spotted something with the kstring so I'll comment on that elsewhere.  The man page change was this:

```
diff --git a/doc/samtools-faidx.1 b/doc/samtools-faidx.1
index fffb0add..99497710 100644
--- a/doc/samtools-faidx.1
+++ b/doc/samtools-faidx.1
@@ -58,7 +58,7 @@ retrieved and printed to stdout in the FASTA format.
 
 The input and output can be files compressed in the
 .B BGZF
-format. The default output compression level is 4.
+format. When output is compressed, the default compression level is 4.
 
 The sequences in the input file should all have different names.
 If they do not, indexing will emit a warning about duplicate sequences and
@@ -74,7 +74,7 @@ any extracted subsequence will be in FASTA format.
 .TP 8
 .BI "-o, --output " FILE
 Write FASTA to file rather than to stdout. With .gz, .bgz, .bgzf file extensions
-the output will be 
+the output will be
 .B BGZF
 compressed.
 .TP
```

(You can't see the "the output will be " space, but it's there.)  Trivial and we don't strictly check man pages, only code, but I spotted it in diff so thought I'd just tidy it during merging.

faidx.c
```diff
@@ -53,6 +56,15 @@ DEALINGS IN THE SOFTWARE.
 #   define ABS(x) ((x)>=0?(x):-(x))
 #endif
 
+//new params required for output creation
+typedef struct output {
+    int isbgzip;                //is bgzip or uncompressed file
+    FILE *fp;                   //uncompressed file pointer
+    BGZF *bgzf_fp;              //bgzf file pointer
+    sam_global_args *gopt;      //options
+    kstring_t *buffer;
```
jkbonfield, line 65: This doesn't need to be a pointer with a calloc / free.  kstring_t is a fixed size struct and the memory allocation is handled internally, so you can just do `kstring_t buffer` instead of `*buffer` and then either `ks_initialise(&buffer);` in place of the calloc, or easier just change the initialisation of `struct output` to use `KS_INITIALIZE` in situ.  Eg:

```
struct output out = { 0, stdout, NULL, &ga, KS_INITIALIZE};
```

I tried it quickly (see ~jkb/work/samtools_master/samtools/faidx.c) and it's just a matter of &ks or ks.s instead of ks and ks->s in the various function calls.


That avoids callocs + the error check there and the free statement.

```
diff --git a/faidx.c b/faidx.c
index 652df65c..ddd376b9 100644
--- a/faidx.c
+++ b/faidx.c
@@ -62,7 +62,7 @@ typedef struct output {
     FILE *fp;                   //uncompressed file pointer
     BGZF *bgzf_fp;              //bgzf file pointer
     sam_global_args *gopt;      //options
-    kstring_t *buffer;
+    kstring_t buffer;
 } output;
 
 static unsigned char comp_base[256] = {
@@ -171,24 +171,17 @@ static int write_output(faidx_t *faid, output *out, const char *name, const int
     if (wrap_len <= 0)
         wrap_len = HTS_POS_MAX;
 
-    if (!out->buffer) {
-        if (!(out->buffer = calloc(1, sizeof(kstring_t)))) {
-            fprintf(stderr,"[faidx] Couldn't create buffer\n");
-            goto exit;
-        }
-    }
-
     seq = fai_fetch64(faid, name, &seq_len);
     if (rev && seq_len > 0) {
         reverse_complement(seq, seq_len);
     }
     //write the name
-    len = ksprintf(out->buffer, "%c%s%s\n", format == FAI_FASTA ? '>' : '@', name, rev ? neg_strand_name : pos_strand_name);
-    if (wrappedwrite(out, out->buffer->s, out->buffer->l) < len) {
+    len = ksprintf(&out->buffer, "%c%s%s\n", format == FAI_FASTA ? '>' : '@', name, rev ? neg_strand_name : pos_strand_name);
+    if (wrappedwrite(out, out->buffer.s, out->buffer.l) < len) {
         fprintf(stderr,"[faidx] Failed to write buffer\n");
         goto exit;
     }
-    ks_clear(out->buffer);
+    ks_clear(&out->buffer);
     //write bases
     if ((ret = write_line(faid, out, seq, name, ignore, wrap_len, seq_len) == EXIT_FAILURE)) {
         goto exit;
@@ -201,12 +194,12 @@ static int write_output(faidx_t *faid, output *out, const char *name, const int
             reverse(qual, seq_len);
         }
 
-        len = ksprintf(out->buffer, "+\n");
-        if (wrappedwrite(out, out->buffer->s, out->buffer->l) < len) {
+        len = ksprintf(&out->buffer, "+\n");
+        if (wrappedwrite(out, out->buffer.s, out->buffer.l) < len) {
             fprintf(stderr,"[faidx] Failed to write buffer\n");
             goto exit;
         }
-        ks_clear(out->buffer);
+        ks_clear(&out->buffer);
         if ((ret = write_line(faid, out, qual, name, ignore, wrap_len, seq_len) == EXIT_FAILURE)) {
             goto exit;
         }
@@ -297,7 +290,7 @@ int faidx_core(int argc, char *argv[], enum fai_format_options format)
     char *gzi_name = NULL; // specified compressed index name
     sam_global_args ga = SAM_GLOBAL_ARGS_INIT;
     int exit_status = EXIT_FAILURE, flushed = 0;
-    struct output out = { 0, stdout, NULL, &ga, NULL};          //data required for output writing
+    struct output out = { 0, stdout, NULL, &ga, KS_INITIALIZE};          //data required for output writing
     faidx_t *fai = NULL;
     hts_tpool *pool = NULL;
 
@@ -531,10 +524,7 @@ exit2:
         hts_tpool_destroy(pool);
     }
     sam_global_args_free(&ga);
-    if (out.buffer) {
-        ks_free(out.buffer);
-        free(out.buffer);
-    }
+    ks_free(&out.buffer);
     return exit_status;
 }
```

jkbonfield: Note there's nothing wrong with what you did!  The code works, but it's useful to learn easier ways of manipulating kstrings.

faidx.c
```diff
@@ -44,6 +44,9 @@ DEALINGS IN THE SOFTWARE.
 #include <htslib/hts.h>
 #include <htslib/hfile.h>
 #include <htslib/kstring.h>
+#include <htslib/bgzf.h>
+#include <htslib/thread_pool.h>
+#include "sam_opts.h"
```
jmarshall, line 49: Should add correspondingly to the rule in _Makefile_:

```diff
-faidx.o: faidx.c config.h $(htslib_faidx_h) $(htslib_hts_h) $(htslib_hfile_h) $(htslib_kstring_h) $(samtools_h)
+faidx.o: faidx.c config.h $(htslib_faidx_h) $(htslib_hts_h) $(htslib_hfile_h) $(htslib_kstring_h) $(htslib_bgzf_h) $(htslib_thread_pool_h) $(sam_opts_h) $(samtools_h)
```

faidx.c
vasudeva8: updated.

PR #2072, opened 2024-06-26
Switch from AppVeyor to GitHub Actions.

jkbonfield:
Also changed test/** to text mode as Actions enforces windows cr-lf on git clone, unlike AppVeyor's git.  We could be more selective in this, but this replicates what we had before.

[If we wish to cross-validate our support for e.g. SAM in both line endings, then this should be done by explicit files with both types so we can validate on both platforms, but I don't know if we wish to support that anyway.]

At the time of writing (NB it's late afternoon UK time so during peak US time for github), GitHub Actions completed in ~6 mins while AppVeyor took approx 12+ minutes (only 10 of which was running with 2-3 spent waiting for a runner to be available).

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -1,0 +1,55 @@
+name: Windows/MinGW-W64 CI
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: windows-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Set up MSYS2 MinGW-W64
+      uses: msys2/setup-msys2@v2
+      with:
+        msystem: mingw64
+        update: false
+        install: >-
+          mingw-w64-x86_64-toolchain
+          mingw-w64-x86_64-autotools
+          mingw-w64-x86_64-curl
+          mingw-w64-x86_64-libdeflate
+          mingw-w64-x86_64-tools-git
+          mingw-w64-x86_64-zlib
+          mingw-w64-x86_64-bzip2
+          mingw-w64-x86_64-xz
+          mingw-w64-x86_64-ncurses
+    - name: Clone htslib
+      shell: msys2 {0}
+      run: |
+        # Todo 4th arg to track commit message of htslib#PRNUM to
+        # check out a specific htslib PR instead.
+        # Other useful vars:
+        # GITHUB_BASE_REF:  e.g. "develop"
+        # GITHUB_HEAD_REF:  e.g. "my-PR-feature" (aka GITHUB_REF_NAME?)
+        export PATH="$PATH:/mingw64/bin:/c/Program Files/Git/bin"
+        export MSYSTEM=MINGW64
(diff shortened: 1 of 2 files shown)


.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,56 @@
...
+        # GITHUB_BASE_REF:  e.g. "develop"
+        # GITHUB_HEAD_REF:  e.g. "my-PR-feature" (aka GITHUB_REF_NAME?)
+        .ci_helpers/clone "https://github.com/${GITHUB_REPOSITORY_OWNER}/htslib" htslib "${GITHUB_REF_NAME}"
+        export PATH=$PATH:/mingw64/bin
+        export MSYSTEM=MINGW64
+        pushd .
+        cd htslib
+        autoreconf -i
+        ./configure
+        make -j32
```
daviesrob, line 39: Configure should use `--enable-werror` so warnings will be obvious.
`make -j32` seems a bit ambitious given that the [runners only have 4 CPUs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,56 @@
...
+        popd
+    - name: Compile samtools
+      shell: msys2 {0}
+      run: |
+        export PATH=$PATH:/mingw64/bin
+        export MSYSTEM=MINGW64
+        autoheader
+        autoconf -Wno-syntax
+        ./configure
+        make -j32
```
daviesrob, line 49: Seem comments on the "Compile htslib" section.

daviesrob (commented):
Again, it would be useful to leave Appveyor in place for a few days while we swap over.

.github/workflows/windows-build.yml
jkbonfield: I switched to make -j6.  I know it has 4 CPUs, but often with mingw it becomes I/O bound so over-committing on CPUs can speed things up.

jkbonfield: I'll try a -j4 and look at timings too :)

daviesrob:
Comparing with the old AppVeyor logs, I noticed that `libcurl` had gone missing, and `libdeflate` was never there but is now available as an MSYS2 package.  I've pushed up a commit to add them both.

One remaining issue is that HTSlib's `configure` gets run twice.  I this should be fixable by only running `autoreconf -i` after cloning HTSlib, and then letting SAMtools configure and make everything itself.  At least, that's how the Appveyor script used to do it, and it's probably easier than trying to work out why samtools' `./configure` feels the need to run the one in `htslib` again.

daviesrob:
Pushed version with tweaked MSYS2 package lists, and added libdeflate.  Comparing logs with AppVeyor show the same results, apart from the usage tests which are skipped due to `no IO::Pty module or can't open pty` on GitHub actions.

PR #2079, opened 2024-07-11
Add NEWS item for PR 1706, which was missed at the time.

daviesrob:


NEWS.md
```diff
@@ -326,6 +326,11 @@ New work and changes:
   aligners.
   (PR #1698)
 
+* Made better use of threading during the merge stage of samtools sort.
+  It also not limits the number of temporary files present by making
```
jmarshall, line 330: This may not be a good time to point out what I anticipated a review would catch…

`not limits` → `now limits` perhaps

jkbonfield: Doh!  I literally just read through this text during our weekly meeting too.

PR #2083, opened 2024-07-16
Handle alignments without quality.

### LINKED ISSUE #2076: Segmentation fault when running `samtools coverage --excl-flags 1540`
(run `samtools --version`)

- OS: Linux 5.15.0-1049-oracle
- compiler: gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

Using the attached aln.sam file:
```
samtools coverage --excl-flags 1540 aln.sam -o coverage.secondary.tsv
```
This gave a segmentation fault and produces a tsv output which stops part way through:
```
...
NOZR01000027.1	1	87550	1	109	0.1245	0.001245	53.9	0
NOZR01000030.1	1	70745	1	204	0.28836	0.0028836	61.7	0
NOZR01000039.1	1	35613	1	178	0.499817	0.00499817	66	0
NOZR01000054.1	1	5463	4	1072	19.6229	0.221856	61.3	0
NZ_RARC01000012.1	1	175888	1	47	0.0267216	0.000267216	50.6	0
NZ_RARC01000016.1	1	129040	2	167	0.129417	0.00129417	67.1	0
NZ_RARC0100002
```

The error occurs when samtools is installed via conda or installed from the github release. It runs fine if `--excl-flags 1540` is removed.
Doing an initial debug it seems to be coming from here, but couldn't figure out beyond that:
```
Program received signal SIGSEGV, Segmentation fault.
0x0000555555593e43 in main_coverage (argc=argc@entry=6, argv=<optimised out>, argv@entry=0x7fffffffdc60) at coverage.c:614
614	                        bam_get_qual(p->b)[p->qpos] < opt_min_baseQ) --depth_at_pos; // low base quality
```

sam file came from mapping ont reads against an multifasta reference, which was then sorted.
```
minimap2 -ax map-ont --secondary yes -N 1000 ref.fa reads.fastq > alignments.sam
```

[aln.zip](https://github.com/user-attachments/files/16143290/aln.zip)

whitwham:
Secondary alignments with CIGAR strings but without quality (valid in the sam spec) could cause segfaults and/or uninitialised memory access in samtools coverage.

The presence of a CIGAR string means they have length (for the purposes of coverage) but they have no sequence or quality data.  The guards on trying to access non-existent quality data were imperfect.

This commit fixes the faulty memory access and changes the quality calculations to just use existing qualities.  This changes existing behaviour but as secondary values are excluded by default the effects should be minor.  A warning is printed if there are missing qualities.

This should fix #2076.

### DIFF
--- a/coverage.c
+++ b/coverage.c
@@ -60,6 +60,7 @@ typedef struct {  // auxiliary data structure to hold stats on coverage
     unsigned long long summed_coverage;
     unsigned long long summed_baseQ;
     unsigned long long summed_mapQ;
+    unsigned long long quality_bases;
     unsigned int n_reads;
     unsigned int n_selected_reads;
     bool covered;
@@ -204,7 +205,7 @@ void print_tabular_line(FILE *file_out, const sam_hdr_t *h, const stats_aux_t *s
             stats[tid].n_covered_bases,
             100.0 * stats[tid].n_covered_bases / region_len,
             stats[tid].summed_coverage / region_len,
-            stats[tid].summed_coverage > 0? stats[tid].summed_baseQ/(double) stats[tid].summed_coverage : 0,
+            stats[tid].quality_bases > 0? stats[tid].summed_baseQ/(double) stats[tid].quality_bases : 0,
             stats[tid].n_selected_reads > 0? stats[tid].summed_mapQ/(double) stats[tid].n_selected_reads : 0
            );
 }
@@ -263,7 +264,7 @@ void print_hist(FILE *file_out, const sam_hdr_t *h, const stats_aux_t *stats, in
             case 5: fprintf(file_out, "Mean coverage:   %.3gx",
                             stats[tid].summed_coverage / region_len); break;
             case 4: fprintf(file_out, "Mean baseQ:      %.3g",
-                            stats[tid].summed_baseQ/(double) stats[tid].summed_coverage); break;
+                            stats[tid].quality_bases > 0? stats[tid].summed_baseQ/(double) stats[tid].quality_bases : 0); break;
             case 3: fprintf(file_out, "Mean mapQ:       %.3g",
                             stats[tid].summed_mapQ/(double) stats[tid].n_selected_reads); break;
             case 1: fprintf(file_out, "Histo bin width: %sbp",
@@ -315,6 +316,7 @@ int main_coverage(int argc, char *argv[]) {
     char **fn = NULL;
     int fail_flags = (BAM_FUNMAP | BAM_FSECONDARY | BAM_FQCFAIL | BAM_FDUP); // Default fail flags
     int required_flags = 0;
+    int print_value_warning = 0;
 
     int *n_plp = NULL;
     sam_hdr_t *h = NULL; // BAM header of the 1st input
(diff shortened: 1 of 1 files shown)


coverage.c
```diff
@@ -609,20 +612,28 @@ int main_coverage(int argc, char *argv[]) {
...
-                        bam_get_qual(p->b)[p->qpos] < opt_min_baseQ) --depth_at_pos; // low base quality
-                else
-                    stats[tid].summed_baseQ += bam_get_qual(p->b)[p->qpos];
+                if (p->is_del || p->is_refskip) {
+                    --depth_at_pos; // having dels or refskips at tid:pos
+                } else if (p->qpos < p->b->core.l_qseq) {
+                    if (bam_get_qual(p->b)[p->qpos] < opt_min_baseQ) {
+                        --depth_at_pos; // low base quality
+                    } else {
+                        stats[tid].summed_baseQ += bam_get_qual(p->b)[p->qpos];
```
jkbonfield, line 621: While not new logic, this means that the "Mean BaseQ" reported by the tool is the average quality of bases which meet the minimum base quality criteria, rather than the average quality of all bases.  This isn't what I imagined it to be and the man page is rather ambiguous given it doesn't explicitly lean one way or the other.

It's probably too late to change the program functionality, unless it's a clear bug (but `depth_at_pos` is also filtered depth).  The man page problem just needs some tweakage to explain that the filtering takes place before computing averaged statistics.

jkbonfield:
From discussions in the office I was of the opinion this was a missing bounds check in the mpileup algorithm, but this is purely samtools coverage code.  If I misunderstood, then great as this is clearly the less bad scenario of the two.  Can you confirm that the same issues don't appear upstream anywhere in htslib?

coverage.c
whitwham: Alignments that don't have the minimum base quality don't get included in the coverage calculations.  So it is the average quality of used alignments.  I think the man page covers that.

This change does mess with the calculations, hence the warning.  It is a compromise so that the program doesn't have to exit when it comes across a missing quality value.

whitwham:
> From discussions in the office I was of the opinion this was a missing bounds check in the mpileup algorithm, but this is purely samtools coverage code. If I misunderstood, then great as this is clearly the less bad scenario of the two. Can you confirm that the same issues don't appear upstream anywhere in htslib?

This is a purely samtools change.  The errors were caused by _bam_get_qual_ being  called in the **samtools coverage** main loop on non-existent quality data.

coverage.c
```diff
@@ -653,6 +664,10 @@ int main_coverage(int argc, char *argv[]) {
         }
     }
 
+    if (print_value_warning) {
+        print_error("coverage", "Warning:  Missing quality values in alignments.  Mean base quality calculated only on available values.\n");
```
jkbonfield, line 668: I'm not sure if it's deliberate or not to add an extra line, but `print_error` already prints a newline.  It's hardly unique in this front though.  It looks like we have 18% of calls adding their own explicit newline.

jkbonfield:
Looks good to me, barring the question about whether we want a double or single newline in the error message.  It's kind of irrelevant, but thought I'd ask.

coverage.c
whitwham: You know, I even checked what `print_error` did and didn't notice that.

jkbonfield: I did notice that nearly all other uses of `print_error` with a newline come from bam_markdup ;-)  [Guilty too of a couple in amplicon_stats]

PR #2084, opened 2024-07-18
Fix undefined behaviour in samtools fastq with empty QUAL.

jkbonfield:
Incrementing via "qt_end += qt != NULL" will be qt_end+=0 when we have a null qt/qt_end, but NULL+0 is technically undefined.

    bam_fastq.c:852:30: runtime error: applying zero offset to null pointer
    SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior bam_fastq.c:852:30 in

Easy to just replace with two loops and avoid such cleverness.  It's not a major portability issue, but it's nice to have no warnings from ubsan as it makes finding other problems easier.

### DIFF
--- a/bam_fastq.c
+++ b/bam_fastq.c
@@ -848,8 +848,13 @@ int output_index(bam1_t *b1, bam1_t *b2, bam2fq_state_t *state,
         }
 
         char *bc_end = bc, *qt_end = qt;
-        while (len ? *bc_end && rem-- : isalpha(*bc_end))
-            bc_end++, qt_end += qt != NULL;
+        if (qt) {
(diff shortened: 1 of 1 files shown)


daviesrob:
Change looks OK, but is `qt_end-qt` used later strictly allowed if both are `NULL`?

jkbonfield:
Possibly not, but my aim wasn't so much to be 100% anal on the rules if they're not triggering problems and instead to simply get to a point where -fsanitize=undefined triggers no errors and could because part of the CI system (with appropriate envs to make it fail on error).

If other code subsequently starts to trigger issues, then the CI would notify us and we can fix it, but too much naval gazing isn't really productive when something is working.

PR #2085, opened 2024-07-23
Fix read length plot for limited variation data.

### LINKED ISSUE #2068: plot-bamstats fails when only one RL is in the stats file
Trying to use plot-bamstats with samtools-1.20 but the command fails.

Command:
```
plot-bamstats -p mypref mytestfile.stats
```

Command error:
```
"mypref-read_length.gp", line 20: Can't plot with an empty x range!

The command exited with non-zero status 256:
	gnuplot mypref-read_length.gp

 at /usr/local/bin/plot-bamstats line 54.
	main::error("The command exited with non-zero status 256:\x{a}\x{9}gnuplot mypref-"...) called at /usr/local/bin/plot-bamstats line 339
	main::plot("mypref-read_length.gp") called at /usr/local/bin/plot-bamstats line 1335
	main::plot_read_len_hist(HASH(0x55fb79728de8)) called at /usr/local/bin/plot-bamstats line 44
```

In the mypref-read_length.gp the xrange is [101: 101]. The input mytestfile.stats file has only one RL:
```
# Read lengths. Use `grep ^RL | cut -f 2-` to extract this part. The columns are: read length, count
RL	101	1959398
# Read lengths - first fragments. Use `grep ^FRL | cut -f 2-` to extract this part. The columns are: read length, count
FRL	101	979699
# Read lengths - last fragments. Use `grep ^LRL | cut -f 2-` to extract this part. The columns are: read length, count
LRL	101	979699
```

When running on a stats file with multiple values for RL, the plot-bamstats command finishes successfully. 

Also, should the generated HTML report include all images (as currently it looks like it is missing the *-coverage.png and *-read_length.png)?

whitwham:
Fixes #2068 where the limited length variations were causing the x range to be wrong.  Some other changes to quiet warnings about y ranges.

### DIFF
--- a/misc/plot-bamstats
+++ b/misc/plot-bamstats
@@ -25,7 +25,7 @@
 use strict;
 use warnings;
 use Carp;
-use POSIX qw/floor/;
+use POSIX qw/floor ceil/;
 use List::Util qw(max);
 
 my $opts = parse_params();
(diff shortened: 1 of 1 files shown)


misc/plot-bamstats
```diff
@@ -1231,6 +1231,44 @@ sub plot_indel_cycles
...
+sub nice_num
+{
+    my ($num) = @_;
+
+    my $expon_num = floor(log($num)/log(10));
+    my $fract_num = ($num / (10.0 ** $expon_num));
+
+    my $nf;
+
+    if ($fract_num <= 1.0) {
```
daviesrob, line 1244: This could be simplified a bit by importing `ceil` from `POSIX`, then doing `my $nf = ceil($fract_num);`.

I don't know if `$num` can ever be zero, which would cause trouble in `log()`?  I suspect not, but you could make sure with `if ($num < 1) { return 1; }`.

daviesrob (commented):
Looks like a good fix for the read length plot, although in testing I found it's still possible to break `plot-bamstats` elsewhere by not giving it much data. E.g.:

```
mkdir /tmp/stats
./samtools stats test/dat/view.002.sam > /tmp/view.002.stats
./misc/plot-bamstats -p /tmp/stats/view.002 /tmp/view.002.stats 
Warning: empty x range [1:1], adjusting to [0.99:1.01]
Warning: empty x range [1:1], adjusting to [0.99:1.01]
Warning: empty x range [2:2], adjusting to [1.98:2.02]

plot '-' smooth csplines with lines lc 1 title 'First fragments' 
     ^
"/tmp/stats/view.002-gc-content.gp", line 12: Can't calculate splines, need at least 3 points

The command exited with non-zero status 256:
	gnuplot /tmp/stats/view.002-gc-content.gp

```
This may be something to look at in the future.

Meanwhile, https://github.com/samtools/samtools/issues/2068#issue-2340100500 notes that the read length plot doesn't appear on the html page.  It doesn't look like it should be too hard to add...

misc/plot-bamstats
whitwham: Yeah, that was left over from a more complicated function I ended up not liking the results from.  I'll change it.

Issue #2086, opened 2024-07-23
RG ID unique suffixes are not fixed width in `samtools merge`


cjw85:
I've filed this as a bug only because my understanding of what it appears to me that the intent was, but maybe its a feature request.

#### Are you using the latest version of [samtools](https://github.com/samtools/samtools/releases/latest) and [HTSlib](https://github.com/samtools/htslib/releases/latest)? If not, please specify.
From bioconda:
```
samtools 1.19.2
Using htslib 1.19.1
```

 #### Please describe your environment.
Linux 5.15.0-88-generic

#### Please specify the steps taken to generate the issue, the command you are running and the relevant output.
Running `samtools merge` without `-p` generates hexadecimal string suffices to RG IDs when they clash between files. From [this line](https://github.com/samtools/samtools/blob/develop/bam_sort.c#L408) of code:

```
static int gen_unique_id(char *prefix, khash_t(cset) *existing_ids,
                         bool always_add_suffix, kstring_t *dest) {
                         
    ...
    ksprintf(dest, "%s-%0lX", prefix, lrand48());
    ...
}
```
It looks to me like the intention was to zero pad this such they are always the same length, but there's no field width after the `0` meaning that leading zeros aren't inserted.

I noticed this whilst trying to "deduplicate" the IDs in a BAM I have.

PR #2087, opened 2024-07-23
#2086 Zeropad unique suffix IDs

cjw85:
See #2086

Formatting of suffix used the code `"%0lX"`, the `0` says to pad with `'0'` (not `' '`), but no field width is specified resulting in no zero-padding


cjw85:
I've updated the expected test results, I must confess I don't know why some of the `@PG` lines are deleted when I run `make test`. All other changes show the expected zero-padded suffixes at a fixed width of 8.

whitwham:
Merged, thank you.

PR #2093, opened 2024-08-06
Add option to specify index location for idxstats.

### LINKED ISSUE #2071: Allow users to specify index location for idxstats
No

`samtools idxstats` assumes the bai file lives in the same directory as the bam file, but in my use case, bai files are in a different directory.

whitwham:
Add the -X option (used in some other tools) to specify the index location.  Not entirely necessary because there is another way to do it, but this is a more straight forward way.

Fixes #2071.

### DIFF
--- a/doc/samtools-idxstats.1
+++ b/doc/samtools-idxstats.1
@@ -3,7 +3,7 @@
 .SH NAME
 samtools idxstats \- reports alignment summary statistics
 .\"
-.\" Copyright (C) 2008-2011, 2013-2018 Genome Research Ltd.
+.\" Copyright (C) 2008-2011, 2013-2018, 2024 Genome Research Ltd.
 .\" Portions copyright (C) 2010, 2011 Broad Institute.
 .\"
 .\" Author: Heng Li <lh3@sanger.ac.uk>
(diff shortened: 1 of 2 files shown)


doc/samtools-idxstats.1
```diff
@@ -61,6 +61,14 @@ name, sequence length, # mapped read-segments and # unmapped
 read-segments. It is written to stdout.  Note this may count reads
 multiple times if they are mapped more than once or in multiple fragments.
 
+.SH OPTIONS
+.TP 11
+.B -X
+This option will allow the user to specify a customised index file
+location. Example usage: samtools idxstat [options] -X </data_folder/data.bam>
+[/index_folder/index.bai]
```
jkbonfield, line 69: I'd be inclined to removed the <> and [] bits in this example for filename and index.  See the EXAMPLES section of samtools-view.1 which has a `-X` option.

If not, then `[`...`]` is misleading as it implies it's optional, but when using `-X` it's mandatory.

Suggestion:

```
This option will allow the user to specify a customised index file
location.  E.g.

.B samtools idxstats [options] -X /data/in.bam /index/in.bai                                           
```

whitwham: I was following the example in samtools-depth.1.

jkbonfield: That's not using angle brackets though and it doesn't have the mandatory components in square brackets (except where they're optional because the BAM filename was optional too).  Ie it's essentially `fn1 [fn2 ...] idx1 [idx2 ...]`. 

Granted we're not great on consistency though.

PR #2098, opened 2024-08-13
Document mpileup --adjust-MQ as best as I can

jkbonfield:
See samtools/bcftools#2217


daviesrob:
I've pushed some suggested formatting adjustments for you to take a look at:

* Wrap line that caused the table to overflow the line width in `T{ ... T}`

* Make bullets consistent with the rest of the document, and remove indents that broke line wrapping.  The original looked like an accidental switch into markdown...

* Wrap `--adjust-MQ` text in `.RS ... .RE` so the margin is consistent.

PR #2101, opened 2024-08-19
Add Wformat=2 and sanitize=undefined to the CI system

jkbonfield:
As requested.

Wformat=2 is a bit of a pain, but undefined behaviour sanitizer seems valid to use.  non-literal format checking is more questionable.  It *is* finding bugs, although oddly in things that it ought to have been spotting before already (so maybe this coincided with a compiler version update on mingw), but it also has false positives to work around: i.e. places that aren't demonstrably the wrong arguments but cannot be validated at compile time are still considered to be wrong.


bam_ampliconclip.c
```diff
@@ -145,6 +145,13 @@ int load_bed_file_multi_ref(char *infile, int get_strand, int sort_by_pos, khash
         }
 
         int cols_parsed = 0;
+
+// The construction of scanf_str based on the number of columns in use is
+// not compatible with -Wformat-nonliteral.  Short of lots of code duplication,
+// there is little we can do other than elide this warning for this piece of
+// code.  This is probably the lesser of two evils.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
```
daviesrob, line 154: An easier solution here would be to always pass the full format list to `sscanf()` and then use the return value to see how many of them it managed to fill out.  Checking against `num_columns` should ensure that it hasn't been confused by whitespace other than tabs.

jkbonfield:
I think I've finally beaten it into submission, and fixed several ampliconclip bugs related to parsing of BED files too which were instigated by attempting to work around some of the more aggressive warnings.   So not found as errors due to the warnings, but serendipity at play.

PR #2103, opened 2024-08-22
Work around CFLAGS overriding --enable-werror.

jkbonfield:
If we have CFLAGS set then ./configure --enable-werror has no impact, as it just sets this in CFLAGS rather than another variable.

In theory we should do ./configure --enable-werror CFLAGS=xyz. However that means parameterising the <<:*COMPILE template.  It's easiest to do this via an environment given the matrix, but we have to be careful to avoid it undoing itself again.  We also need to take care to not do --enable-werror when using ubsan.

This is a long-standing issue of removal of -Werror that's been present right since the first day we used cirrus-CI.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,8 +30,10 @@ htslib_compile_template: &HTSLIB_COMPILE
     cd $HTSDIR
     autoreconf -i
     if test "x$WERROR" != "xno"; then CONFIG_OPTS="--enable-werror"; fi
-    eval ./configure $CONFIG_OPTS --prefix=`pwd`/../inst CFLAGS=\"$CFLAGS\" || (cat config.log; /bin/false)
-    make -j3 -e install
+    eval ./configure $CONFIG_OPTS --prefix=`pwd`/../inst \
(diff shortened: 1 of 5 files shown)


daviesrob:
The `_CFLAGS` method is probably the way it ought to be done.  Will take a look...

jkbonfield:
I added that change as a separate commit as I still had it around.  As expected it's failing atm, but this is an issue of cirrus-ci and branches I think.  It treats PRs different from own branching.

daviesrob:
Comparing with HTSlib, it seems the problem was due to using `make -e`, so the environment variables were overriding the results from `./configure`.  I have a version that goes down that route which seems to work OK.

jkbonfield:
> Comparing with HTSlib, it seems the problem was due to using `make -e`, so the environment variables were overriding the results from `./configure`. I have a version that goes down that route which seems to work OK.

Interesting.  This seems to have been inherited from the old Travis config, all the way back to 050bc54db.

I'm not really sure what it's there as it's specifying the variable on the make command line rather than the environment, but maybe it's to pick up something else.  Did you try building without the `-e` to see what breaks?  I'm guessing it's necessary for things like CFLAGS and LDFLAGS, but we could control that another way potentially.

daviesrob:
Non `-e` version pushed...

PR #2106, opened 2024-08-26
Misc trivia: Avoid untracked test file and fix Makefile dependencies

jmarshall:
Running tests via `make test` leaves an untracked unignored file:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	test/mpileup/mp_2018.fa.fai
```

The other .fai index files in _test/mpileup/_ are committed rather than ignored, cleaned, and regenerated on each test run, so commit this one too. Also correct inadvertently executable .sam file.

Also the usual accumulated dependency trivia.


jmarshall:
Belatedly adding _test/mpileup/mp_2018.fa.fai_ does actually make it annoying to update to this commit if you've previously run the tests, and would interfere with bisecting and running the tests if you didn't also run `make testclean` between steps that crossed this commit.

So it might be better to change _test/mpileup_ to git-ignore and `testclean` these .fai files instead.

jmarshall:
The short answer is yes, I think. But really I suspect adding them was just a `git add *`-esque oversight when these test cases were added.

The goal is to stop them from appearing as untracked files. That can be done either by git-adding them, or by git-ignoring and `make testclean`-ing them (, or by always deleting them at the end of `make test` but that would be fragile and very tedious).

There is an additional benefit to having these files in git: an advantage of the git-adding approach is that you could notice the “change not staged for commit” if the output from indexing ever changed. But if one wanted to check that, one should really add a proper test comparing to expected output instead.

The disadvantage of now git-adding the file at this late date is that if the file already exists on current **develop** (because you've run the tests), updating to this commit produces “error: The following untracked working tree files would be overwritten by checkout” and you must delete it yourself first. The same thing would happen forevermore when bisecting across this commit, which would be quite annoying.

I expected that the same thing would happen in reverse if I instead removed the other _test/mpileup/*.fa.fai_ files. However (perhaps because they go from tracked to ignored from Git's point of view) it does not, and both updating to such a commit and back again and bisecting across it work just fine.

Hence I've replaced this PR with an alternative that nukes _.fai_ files from Git in this subdirectory. (I think there are still checked-in index files in other _test/_ subdirectories.)

PR #2109, opened 2024-09-05
Adjust -fsanitize=address,undefined test optimisation level and fix warnings

daviesrob:
Enabling sanitizers leads to long compilation times at high optimisation levels.  Testing shows that with the sanitizers, changing from `-O3` to `-Og` makes almost no difference to the `make test` time, but gives a significantly faster build.

On a local machine, the total times for `make -j 3 && make test` were:

| Setting | Time      |
| ------: | --------: |
|     -O3 | 6m13.677s |
|     -O2 | 5m11.763s |
|     -Os | 4m11.703s |
|     -Og | 3m24.097s |

In all cases `make test` took around 2m10s +/- 2s.

Using this lower optimisation level triggers a couple of "may be used uninitialized" warnings, where the compiler isn't able to work out that they are actually always set before use.  These are silenced by making it more obvious that they are set.


daviesrob:
.. and add `UBSAN_OPTIONS`, so undefined behaviour sanitizer stops with a stack trace if it finds anything.

PR #2110, opened 2024-09-11
Extend samtools sort -M to distinguish between mapped and unmapped files

### LINKED ISSUE #2105: Is minhash really sorted by coordinate?
On the PacBio Revio instrument we are sorting our `hifi_reads.bam` files with `samtools sort -M -R` to save disk space and improve file transfer time. The `SO:coordinate` results in a lot of confusion, as users think that the file is already aligned.

I could override the sort order using a new option and set it to `unknown` in my use case. If I were to implement this, would you accept such solution or do you have a different idea?

jkbonfield:
Files with some mapped data and some unmapped (in chr "\*") still get a sort order of "coordinate" and sub-sort of "coordinate:minhash" as before.  However input that is entirely unmapped with everything in chr "\*" now use "SO:unsorted SS:unsorted:minhash" to avoid confusion.

Fixes #2105

Note that due to moving the header manipulation the temporary files produced have the original sort order and not the new sort order, but this is irrelevant as their headers are unused and it's correct in the final merged file.  (They're also the source of some bugs, so they may at some go away entirely at some later date.)


doc/samtools-sort.1
```diff
@@ -148,6 +148,11 @@ option.  Note data compressed in this manner may need to be name
 collated prior to conversion back to fastq.
 .IP
 Mapped sequences are sorted by chromosome and position.
+.IP
+Files with at least one aligned record (being placed at a position on
+a chromosome) use the sort order "coordinate" with a sub-sort of
+"coordinate:minimiser".  Files entirely consisting of unaligned data
+use sort order "unsorted" with sub-sort "unsorted:minimiser".
```
jmarshall, line 155:
```suggestion
"coordinate:minhash".  Files entirely consisting of unaligned data
use sort order "unsorted" with sub-sort "unsorted:minhash".
```

jmarshall:
While we're here, I notice that _samtools-sort.1_ (line 78) says that `-M` produces output that cannot be indexed. I haven't tested it, but I don't see why this would be the case…?

bam_sort.c
```diff
@@ -3402,9 +3420,12 @@ int bam_sort_core_ext(SamOrder sam_order, char* sort_tag, int minimiser_kmer,
     // write sub files
     k = max_k = bam_mem_offset = 0;
     size_t name_len = strlen(prefix) + 30;
+    int placed = 0;
     while ((res = sam_read1(fp, header, b)) >= 0) {
         int mem_full = 0;
 
+        placed |= b->core.tid >= 0;
```
jmarshall, line 3427: Nice bit of conditional avoidance :smile:

doc/samtools-sort.1
jkbonfield: Oops, yes I'm forgetting my own code!  Thanks

bam_sort.c
jkbonfield: Lol yeah it's my usual coding style.   IOCCC learnt probably. ;-)

jkbonfield:
> While we're here, I notice that _samtools-sort.1_ (line 78) says that `-M` produces output that cannot be indexed. I haven't tested it, but I don't see why this would be the case…?

I think it's wrong and probably written forgetting that it can do partially mapped files still.

Rather curiously, `--write-index` does fail with an inappropriate "Ignoring --write-index as it only works for position sorted files" message.  A subsequent `samtools index` works just fine.  That's also a buglet of samtools sort.

It's not something we can determine up-front whether it is entirely unmapped, but actually that's not a problem anyway as we can still build an index and still query it via things like `idxstats` so it has some benefit.  Technically we could even still use the index for random access and sub-setting of the data, albeit non-trivially.  So there's no reason to forbid indexing of unmapped data.

(As an aside, bizarrely `samtools import` accepts `--write-index` but doesn't make an index.)

jkbonfield:
Also sadly, the decision of making minhash either coordinate or unsorted is making the index statement invalid.  Fortunately our index doesn't check (but it *could* and maybe some do).  However the man page for sort explicitly states we must use coordinate sort if we want an index, which isn't true as we can build an index on the unsorted data (because technically it's still coordinate sorted, as noted above).

I'll try picard.  Once I find a version of java that works.

jkbonfield:
Updated as a separate commit for ease of review.  I'll squash later (it won't make it into this upcoming release, sorry).

Also this talk of indexing reminds me of the original inspiration which came from Illumina's Tony Cox.

His code was computing a minimiser and actually having that minimiser as a coordinate on a fake chromosome (which sadly limits the size due to BAM indexing capabilities, unless we have many fake chromosomes).   An index then is obviously more sensible.  The benefit of this is then you can provide a query sequence, produce minimisers from that query, and then do random access to pull out sequences that share that minimiser.  It's a crude way of finding data that may potentially align to a query sequence without first assembling or aligning it all.  The hit rate wasn't that great though as you can only place a sequence in one location, so you can't have proper sequence sketches consisting of multiple minimisers.   That also limits it to short read data.

A cool idea though. :)   I just took the locality => better compression component of it and ran with that.

doc/samtools-sort.1
```diff
@@ -75,16 +75,15 @@ instead if you need name collated data without a full lexicographical sort.
...
+Thus the \fB-n\fR, \fB-N\fR and \fB-t\fR options are incompatible with
 .BR "samtools index" .
 
-When sorting by minimisier (\fB-M\fR), the sort order is defined by
-the whole-read minimiser value and the offset into the read that this
-minimiser was observed.  This produces small clusters (contig-like,
-but unaligned) and helps to improve compression with LZ algorithms.
-This can be improved by supplying a known reference to build a
-minimiser index (\fB-I\fR and \fB-w\fR options).
+When sorting by minimisier (\fB-M\fR), the sort order for unplaced
```
whitwham, line 81: I don't minimisier is a word.

jkbonfield: Indeed not.  Good spot.  That was in the original too.  Now corrected

PR #2123, opened 2024-10-08
Documents the fastq* options used in htslib's hts.c and sam.c

### LINKED ISSUE #2121: Fastq --import-format-options are not documented in the samtools.1 manual page

HTSlib's fastq support understands a [number of format options](https://github.com/samtools/htslib/blob/2ff207bd16bb3094843395450aa48936dd774ecb/hts.c#L1188).  Unfortunately they are not listed in the [fmt-option section](https://github.com/samtools/samtools/blob/eb0992ff8a99b895364a7a861b418a1beb77d540/doc/samtools.1#L886) of the samtools manual page, so they're not easy to use at the moment.

Please add documentation for these options to the manual page.

jkbonfield:
These were previously only defined within samtools import, and specified in terms of their associated command line parameters rather than the input and output format options.

Fixes #2121

### DIFF
--- a/doc/samtools.1
+++ b/doc/samtools.1
@@ -1044,6 +1044,47 @@ if it has been explicitly changed already.  Additionally
 \fBbases_per_slice\fR is set to \fB500*seqs_per_slice\fR unless previously
 explicitly set.
 
+.TP
+.B fastq_name2
+FASTQ input only.  Indicates that the names are not the first word in
+the header, but the second.  This is a FASTQ variant commonly used in
+the SRA and ENA archives.
+
+.TP
+.B fastq_casava
+FASTQ input and output only.  The Illumina CASAVA identifiers are
+stored in the second word of the FASTQ header lines and store read
+meta-data.  The CASAVA tag defines the data held in the READ1, READ2
+and QCFAIL flags and the barcode auxiliary tag ("BC" by default).
+This option may be used to both read and write CASAVA identifiers.
+
+.TP
+.BI fastq_barcode= TAG
+FASTQ input and output only.  When the \fBfastq_casava\fR option is
+used, this controls the name of the barcode aux tag to be used. \fITAG\fR
+defaults to "BC" if not specified.
+
+.TP
+.BI fastq_aux= LIST
+FASTQ input and output only.  Processes SAM format auxiliary tags
+following the other fields on the record identifier lines.  If no
+\fB=\fR\fILIST\fR is specified or \fILIST\fR is "1" then all aux tags
+listed are copied to/from the SAM record.  Otherwise it is a comma
+separated list of 2-letter tag types and is used to control which tags
+are processed with any others being omitted.
+
+Note as commas are used to separate options in the \fB--output-fmt\fR
+string detailing file format and options combined together, you will
+need to use the \fB--output-fmt-option\fR option if you want to
+specify a comma separated list of tag types.
+
+.TP
+.B fastq_rnum
+FASTQ output only.  If set, paired reads will have "/1" and "/2"
+appended to their read names.  This has no effect on unpaired reads.
+When reading FASTQ these suffixes are automatically detected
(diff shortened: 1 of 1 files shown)


doc/samtools.1
```diff
@@ -1044,6 +1044,47 @@ if it has been explicitly changed already.  Additionally
 \fBbases_per_slice\fR is set to \fB500*seqs_per_slice\fR unless previously
 explicitly set.
 
+.TP
+.B fastq_name2
+FASTQ input only.  Incdicates that the names are not the first word in
```
daviesrob, line 1049:
```suggestion
FASTQ input only. Indicates that the names are not the first word in
```

doc/samtools.1
```diff
@@ -1044,6 +1044,47 @@ if it has been explicitly changed already.  Additionally
...
+.TP
+.BI fastq_barcode= TAG
+FASTQ input and output only.  When the \fBfastq_casava\fR option is
+used, this controls the name of the barcode aux tag to be used. \fITAG\fR
+defaults to "BC" if not specified.
+
+.TP
+.BI fastq_aux= LIST
+FASTQ input and output only.  Processes SAM format auxiliary tags
+following the other fields on the record identier lines.  If no
```
daviesrob, line 1070:
```suggestion
following the other fields on the record identifier lines.  If no
```

doc/samtools.1
```diff
@@ -1044,6 +1044,47 @@ if it has been explicitly changed already.  Additionally
...
+\fB=\fR\fILIST\fR is specified or \fILIST\fR is "1" then all aux tags
+listed are copied to/from the SAM record.  Otherwise it is a comma
+separated list of 2-letter tag types and is used to control which tags
+are processed with any others being omitted.
+
+Note as commas are used to separate options in the \fB--output-fmt\fR
+string detailing file format and options combined together, you will
+need to use the \fB--output-fmt-option\fR option if you need to
```
daviesrob, line 1078: I think this reads better:
```suggestion
need to use the \fB--output-fmt-option\fR option if you want to
```

doc/samtools.1
```diff
@@ -1044,6 +1044,47 @@ if it has been explicitly changed already.  Additionally
...
+specify a comma separated list of tag types.
+
+.TP
+.B fastq_rnum
+FASTQ output only.  If set, paired reads will have "/1" and "/2"
+appended to their read names.  This has no affect on unpaired reads
+and it not change reading FASTQ as these are automatically detected
+and processed irrespective of the \fBfastq_rnum\fR option.
```
daviesrob, line 1086:
```suggestion
appended to their read names.  This has no affect on unpaired reads.
When reading FASTQ these suffixes are automatically detected
and processed irrespective of the \fBfastq_rnum\fR option.
```

daviesrob (commented):
A few typos need to be fixed.  The current ordering looks a bit random - maybe they would be better in alphabetical order?  Although I could be persuaded that `fastq_barcode` is better listed after `fastq_casava` given the relationship between them.

doc/samtools.1
jmarshall: Also “This has no **effect** …”

jkbonfield:
Updated text.  Thanks

jkbonfield:
> A few typos need to be fixed. The current ordering looks a bit random - maybe they would be better in alphabetical order? Although I could be persuaded that `fastq_barcode` is better listed after `fastq_casava` given the relationship between them.

The order was deliberate, in the order they are applied basically.

FASTQ headers have name first, so `fastq_name2`, then comes the CASAVA string if present, then aux tags.  I put `fastq_barcode` before `fastq_aux` as it refers to both CASAVA and general aux tags so it fits neatly between the two.  Basically I wrote them so if you read them naturally one after the other then they form a sensible ordering which relates to the file format.  If you're just trying to look up a specific one, then I suggest searching as frankly you'd need that anyway given the size of the man page.

PR #2162, opened 2024-12-24
cat man page update (for #2160)

### LINKED ISSUE #2160: samtools cat fails to accept -@ option
#### Are you using the latest version of [samtools](https://github.com/samtools/samtools/releases/latest) and [HTSlib](https://github.com/samtools/htslib/releases/latest)?
Yes (both v1.21).

- OS (`uname -sr`): `Linux 4.18.0-553.30.1.el8_10.x86_64`
- machine architecture (`uname -m`): `x86_64`
- compiler (`gcc --version`): `gcc (GCC) 11.3.0`

Running `samtools cat` with option `-@n` (replacing n with some INT number of threads, in my case 2) yields the following error:
```
cat: invalid option -- '@'
cat: invalid option -- '2'
```
followed by the usage/help message. Note that the usage message does _not_ mention the `-@` option, however, the man page does (`man samtools-cat`):
```
       -@, --threads INT
               Number of input/output compression threads to use in addition to main thread [0].
```

Using `samtools cat --threads 2` does not display the same error message, but it does fail to concatenate the files. This causes any subsequent command in a pipeline to see no input on its stdin. As a simple example, piping the output to `samtools view` yields the following (perfectly reasonable) error (`samtools cat --threads 2 -b bams.list | samtools view`):
```
[main_samview] fail to read the header from "-".
```

Notably, using `--output-fmt-option nthreads=2` does _not_ yield any errors and I do see output (though I tested with only a quick pipe to `head`, so I don't technically know if the full file would have worked).

Note that a similar issue was reported in 2020: #1336. It was addressed with pull request #1337, though these changes appear not to persist in v1.21.

If `samtools cat` is able to make use of additional threads, then some changes need to be made to the getopt argument parsing and possibly other places to pass the correct things to subsequent functions. If it is not able to make use of additional threads, then the manual page should be modified to at least remove `-@, --threads` and possibly also add a note that additional threads cannot be utilized here.

vasudeva8:
Fixes #2160
samtools cat does not make use of threads. Man page is updated to remote reference to threads option.


doc/samtools-cat.1
```diff
@@ -122,8 +122,6 @@ container but outside of the requested region.
 .BI --no-PG
 Do not add a @PG line to the header of the output file.
 .TP
```
daviesrob, line 124: This line needs to be removed as well, else the "EXAMPLES" header underneath ends up in the wrong place.

vasudeva8: thanks, the issue is visible in linux and now the indentation is removed.

PR #2164, opened 2025-01-07
Update samtools-merge.1 to include --template-coordinate

nh13:


daviesrob:
Strictly this should be `.B` as there's no argument so the `I` in `.BI` isn't needed...

This option might be better placed after `-t` so the ordering-based options are closer together.  It should also be mentioned in the "The ordering of the records in the input files must match..." paragraph.

I'll push a commit here to make the changes above.  There are some other improvements that could be made (e.g. moving `-o`) but they may be better done separately.

Issue #2167, opened 2025-01-14
Samtools consensus --min-depth option produces longer sequence


sanschiffre:
#### Are you using the latest version of [samtools](https://github.com/samtools/samtools/releases/latest) and [HTSlib]
samtools 1.21
Using htslib 1.21

 #### Please describe your environment.
Linux 5.14.21-150500.55.36-default
x86_64
gcc (SUSE Linux) 7.5.0

#### Please specify the steps taken to generate the issue, the command you are running and the relevant output.

When i used `samtools consensus BAM -r SC:1-100` i get a consensus of 100 bp
When i used `samtools consensus BAM -r SC:1-100 -d 70` i get a consensus of 130 bp

If a check the PILEUP format result 

Without -d
```
...
SC	28	0	53	G	157	...
SC	29	0	53	C	122	...
SC	30	0	53	G	154	...
...
```

With -d
```
...
SC	28	0	53	N	0	...
SC	29	0	53	N	0	...
SC	29	1	53	N	0	**************A*******###############################	...
SC	29	2	53	N	0	**************T*******###############################	...
SC	30	0	53	N	0	...
...
```

Low coverage insertion, ignored by default, seem to be taken into account with -d

jkbonfield:
I'm a bit surprised, as I thought `-d` was simply filtering out low depth cases, so why `-d 1` would miss things while `-d 70` doesn't I'm unsure.   I'm assuming it's something neighbouring which is skewing statistics somehow.  I'd really need to see the data to diagnose what's happening.

sanschiffre:
I only keep the first 56 reads in the sam. I can still reproduise the problem with them.
 [Data.zip](https://github.com/user-attachments/files/18410396/Data.zip)

The problem only appear when `-d X` remove all reads at a position.

PR #2174, opened 2025-01-20
Add multi-threading to samtools consensus. Needs htslib#1875

jkbonfield:
It's a major refactoring, but paves the way for similar work in mpileup. When an index is present, this breaks the queries down into continguous regions and performs consensus fasta/fastq/pileup generation in parallel.  Each thread independently operates on its own file descriptor and iterator.

This gives considerably higher multi-threading capabilities than the old system which simply performed parallel file decompression. 16 threads gives 12-13x speed increase.

More work needs to be done on reference sequence handling, but this is probably a task for htslib.  Ideally we need a dedicated thread-safe reference pool that permits ref-incr and ref-decr calls to share references between multiple threads and tracks those in active usage. (This exists in CRAM, but needs generalising.)

A few subtle changes occurred during the refactoring too.

- Mpileup and fasta/fastq consensus are now consistent in for heterozygous gaps that fail to meet the consensus cutoff.  We prefer to call the most likely base now, but will keep quality 0.

- The combination of region queries and whether or not to pad out with preceeding Ns for zero coverage is now consistent.  In some cases it used to always count from 1 (or region start) irrespective of where data coverage was, but did stop early if coverage didn't extend to the end.  This sort of half-way -a (all) mode no longer exists.

  Sequence reported is always first observed base to last observed, unless -a is used in which case it is 1 to chr length.


jkbonfield:
Draft while I do some final checking.  Some benchmarks as it currently stands:

```
$ for t in 0 4 16;do time samtools consensus -f fastq -C0 ~/scratch/data/novaseq.10m.cram -@$t -r 1 -o /dev/null;done
real	2m53.528s
user	2m53.172s
sys	0m0.340s

real	0m45.231s
user	2m57.295s
sys	0m1.177s

real	0m13.398s
user	3m19.251s
sys	0m1.249s
```

jkbonfield:
Also added a trivial one line change to fix #2167, as the code has moved and it'd just cause a conflict to make a new PR.  We can delay it until later if desired though.

(Plus documented a strange quirk of `--show-del yes` impacting insertion visibility in pileup.  That may be a bug - I'm unsure currently!  It's debateable, depending on your point of view as to what a deletion means.

PR #2178, opened 2025-01-27
Revise the REF_PATH / REF_CACHE documentation.

jkbonfield:
Updates to reflect the changes in the htslib/no-ebi-ref PR.

Also moved the section listing the reference discovery ordering to be adjacent to the environment variable section so all relevant text is together.

### DIFF
--- a/doc/samtools.1
+++ b/doc/samtools.1
@@ -1136,34 +1136,6 @@ and 0 or 1 also reduces some error messages, while values greater than 3
 produce increasing numbers of additional warnings and logging messages.
 
 .PP
-.SH REFERENCE SEQUENCES
-.PP
-The CRAM format requires use of a reference sequence for both reading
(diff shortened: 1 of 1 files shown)


jkbonfield:
For @daviesrob: one thing you may wish to replace is "If that is done, it is strongly encouraged you also specify a local REF_CACHE
directory" with wording appropriate for your new ref-cache tool.  IMO using that is preferable over using REF_CACHE now.

The bit about reference sequences and CRAM is incorrect in saying CRAM requires a reference, but I didn't change that text - only punted it lower down in the document.  I don't think it's worth getting overly worried about that though as it's how 99.something% of CRAMs are produced.  While I think on-the-fly embedded consensus refs are a better solution for many people, they're not really well supported by the archives (I think) and until that happens people will just go with the standard workflow.

PR #2203, opened 2025-03-20
Added fish completions

LunarEclipse363:
## Introduction

This PR adds comprehensive completions for use with the [fish shell](https://fishshell.com/).

This includes:
- finding all subcommands and their descriptions from the output of `samtools`
- if `python3` is available and samtools manpages are installed, generates comprehensive completions with descriptions for the subcommands
- otherwise generates barebones completions for subcommands from the outputs of `samtools help $subcommand` (slightly better than the already provided bash completions)

I also added an entry in the changelog.

## Demonstration of the full capabilities
image
image

## Notes
I initially tried using the fallback approach as the primary one but after half a day of work realized that the `samtools help` outputs aren't consistent enough to write a regex to parse them reliably and which could extract all the data (all versions of each flag, what kind of value they take if any, descriptions).  
Or if it can be done it's so complex I decided it's now worth it.  
Doing that somewhat reasonably would require enforcing a strict style guide for help pages like "exactly one space between a flag and its value" and "at least two spaces between the flags/values and descriptions" and I feel like that's not feasible, so I went for extremely limited parsing in this regard and instead used a script bundled with fish to generate high quality completions from manpages (somehow this only worked well for the subcommands, not the main samtools command).

In any case, I'm quite happy with how this turned out and hope that I've fixed all the little bugs in it by now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,7 @@ bindir      = $(exec_prefix)/bin
 datarootdir = $(prefix)/share
 mandir      = $(datarootdir)/man
 man1dir     = $(mandir)/man1
+fishcompdir = $(datarootdir)/fish/vendor_completions.d
 
 # Installation location for $(MISC_PROGRAMS) and $(MISC_SCRIPTS)
 misc_bindir = $(bindir)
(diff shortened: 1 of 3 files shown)


jmarshall:
There's also bash completions already in _misc/samtools_tab_completion_ (which the maintainers might consider renaming to _misc/samtools_tab_completion.bash_).

If installing this fish completion is added to `make install` then steps to install the bash completion should be added too. (Probably the destination is `$(datarootdir)/bash-completion/completions` but it's not super clear as this is a convention from bash-completion, not bash itself.)

However my suggestion might be to add this fish script to _misc/_ but continue to not install these completions in `make install`, install leaving that to distribution packagers.

vasudeva8:
Thanks for the completion file, I have merged it.
We don't want to update the makefile, we leave it to the users.

PR #2213, opened 2025-04-22
Miscellaneous tidy ups

jmarshall:
Miscellaneous and very squashable trivia: spelling, one minor usage formatting change, _Makefile_ dependencies, and ignoring and cleaning some recent temporary test output files.

I was lazy and made bespoke rules for _test/consensus_ but could do some tedious renaming in _test/consensus/consensus.reg_ instead if that was preferred.


vasudeva8:
Thanks John, I have reinstated a newline in sam view usage (6822a89) and merged it.

jmarshall:
Thanks, but note that — as per the commit message — that newline was intentionally removed:

> Reinstate a blank line before "Input options", which previously had a visual blank line due to the blank left column under "-p, --unmap".
>
> Remove the apparent double blank line before "General options", which no longer needs a real blank line due to the blank left column under "-z, --sanitize".

vasudeva8:
It was reinstated to have same format as other section headers in the usage, that any automated processing can separate sections.

jmarshall:
That argument does not hold water. Prior to this PR, “Input options” has not been set off by a blank line since this sectioned usage display was introduced in 1.13 nearly five years ago. So any automated processing will not be depending on this.

PR #2222, opened 2025-05-19
Fix URL to online documentation in `seq_cache_populate.pl`

charles-plessy:
<https://www.htslib.org/workflow/#the-ref_path-and-ref_cache> returns 404 but <https://www.htslib.org/workflow/cram#the-ref_path-and-ref_cache> works…


daviesrob:
Thanks for the fix.  The linked-to page probably needs some updates, but it's useful to have a link that works in the first place...

PR #2224, opened 2025-05-22
reference status in to stats output (fixes #2139)

### LINKED ISSUE #2139: samtools stats reference information
Would it be possible to have some stats on the references (target sequences) also?
I think it would be nice to have, at least, the number/list of target sequences (from the header of the BAM file)

Optional, if the ref file is also provided, one could also have other metrics (e.g.):
- GC (or avg)
- length (or sum/min/max/avg)
- number of undetermined bases

vasudeva8:
Fixes #2139.
Reference status is calculated and added to stats output.
It calculates the length from header and region specifications and doesn't need reference file for it.
GC content and unknown base counts are calculated when ref. file is provided and will have -1 when it is not provided.

1st line in RFS section gives the total sequence count in input file, count of regions in status, average GC, min, max, average and total counts.
2nd line onwards gives regions, lengths, GC and unknown base count.

Regions are selected based on options, it is given by target when -t is used and given by region specification when -t is not there. All targets in input file is used when neither -t and region spec are present.


vasudeva8:
Based on review, removed the unnecessary option argument in ref-stats.
Also removed the gap in test output names.

vasudeva8:
Updated man page based on review suggestions.
Added thread pool to fai file handling, to improve performance.
Updated base counting for GC content based on discussions, to improve performance.

PR #2233, opened 2025-06-30
removes soft clips before saving as fasta/q

### LINKED ISSUE #1275: samtools fastq - trim reads based on soft clipping

As i understand it, samtools fastq currently allows outputs reads "as-is" from the BAM. If they were hard-clipped then they are already existing in clipped form and output as such, but if they are soft-clipped, and i want the clipped reads out, I can not get them (?)

Would like a new option for samtools fastq to obvserve the soft clipping in the CIGAR and output the shorter clipped reads.

vasudeva8:
Fixes #1275 
Removes bases corresponding to soft clips. CIGAR and quality values are updated accordingly.
The modified data are stored in aux tags with tag 's0', as 'Z' type array. The aux tag can be controlled with --sc-aux option.
If such aux tag storing is not required, it can be disabled with --no-sc-bkp option.
With reads with reverse flag, the output is reversed and flipped, and the same is done with dumped aux tags as well; the cigar, bases and quality values are reversed and flipped as required.

No other aux tag is updated and at the end of soft clip removal, other aux tag may not be in sync with soft-clipped data if they are about removed data.
Doesn't handle very large cigar tags in bcf aux tag, just uses the cigar in bam.


GWBarberis:
> We have a couple of pieces of work to finish off before a new release. Hopefully if shouldn't be long.

Great, thank you for your prompt reply.

PR #2275, opened 2025-10-14
Adding badges for usegalaxy.org and usegalaxy.eu

nekrut:
This badge is linked to <tt>samtools view</tt>, which was used close to 1M times of usegalaxy.eu and over 5M times on usegalaxy.org (eu usage is [here](https://stats.galaxyproject.eu/d/DYDPhX-Sk/tool-usage-stats?orgId=1&from=now-6h&to=now&timezone=browser&var-tool_id=samtools_view))

Thank you for these!


nekrut:
If you are planning a separate page with explaining where to get samtools, let me know. We can add more info there.

PR #2287, opened 2025-11-25
Ignore hard clips in template coordinate sort

nh13:
Hard clipping may be variable for reads from the same biochemical template, causing them to sort to different positions. Only soft clips are now considered when calculating unclipped coordinates for template coordinate sorting, matching fgbio behavior.

See: https://github.com/fulcrumgenomics/fgbio/pull/1065

The unclipped_start/end functions now take an include_hard_clips parameter. Template coordinate sort passes 0 (soft clips only), while markdup passes 1 to preserve existing behavior.

Adds test for hard clip handling in template coordinate sort.

### DIFF
--- a/bam.c
+++ b/bam.c
@@ -157,8 +157,10 @@ rmB_err:
     return -1;
 }
 
-/* Calculate the current read's start based on the stored cigar string. */
-hts_pos_t unclipped_start(bam1_t *b) {
+/* Calculate the current read's start based on the stored cigar string.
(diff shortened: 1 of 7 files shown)


clintval:
Hi @jkbonfield we merged the companion PR for this change in fgbio and I'm wondering if you'd have the time to review this PR in samtools so both implementations behave similarily?

We had a user issue this week which highlighted that we should be ignoring hard-clipping operators when determining "unclipped ends":

- https://github.com/fulcrumgenomics/fgbio/issues/1141

PR #2291, opened 2025-12-17
Fix variable expansion error in seq_cache_populate.py

rhpvorderman:
In an effort to mimick `seq_cache_populate.pl` the output messages are kept the same as much as possible. Unfortunately I missed a perl variable expansion back then.


whitwham:
Thank you.

Issue #2293, opened 2026-01-02
mpileup on extremely high depth region


gevro:
Hi, samtools v1.20 mpileup on an extremely high depth region (> 100,000 reads) takes an extremely long time even when setting the -d (max depth) option to a very low number such as '10'. This doesn't make sense, since the first 10 reads from the region should be sufficient and output should be immediate.

gevro:
I'm not doing any options, just samtools mpileup. I tried with -B as well, but that did not speed it up. I think it is getting stuck somewhere in the parsing of the index file.

jkbonfield:
Just to be explicit on what we believe to be the cause of speed issues:

#2301 sped up generation of textual mpileup data and printing it out, which is a general speed improvement.  However it doesn't really impact deep data as the amount of text to print is small.

We believe the problem you are seeing is simply how long it takes to decode SAM/BAM/CRAM.  In the general case of streaming through a large file with a shallow-deep-shallow pattern, even if we only read 10 reads out of 100,000 in that deep section, we cannot simply skip ahead to get to the following shallow bit.  Arguably it may be possible with an index, but it wouldn't work when streaming and the index resolution is highly unlikely to be fine-grained enough when not streaming.

So how it works is basically a read-decode-and-discard loop.  This speeds up the mpileup data rotation transformation as well as printing up of the results and reducing memory, but we're still left with the decoding of the alignment records.  Our benchmarking showed on deep regions with depth filtering that was the bottleneck.

You should be able to test it by timing `samtools mpileup -d 10` on a small highly deep region versus `samtools view -F 0xffff` on the same region.  (The `-F` flag discards all reads.)

gevro:
Thank you! Impressed you all dug so deep into this. Will the #2301 fix resolve this and will it be incorporated into an upcoming samtools version?

jkbonfield:
As I said above, it's unlikely that #2301 will fix your problem, however it will slightly alleviate it due to being a general across the board speed increase to mpileup textual output.

Unless you can provide data showing otherwise, we believe the speed issue you are seeing is simply the time taken to decode (and discard) that many alignment records.

gevro:
Understood. I agree the speed issue must be just the decode/discard cycle. I don't have data arguing otherwise.

PR #2301, opened 2026-02-05
mpileup changes to use kstring for speed up

vasudeva8:
Changed direct file writes to kstring buffer and wrote it at end of processing for each line (main contributor in improvement).
Updated reference data retrieval function invoke to execute only on tid change (not a big contributor but gets invoked on every read).
Also there is an htslib update, to prefetch next data to use.

This showed 10 - 30% improvement with different test data.

Attempts to improve #2293


vasudeva8:
Updated based on review comments and corrected an fwrite as well.

PR #2308, opened 2026-02-14
Add support for single-end FASTQ imports with CASAVA tags for BC auxtag

dkj:
Partial fix for #2307 : respects the `-0` over read single/1/2 inference from `-i` parsing.

Generated with Claude Opus 4.6 . Looks good to this human.


whitwham:
Currently -0 lets itself be overridden by /1, /2 and/or the CASAVA tag (this is in the man page but not anywhere else).  As CASAVA has no way of declaring that it is single ended (the 1 meaning single or the first of a pair) then letting -0 do the setting is probably a good change.  It does need to fail if it comes across a 2 just so people are not accidentally trying to force paired end reads into a single format.

Issue #2181, opened 2025-02-04
samtools markdup: move UMI from read name to RX tag


matthdsm:
Currently, `samtools markdup` allows for duplicate marking based on UMI's in the read name, which is a great feature!
It would be nice though if there were a flag that _edits_ the read name and moves the UMI into the appropriate tag.

#### Describe the solution you would like.
Add an `--add-umi-tag` flag that works in conjunction with the `--barcode-name` flag that takes the UMI from the read name and puts it in the appropriate `RX` tag.

Please do not hesitate o tell me to RTFM if this option already exists.

Cheers
Matthias

whitwham:
There have been a few good suggestions (like this one) to add to the UMI handling in markdup.  I am not sure markdup is the best place to put a lot of them.  We might need either an addition to fixmate to preprocess UMIs or an entirely new tool dedicated to UMI fixing and manipulation.

So yes, it would be good to do something like this but I am not sure we have decided quite how to go about it.

matthdsm:
Hi!

Is there an update on any possible progress on this? We kind of really need this, but don't have the knowledge to do it ourselves 😐 

I'd loathe to have to use an unnamed java based tool for this....

whitwham:
I am sorry but we have not looked at this yet, we have had other commitments.

It is not that hard in itself (I think) but I am still not sure where it belongs amongst the samtools tools.

matthdsm:
Would you mind a (AI assisted) PR that adds the initials feature requested?

whitwham:
> Would you mind a (AI assisted) PR that adds the initials feature requested?

I will definitely take a look at it (as long as it isn't 50,000 lines of AI generated code).

whitwham:
Fixed by #2316.

PR #2316, opened 2026-03-19
feat: Add --move-umi-to-tag option to samtools markdup to extract UMI from QNAME and add to RX tag.

matthdsm:
Proposed fix for #2181
 cc @whitwham


whitwham:
It really needs to pass all its tests before we look at it.

matthdsm:
Sorry about that, tests pass now

bam_markdup.c
```diff
@@ -2103,6 +2192,10 @@ static int bam_mark_duplicates(md_param_t *param) {
                     uint8_t* data = bam_aux_get(b, "dc");
                     if(data) bam_aux_del(b, data);
                 }
+                if (param->move_umi) {
+                    if (move_umi_to_tag(param, b) < 0)
+                        goto fail;
+                }
```
whitwham, line 2203: All of the reads should have been altered by this point so this is not really necessary.

whitwham:
The code only works properly if the UMI is at the end of the read name but not if it is anywhere else.  The regex can pick up the UMI wherever it is in the QNAME.  Now, I think all the examples of embedded UMIs that I have seen have been at the end but I have absolute faith that some researcher somewhere has created files where the UMI is in the middle (because of reasons).  In that case the QNAME will be truncated from the beginning of the UMI.

The code assumes the separator is a `:` for removal purposes which, again, is usually the case but does not have to be.  Maybe another option to set the separator and having it default to colon.

Finally, the man page needs updating with the new options.

matthdsm:
>  Now, I think all the examples of embedded UMIs that I have seen have been at the end but I have absolute faith that some researcher somewhere has created files where the UMI is in the middle (because of reasons).

.. Did you visit our lab at some point? You seem to have met several people here 🤣 

Jokes aside, thank for reviewing. I'll look into fixing the issues.
Also, feel free to just say "No" if this PR or feature doesn't fit in the roadmap for samtools. I've been thinking about this and I do agree there should be some dedicated UMI manipulation subcmd at some point.

There's been some development over at fulcrum genomics for [a rust based toolkit to replace the java one](https://github.com/fulcrumgenomics/fgumi). It might serve as inspiration.

whitwham:
> Also, feel free to just say "No" if this PR or feature doesn't fit in the roadmap for samtools. I've been thinking about this and I do agree there should be some dedicated UMI manipulation subcmd at some point.

Having something that works now is better than waiting for something perfect in the indeterminate future.  We do plan on doing a UMI subcommand but between bug fixing and other commitments I don't know when that will be.  So thank you for doing this.

> There's been some development over at fulcrum genomics for [a rust based toolkit to replace the java one](https://github.com/fulcrumgenomics/fgumi). It might serve as inspiration.

I should, especially since @nh13 is a contributor here.

bam_markdup.c
matthdsm: Removed the snipped

matthdsm:
Thanks for reviewing @whitwham. I've addressed the comments. Let me know what you think

whitwham:
Thanks @matthdsm I will take a look when I get back to work in a couple of weeks time.

whitwham:
Looks good @matthdsm.  It is failing the extra spaces maintenance check, though I can tidy that up for you if you want.  After that it will be ready to merge.

matthdsm:
Thanks Andrew, feel free to amend as you see fit. I'm already happy you're considering merging.

whitwham:
Thank you.

PR #2317, opened 2026-03-23
Option Fixes

whitwham:
In stats.c: removed the mistaken _required_argument_ entry from `--customized-index-file`.  This makes the long and short (`-X`) option work in the same way.  Fixes #2315.

Also removed some unused and undocumented options from bam_plcmd.c.


bam_plcmd.c
```diff
@@ -1134,15 +1134,9 @@ int bam_mpileup(int argc, char *argv[])
         {"output-bp-5", no_argument, NULL, 14},
         {"output-MQ", no_argument, NULL, 's'},
         {"output-mq", no_argument, NULL, 's'},
-        {"ext-prob", required_argument, NULL, 'e'},
-        {"gap-frac", required_argument, NULL, 'F'},
-        {"tandem-qual", required_argument, NULL, 'h'},
         {"skip-indels", no_argument, NULL, 'I'},
-        {"max-idepth", required_argument, NULL, 'L'},
-        {"min-ireads ", required_argument, NULL, 'm'},
         {"per-sample-mF", no_argument, NULL, 'p'},
```
vasudeva8: We are not using skip-indels/'I' and per-sample-mF[f]/'p', can those be removed as well?

whitwham: Done!

PR #2319, opened 2026-04-06
markdup: O(n²) → O(n) optical duplicate detection via grid spatial index

### LINKED ISSUE #2296: samtools markdup memory usage on high-depth regions
samtools 1.22.1

When markdup hits high-depth regions, it can use a lot of memory as it keeps hold of all records overlapping the current one.  On a particularly bad input we have, where the depth was over 120 million, this caused processing to fail due to running out of memory.  The problematic region has been extracted from this for testing purposes and will be shared separately.

HeyangQin:
## Summary

Fixes #2296.

`check_duplicate_chain` previously compared every pair of duplicates at the same genomic position to find optical duplicates (those within `opt_dist` on the flowcell). This is O(n²) and causes runaway runtime/memory on high-depth regions — the issue reports 120M+ reads at a single position crashing clinical WGS pipelines.

This PR replaces the pairwise scan with a **grid-based spatial hash**:
- Flowcell coordinates are bucketed into cells of size `opt_dist`
- For each point, only the 9 neighbouring cells (own + 8 adjacent) are queried
- Two points in non-adjacent cells are provably > `opt_dist` apart (Chebyshev distance), so no valid pairs are missed

The duplicate resolution logic (score comparison, QC-fail handling, paired vs single, name tiebreaking) is unchanged.

## Complexity

| Case | Before | After |
|------|--------|-------|
| Typical (coordinates spread across flowcell) | O(n) amortized | O(n) |
| Worst case (all x within opt_dist) | **O(n²)** | **O(n)** expected |

## Benchmarks

Worst-case synthetic data: all reads from one tile, all x-coordinates within `opt_dist=2500`, y spread across 0–100k. Measured wall-clock time on Linux x86_64.

| Duplicates | Before | After | Speedup |
|-----------|--------|-------|---------|
| 5,000 | 0.27s | 0.20s | 1.3× |
| 20,000 | 2.2s | 0.67s | 3.3× |
| 50,000 | 11.6s | 2.8s | 4.1× |
| 200,000 | 3m 45s | 54s | 4.2× |

The original scales quadratically (4× input → ~15× time); the grid scales linearly (4× input → ~4× time). At the 120M scale in the issue, the original is computationally intractable while the grid completes in bounded time.

## Correctness

- All 11 existing optical duplicate tests pass (tests 8–16 + regex/barcode/read-group variants)
- Output is **byte-identical** to the original implementation on both spread-out and worst-case synthetic data, verified up to 200k duplicates

## Test plan

- [x] All existing `markdup` optical duplicate tests pass (tests 8–16, including regex, barcode-tag, barcode-name, barcode-rgx, read-group variants)
- [x] Byte-identical output vs original on worst-case 50k and 200k synthetic data
- [x] Byte-identical output vs original on spread-out 50k synthetic data
- [x] Non-optical markdup tests unaffected (tests 5–7)

### DIFF
--- a/bam_markdup.c
+++ b/bam_markdup.c
@@ -235,6 +235,14 @@ KLIST_INIT(read_queue, read_queue_t, __free_queue_element) // the reads buffer
 KHASH_MAP_INIT_STR(duplicates, dup_map_t) // map of duplicates for supplementary dup id
 KHASH_MAP_INIT_STR(read_groups, int) // read group lookup
 
+typedef struct {
+    size_t *idx;
+    size_t count;
+    size_t cap;
+} grid_cell_t;
+
+KHASH_MAP_INIT_INT64(grid, grid_cell_t) // spatial grid for optical dup detection
+
 /* The Bob Jenkins one_at_a_time hash to reduce the key to a 32 bit value. */
 
 static khint32_t do_hash(unsigned char *key, khint32_t len) {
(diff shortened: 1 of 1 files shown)


whitwham:
Thank you for your try but this does not fix #2296 and it does not address that problem in any way.  It may solve a different problem (and the duplicate chain code does need improvement) but that would be a different PR.

whitwham:
Testing this on real data (80 million reads, 5.5 million duplicates of which 4.4 million were optical) showed no real difference between this code and the original in time or resource usage.

At this time I do not think it is worth adding to the code complexity unless this worse case data starts cropping from the sequencing machines.

PR #2320, opened 2026-04-11
Fix for the footgun of repeated use of "samtools view --subsample" with the same (or default) seed

### LINKED ISSUE #1201: bug when downsampling twice with the same seed
This is the culmination of weeks of hunting this bug down..
In the following sequence, the second command will not sub-sample, instead it will output the input bam unchanged (except for an additional header tag):

    samtools view -s .01 -b -o ss1.bam a.bam
    samtools view -s .5 -b -o ss2.bam ss1.bam # doesn't work

The second command will work if the seed is changed, e.g.

    samtools view -s 123.5 -b -o ss2.bam ss1.bam # works

This is using samtools 1.10

tfenne:
Changes the default seed from 0, to a value derived from the hash of the incoming file's header. The idea is that the default seed should be deterministic given the same input, but different for different inputs.  This ensures the seed is the same regardless of format (SAM/BAM/CRAM) or compression level, but will generally cause the seed to vary for files that have been processed differently due to the differing @PG lines (and possibly others) in the header.  Since "samtools view" will insert a @PG line into the header, serial downsampling of file A to file B to file C will now use different seeds at each step.

I think this is my first PR in samtools, and this changes the a command line default, so I am mindful that this may not be accepted, or not accepted without changes.  I'm very open to feedback here, but would love to fix this footgun, and this seems like a reasonable way to do it.

Fixes #1201, which is something I also run into once in a while when I forget about the default behavior.

### DIFF
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -470,10 +470,14 @@ Subsampling seed used to influence
 subset of reads is kept.
 .\" Reads are retained based on a score computed by hashing their QNAME
 .\" field and the seed value.
-When subsampling data that has previously been subsampled, be sure to use
-a different seed value from those used previously; otherwise more reads
-will be retained than expected.
-[0]
+By default, the seed is derived from a hash of the input file's header text.
+Since each subsampling run adds a @PG header line, re-subsampling a file will
+automatically produce a different seed value, ensuring the expected fraction
+of reads is retained at each step.
+When subsampling data that has previously been subsampled using an explicit
+seed, be sure to use a different seed value from those used previously;
+otherwise more reads will be retained than expected.
+[hash of header text]
 .TP
 .BI "-s " FLOAT
 Subsampling shorthand option:
--- a/doc/samtools-view.1
+++ b/doc/samtools-view.1
@@ -464,20 +464,28 @@ Output only a proportion of the input alignments, as specified by 0.0 \(<=
 This subsampling acts in the same way on all of the alignment records in
 the same template or read pair, so it never keeps a read but not its mate.
 .TP
-.BI "--subsample-seed " INT
+.BI "--subsample-seed " INT | auto
 Subsampling seed used to influence
 .I which
 subset of reads is kept.
 .\" Reads are retained based on a score computed by hashing their QNAME
 .\" field and the seed value.
-By default, the seed is derived from a hash of the input file's header text.
-Since each subsampling run adds a @PG header line, re-subsampling a file will
-automatically produce a different seed value, ensuring the expected fraction
-of reads is retained at each step.
+If the value is the literal string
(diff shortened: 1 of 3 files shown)


tfenne:
CC @yfarjoun who I know has also been bitten by this recently

daviesrob:
I link the idea, although I'm a bit less keen on the change in behaviour.  Maybe it would be better to make enabling this more implicit, for example by using the string "auto" as the seed value?  That could also be made to work with the `-s` option, so `-s auto.1` would be the same as `--subsample 0.1 --subsample-seed auto`.

This does rely on not using the `--no-PG` option so that the header changes.  Rather than relying on that, it could append a `@CO` comment tag with text like "Sub-sampled using seed 1234".  That would both ensure the header does change, and also document the seed values that were actually used.

I see you've used `__ac_X31_hash_string()` to make the seed.  That's not a very good hash function as it has fairly poor diffusion, although the value does get passed through `srand() / rand()` later which may help a bit.  A better alternative would be [`__ac_FNV1a_hash_string()`](https://github.com/samtools/htslib/blob/cb88c90b4620bf4c910b9c3b60e9738c48b09ddb/htslib/khash.h#L466), which has better mixing and runs just as fast.

tfenne:
Thanks for the feedback @daviesrob - I'm happy to make those changes.  Are you ok with "auto" becoming the new default for the seed, or are you thinking it would be opt-in?  I can see both perspectives (it's nice not to alter default behavior) but I come down on the side of having "auto" be the default so users don't have to _find_ the footgun first in order to avoid it.

daviesrob:
In general we try not to change behaviour if possible, even though it means some of the defaults are a bit off (see `samtools mpileup`).  It is occasionally tempting to make a `samtools2` so we could reset everything to a more sensible starting point...

I can see the benefit of avoiding the foot-gun by default, although it would only happen for those using `--subsample` as the `-s` option always sets a seed at the moment.

I've also noticed that before version 1.9, `-s` didn't do much checking for valid input.  Using something other than a number would result in a fraction of 0, and so no output.  Hopefully that would be noticed if anyone tried it.

I'll have a think about this...

tfenne:
Thanks again @daviesrob - I've taken a shot at addressing your feedback.  I've left "auto" as the default, because without that I feel like there's a good chance most users are still exposed to the existing footgun.  That said, if you like the rest of the PR but want the default to remain `0`, just say the word and I'll update the default and the docs.

daviesrob:
The new changes look good, thanks.  On reflection, I think the "auto" default is probably the best option, given that it's reasonably easy to get the old behaviour back.  We'll just have to highlight it in the release notes.

tfenne:
Thanks @daviesrob !

PR #2326, opened 2026-04-20
Change nquals from 256 to 257 for samtools stats.

jkbonfield:
The maximum quality in ASCII is 94 (127-33) so anything else is outside of the legal range.  However even if we wanted to handle illegal BAM and ignore the unprintable nature of SAM, using 256 is a bad choice.

The reason is down to how CPU caches work.  They're typically an N-way associative hash when each hash bucket can cache results for N addresses and the hash index is just the last bits of the memory address.  This means a power of 2 in the memory address will put a lot pile many addresses into the same bucket, saturating "N".

A trivial workaround is to multiply by 257 instead of 256, causing the bottom bits of the address to not repeatedly collide.  95 is a better multiplier however as it uses less memory and hence with it further reduces cache miss rates.

Benchmarks from perf stat:

nquals=256:

         267540049      cache-misses              #   11.334 % of all cache refs
        2360568687      cache-references

      20.572468682 seconds time elapsed

nquals=257:

          78089279      cache-misses              #    3.391 % of all cache refs
        2302704223      cache-references

      15.710843218 seconds time elapsed

nquals=95:

          45394053      cache-misses              #    2.102 % of all cache refs
        2159597370      cache-references

      13.713247565 seconds time elapsed

Quite a good speed up for such a tiny change.
*However* setting nquals=95 isn't sufficient as qual "*" is an array of 255 values in BAM (sigh).  So it needs to be >= 256.  We could fix this by mapping 255 to 95 say or better still using a generic mapping table based on real data which would vastly reduce memory for NovaSeq, but this needs buffering and multi-pass, or the ability to add new entries to the lookup table and shuffle around existing entries.  Hence this PR is the trivial win only.

The actual code in CPU cycles for the original was:

```
  0.12 │       addsd    0x218(%r15),%xmm0                                                                        
       │     quals[ i*stats->nquals+qual ]++;                                                                    
  0.22 │       addq     $0x1,(%rdx,%r8,8)                                                                        
       │     stats->sum_qual += qual;                                                                            
 52.52 │       movsd    %xmm0,0x218(%r15)                                                                        
       │     for (i=0; i<seq_len; i++)                                                                           
```

With nquals=95 the change it is:

```
(as above)
 35.44 │       movsd    %xmm0,0x218(%r15)
```

(perf stats -n showed 37176 counts to 18470 counts).  I didn't record it for nquals=257 as I didn't notice the qual "*" issue until later, however the elapsed time is the key indicator.

### DIFF
--- a/stats.c
+++ b/stats.c
@@ -1613,7 +1613,8 @@ void output_stats(FILE *to, stats_t *stats, int sparse)
 
     int ibase,iqual;
     if ( stats->max_len<stats->nbases ) stats->max_len++;
-    if ( stats->max_qual+1<stats->nquals ) stats->max_qual++;
+    if ( stats->max_qual+1<stats->nquals && stats->max_qual<255 )
+        stats->max_qual++;
     fprintf(to, "# First Fragment Qualities. Use `grep ^FFQ | cut -f 2-` to extract this part.\n");
     fprintf(to, "# Columns correspond to qualities and rows to cycles. First column is the cycle number.\n");
     for (ibase=0; ibase<stats->max_len_1st; ibase++)
@@ -2318,7 +2319,7 @@ stats_t* stats_init(void)
         return NULL;
 
     stats->ngc    = 200;
-    stats->nquals = 256;
+    stats->nquals = 257; // NB: 95 is better, but need to handle qual "*"
     stats->nbases = 300;
     stats->rseq_pos     = -1;
     stats->tid = -1;
--- a/stats.c
+++ b/stats.c
@@ -190,7 +190,7 @@ typedef struct
     int nindels;        // The maximum indel length for indel distribution
 
     // Arrays for the histogram data
-    uint64_t *quals_1st, *quals_2nd;
+    uint64_t **quals_1st, **quals_2nd;
     uint64_t *gc_1st, *gc_2nd;
     acgtno_count_t *acgtno_cycles_1st, *acgtno_cycles_2nd;
     acgtno_count_t *acgtno_revcomp;
@@ -205,6 +205,7 @@ typedef struct
     int max_len_1st;        // Maximum read length for forward reads
     int max_len_2nd;        // Maximum read length for reverse reads
     int max_qual;           // Maximum quality
+    int max_qual_alloc;     // Size of quals_1st / quals_2nd
     int is_sorted;
 
     // Summary numbers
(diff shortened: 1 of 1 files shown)


jkbonfield:
Timings for a gcc -O2 build on 3 different data sets and different variants of this code.

```
                HiSeq          PB-CCS         NovaSeq
v123            15.96           18.72           24.14
PMAJOR          15.30           15.03           23.86
QMAJOR          15.29           12.37           23.13
QUAL2           15.11           11.94           22.65         
```

PMAJOR is the first commit: a 2D array implemented as a 1D lookup with position first (major column).
QMAJOR is the same code but the maths flipped around so quality values are the major column.
QUAL2 is the result of the second commit, implementing quals as a 2D-style array via an array of pointers.

Arguably this speed up isn't worth as much as it appears on long read data as frankly the output shouldn't be created in this manner anyway.  It's gets exceptionally unwieldy and worthless to see base-resolution quality plots, and the error rate isn't decaying unlike, say, Illumina cycles anyway so binning is considerably more useful.  So there's something to be said for using the easy win only.

jkbonfield:
Note this really is just low hanging fruit.  It's not hard to speed up this by another 10% or so with more work.

Eg why is it that nquals is set to 256 (or now 257) as a variable, but there is no way the user can change it?  Also it's bigger than the maximum value for a uint8_t, so why is it checked for in code?  This isn't optimised away as it's a struct member rather than a constant value.

I think it just shows a lot of legacy and it needs a major rewrite frankly to permit users to select which stats they are interested in, as well as support for long read technology.  I don't think I'd start from here!

jkbonfield:
Following our office discussion yesterday, when giving a reference we call `count_mismatches_per_cycle` too which also has a memory access equivalent to `mpc_buf[icycle * nquals + qual]` for a nested loop incrementing `icycle`.  This could be `mpc_buf[qual * nbases + icycle]` to invert the columns and rows, but in this case it makes minimal impact.

Before and after perf counts:

```
    10.66%          7218  samtools  samtools              [.] count_mismatches_per_cycle
    10.51%          7273  samtools  samtools              [.] count_mismatches_per_cycle
```

Differences are in the noise threshold.  Even if it did make a difference, it's a tiny change compared to the FFQ code.  I think that's because this is quite a complex loop doing cigar processing and reference comparisons, so the cache hit rate isn't such a big deal.  Unlike the quality-by-cycle array which is a tight loop.  Although how it only ends up as 10% of the CPU for a much harder task I don't know!  Maybe that just demonstrates something more is flawed with the quality stats, but we already know it wants a major rewrite for long read data anyway as the output is basically useless.

jkbonfield:
Also I noted when initially looking at this code that there are all sorts of bits of pointless dead code, such as checking `qual>=stats->nquals` where `stats->nqual` is hardcoded as 256 (now 257) and cannot be changed by the user.  I imagine the more I look the more oddities I'd find, so I decided not to ignore such things and focus only on the CPU aspect.

I'm still unsure if it warrants the complexity of this PR over the one-character initial change (or the minor non-committed one of `i*nquals + qual` to `qual*nbases + i`.

daviesrob:
I think the main speed-up comes from quality values mainly occurring in runs, so the pivoted layout hits recently loaded cache lines more frequently.  As you note, it really needs tidying up, but this does help a bit.

jkbonfield:
> I think the main speed-up comes from quality values mainly occurring in runs, so the pivoted layout hits recently loaded cache lines more frequently. As you note, it really needs tidying up, but this does help a bit.

No - it was due to cache pollution due to using a power of 2.

The reason why the first commit make a significant difference is &array[256*x] means the bottom bits of the address are the same for many values of 'x'.  This purges the cache lines.  With 257 this doesn't happen.  See the initial 11.3% to 3.4% cache miss rate.  That was purely down to the array size and not the array tranposing.  Even if we had one quality value for the entire sequence we'd still get cache thrashing with 256. 

Anyway thanks for merging.  We need to get back to it sometime with a better long read fix!

Issue #2302, opened 2026-02-05
How "bases mapped (cigar)" value in samtools stats is calculated?


YuanTian1991:
Hi:

I am trying to use `samtools stats` to estimate "bases mapped (cigar)" value from some target regions. But I can't figure out how the number is calcualted out. Below is a small test:

This is a sam file subset  "test_subset.sam", contains only 15 reads that all covers posiotion `chr22:31656041`. All 15 reads are 34M in cigar.:

```
@HD     VN:1.3
@SQ     SN:chr1 LN:249250621
@SQ     SN:chr2 LN:243199373
@SQ     SN:chr3 LN:198022430
@SQ     SN:chr4 LN:191154276
@SQ     SN:chr5 LN:180915260
@SQ     SN:chr6 LN:171115067
@SQ     SN:chr7 LN:159138663
@SQ     SN:chr8 LN:146364022
@SQ     SN:chr9 LN:141213431
@SQ     SN:chr10        LN:135534747
@SQ     SN:chr11        LN:135006516
@SQ     SN:chr12        LN:133851895
@SQ     SN:chr13        LN:115169878
@SQ     SN:chr14        LN:107349540
@SQ     SN:chr15        LN:102531392
@SQ     SN:chr16        LN:90354753
@SQ     SN:chr17        LN:81195210
@SQ     SN:chr18        LN:78077248
@SQ     SN:chr19        LN:59128983
@SQ     SN:chr20        LN:63025520
@SQ     SN:chr21        LN:48129895
@SQ     SN:chr22        LN:51304566
@SQ     SN:chrX LN:155270560
@SQ     SN:chrM LN:16571
USI-EAS28:3:8:1631:168#0        16      chr22   31656010        37      34M     *       0       0       CGTGTCCGTGGAGAGTGCCTGCTCCAACTACGCC      \ZZNON\QTUYEGa``R^^``\^`a_`a`X^\`_     XT:A:U  NM:i:0  X0:i:1  X1:i:0  XM:i:0  XO:i:0  XG:i:0  MD:Z:34
USI-EAS28:1:82:1303:1490#0      16      chr22   31656011        37      34M     *       0       0       GTGTCGGTGGAGAGTGCCTGCTCCAACTACGCCA      BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB     XT:A:U  NM:i:1  X0:i:1  X1:i:0  XM:i:1  XO:i:0  XG:i:0  MD:Z:5C28
[...]

PR #2328, opened 2026-04-23
Stats: check before and after region  with --target-regions

whitwham:
For the calculation of "bases mapped (cigar) check for positions before and after the region specified.


whitwham:
Fixes #2302.

vasudeva8:
The changes gave results as required for the matches.
Regarding inserts, they are counted only if insert's end falls with in target region. Shouldn't we be considering partial matches of inserts as well (i.e part insert starting in target region)?

vasudeva8:
Discussed and it is found that the behaviour is correct.

PR #2329, opened 2026-04-23
changes for biobambam2 compatible checksum file merging

### LINKED ISSUE #2323: samtools checksum fails to merge checksums created with --bamseqchksum
(run `samtools --version`)

1.23

Linux 6.17.0-20-generic
x86_64
gcc (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0

1. Create samtool checksum files using the --bamseqchksum option

e.g. samtools checksum --bamseqchksum --output checksum1.chk reads1.bam

2. Attempt to merge the files (the error is also triggered by a no-op merge of one file)

e.g. samtools checksum --output merged.chk --merge *.chk 

```

Incorrect number of columns in line:    all     0               1       1       1       1
Failed to parse checksum file 'checksum1.chk'
[checksum] Failed to process data

```

Here's a real checksum file (taken from a BAM file basecalled from one of ONT's public datasets) that exhibits the issue:

```
###	set	count		b_seq	name_b_seq	b_seq_qual	b_seq_tags(BC,FI,QT,RT,TC)
all	all	1		260296a5	3403a891	4f280cb8	260296a5
all	pass	1		260296a5	3403a891	4f280cb8	260296a5
	all	0		1	1	1	1
	pass	0		1	1	1	1
90b40ed4-85a0-47db-8f1a-1dc8c7ed40bd_dna_r10.4.1_e8.2_400bps_fast@v4.3.0_barcode01	all	1		260296a5	3403a891	4f280cb8	260296a5
90b40ed4-85a0-47db-8f1a-1dc8c7ed40bd_dna_r10.4.1_e8.2_400bps_fast@v4.3.0_barcode01	pass	1		260296a5	3403a891	4f280cb8	260296a5
```

vasudeva8:
Fixes #2323.
Biobambam2 has different no. of fields in checksum output and have empty group name, to cause this.
Updated to handle compatibility mode outputs.
Added aux tags check before merging.
Updated a checksum calculation length as well.

Note: biobambam creates empty/no RG checksum only if header has RG lines and it could be an issue in it though in practice there may not be any such outputs. That is ignored and noRG output lines are created in all cases as reads always belong to either an RG group or noRG group.


bam_checksum.c
```diff
@@ -924,6 +936,65 @@ static int sums_parse(opts *o, char *fn, sums_t *sums, sums_t *noRG,
...
+                    header[i] = H_AUX;
+                    if (!o->tag_str) {
+                        tag = strdup(token+11);
+                        if (tag) {
+                            size_t ln = strlen(tag);
+                            if (ln) {
+                                tag[ln-1] = '\0';   //remove ')'
+                            }
+                            o->tag_free = o->tag_str = tag;
+                        }
```
jkbonfield, line 980: We could add an error check here.  eg:

```
} else {
    fprintf(stderr, "Out of memory\n");
    goto err;
}
```

jkbonfield (commented):
I've stress tested it a bit and deliberately tried to break it.  Looks pretty solid and it passes my real world tests too with separate checksum + checksum -m vs samtools merge + checksum.

I only have one very minor issue with checking strdup.

Thanks Vasu.

bam_checksum.c
vasudeva8: Added the error check and failure.

jkbonfield: Thanks.  Squashed and merged

bam_checksum.c
```diff
@@ -731,7 +731,7 @@ int checksum(sam_global_args *ga, opts *o, char *fn) {
             u32_to_le(b->core.mtid,  mate);
             u64_to_le(b->core.mpos,  mate+4);
             u64_to_le(b->core.isize, mate+12);
-            c.mate = hts_crc32(c.seq, mate, 12);
+            c.mate = hts_crc32(c.seq, mate, 20);
```
rhpvorderman, line 734: @vasudeva8 

What is the rationale behind this change? This makes all checksums generated by 1.23 and 1.23.1 fail even though they were correct.
I have quite a lot of checksums of patients on hand here that I would like to backwards-compatibly check by samtools.

whitwham: The original length 12 was wrong.  The length of mate is 4 + 8 + 8 (line 730), it should have been 20. Not all the data was being used in the crc32 calculations.

whitwham: Any further discussion should be in #2354 so we don't have this in multiple places.

PR #2347, opened 2026-07-01
Fill in mate information on supplementary alignments in fixmate

nh13:
`samtools sort --template-coordinate` requires an `MC` tag on every paired, mate-mapped read, but `samtools fixmate` sets `MC` (and the other mate information) only on primary alignments. Supplementary alignments — emitted for split reads by e.g. `bwa mem` — reach the sorter without `MC`, so it aborts:

```
[bam_sort] error: no MC tag. Please run samtools fixmate on file first.
```

fixmate now fills in mate information (coordinates, orientation, `MC` and `MQ`) on supplementary alignments too, taken from the mate's primary and reusing the helpers already applied to the primary pair. Secondary alignments are left unchanged, matching Picard/htsjdk and fgbio: a supplementary shares its read's primary mate, whereas a secondary is an alternative mapping with no well-defined mate.

The change only adds mate information that was previously absent on supplementary records, so existing output is unaffected and the test suite passes unchanged. Adds a `fixmate` fixture and an end-to-end `sort --template-coordinate` test covering a supplementary alignment.


whitwham:
Thank you.

nh13:
Thanks for getting this in so quickly!

PR #2348, opened 2026-07-02
News update, summer 2026

whitwham:
Additional news,

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -25,12 +25,17 @@ New work and changes:
   (PR #2329, fixes #2323.  Reported by Keith James)
 
 * Speed improvement for `samtools stats`.
-  (PR#2326)
+  (PR #2326)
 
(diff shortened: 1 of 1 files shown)


jkbonfield:
The language looks good.  A very minor point (only worth while as you have a new PR open to change things) is that you have 'PR#' instead of the more usual 'PR #', although I note the speed increase one a couple entries above does the same.

I checked out of curiosity and we do use both, but it's an 84/16% split.  I'm rather ambivalent, so up to you. :)

I checked htslib too - I must be looking for CI distractions - and it's 53% in favour of PR# instead there.  It's strongly skewed by person, with you and I favouring a roughly 2:1 ratio of no space, and Rob is reversed at 1:5 instead (favouring the space).  He's probably right, but shh :-)

whitwham:
There are two PR#s in the new stuff.  I think I will change them.

PR #2350, opened 2026-07-11
Add shared helper for samtools @PG lines; fix reset/rmdup provenance

BenjaminDEMAILLE:
## Summary

Consolidates the duplicated `@PG`-adding code into one shared helper and fixes two subcommands whose `@PG` records were incomplete or missing.

## What changed

**Shared helper.** Every subcommand that writes an alignment file called `sam_hdr_add_pg()` directly with a near-identical block. Those copies are replaced by `samtools_add_pg_line()` (declared in `samtools.h`, implemented in `sam_opts.c`, which is already linked into every subcommand and into `libst.a`). It also folds in the `--no-PG` check, so callers pass their opt-out flag through instead of repeating the condition.

**HTSlib version in `VN`.** When the HTSlib version cannot be deduced from the SAMtools version it is appended to `VN`, e.g. `VN:1.24.1 (with HTSlib 1.25)`. Since the two are released together with matching major and minor version numbers, **`VN` is unchanged in the usual case** and no expected test outputs move.

**Provenance fixes.**
- `reset` wrote a `@PG` line with a command line but no `VN`, unlike every other subcommand. Fixed.
- `rmdup`/`rmdupse` wrote no `@PG` line at all. They now add one and accept `--no-PG`.

**Consistency.** `ampliconclip`'s inverted `add_pg` flag is renamed to the `no_pg` convention used everywhere else.

## Testing

- `make test` — 1000 passed, 0 failures.
- `make maintainer-check` — PASS.
- `test/test.pl` is untouched: because `VN` is unchanged when the versions match, no harness change or expected-output regeneration is needed.
- The version-mismatch branch cannot be exercised through the CLI (`hts_version()` is fixed at link time), so its logic was verified separately across matching, patch-differing (`1.23.1`/`1.23.2`), major/minor-differing, and unparseable-version cases.

## Changes from the first revision

Addresses @daviesrob's review:
- HTSlib version moved from a `DS` field to being appended to `VN`, and omitted when it can be deduced from the SAMtools version.
- Dropped `REPRODUCIBILITY.md`, the `README.md` change, and the sort thread-comparison test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,7 +214,7 @@ faidx.o: faidx.c config.h $(htslib_faidx_h) $(htslib_hts_h) $(htslib_hfile_h) $(
 padding.o: padding.c config.h $(htslib_kstring_h) $(htslib_sam_h) $(htslib_faidx_h) $(sam_opts_h) $(samtools_h)
 phase.o: phase.c config.h $(htslib_hts_h) $(htslib_sam_h) $(htslib_kstring_h) $(sam_opts_h) $(samtools_h) $(htslib_hts_os_h) $(htslib_kseq_h) $(htslib_khash_h) $(htslib_ksort_h)
 reference.o: reference.c config.h $(htslib_sam_h) $(htslib_cram_h) $(samtools_h) $(sam_opts_h)
-sam_opts.o: sam_opts.c config.h $(sam_opts_h)
+sam_opts.o: sam_opts.c config.h $(sam_opts_h) $(samtools_h)
 sam_utils.o: sam_utils.c config.h $(sam_utils_h)
 sam_view.o: sam_view.c config.h $(htslib_sam_h) $(htslib_faidx_h) $(htslib_khash_h) $(htslib_kstring_h) $(htslib_hfile_h) $(htslib_thread_pool_h) $(htslib_hts_expr_h) $(samtools_h) $(sam_opts_h) $(bam_h) $(bedidx_h) $(sam_utils_h)
 sample.o: sample.c config.h $(sample_h) $(htslib_khash_h)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,20 @@
 Release a.b
 -----------
 
+New work and changes:
+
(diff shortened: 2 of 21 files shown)


daviesrob:
This could do with a tidy up.  There's some weird stuff about sort which seems unnecessary.  I don't like the `REPRODUCIBILITY.md` file or the change to `README.md`; they're both examples of low benefit LLM output and should be removed.

As to the functionality, I don't think `DS` is the right place to put the HTSlib version.  If we want to record the HTSlib version, it would be better appended to the `VN` field after the SAMtools version (e.g. `1.23.1 (with HTSlib 1.23.2)`.  Also, SAMtools and HTSlib are released together and so far have always had matching major and minor version numbers (the patch level has occasionally differed).  Given this, the HTSlib part could be omitted if both numbers match as it's possible to work out one from the other.

BenjaminDEMAILLE:
Revised as suggested.

**HTSlib version.** Moved from `DS` to being appended to `VN`, and omitted when it can be deduced from the SAMtools version, i.e. it is only recorded when the major or minor numbers differ:

```
VN:1.24.1 (with HTSlib 1.25)     # only when they diverge
VN:1.24.1                        # normal case, unchanged
```

Patch-level differences are treated as deducible and so omitted, per your point that only major/minor have always matched. Unparseable version strings fall back to recording both.

A useful side effect: since `VN` is unchanged in the normal case, `test/test.pl` needed no modification at all, and the `sam_compare()` hack from the first revision is gone.

**Removed** `REPRODUCIBILITY.md`, the `README.md` change, and the sort thread-comparison test. The latter was also redundant, as the harness already runs the sort tests both single-threaded and with `-@ 2` against the same expected outputs.

**Remaining** is the shared `samtools_add_pg_line()` helper replacing ~20 copies of the same `sam_hdr_add_pg()` block, plus the two provenance fixes (`reset` missing `VN`, `rmdup` missing `@PG` entirely). History tidied to two commits.

`make test` and `make maintainer-check` pass.

daviesrob:
Thanks, this looks much better now.

PR #2353, opened 2026-07-22
Add more CI targets

jkbonfield:
Adds container-build.yml for various containers and machine images (other linux releases), and vm-build.yml for freebsd, openbsd and opensolaris.


jkbonfield:
I've done a lot of updating of the workflows and also refactored .ci_helpers/clone a bit.  This allows all builds to support mention of htslib#PR in the first line of the most recent commit, for both in-development branches on forked repos and also when running as a PR itself (with an implicit merge commit to develop).  It also still supports having forked samtools and htslibs using the same branch name in each, for ease of developing combined changes prior to making a PR.  (Both mechanisms have been tested)

- We now properly support both branches and merge commits.

- Changes ci_helper/clone to use a samtools commit hash instead of an htslib PR.  The log scanning and parsing code is now in this script instead of duplicated throughout the YML files.

- Fix a bug in the original linux-build and macos-build workflows as previously the htslib_pr assessment always failed due to not running within a git clone.

- Add fetch-depth to windows and mac YML, so we have a minimal amount of history instead of none.

- Remove the specific ref request in windows build.  Let it do the merge as done in other workflows.

I'm now happy for this to have a second review.  I'll also update bcftools for that, but will hold off making a PR until this has been reviewed.

jkbonfield:
Removed defunct comments, and also added `--disable-ref-cache` to the VM builds with the comments above them.
Neither really needed that as htslib has since been patched to work around the system specific build issues with ref-cache, but it doesn't harm and it does sort of documents historic issues with particular OSes.

Frankly we could add it to every single CI build as it's irrelevant code for samtools & bcftools CI, but I didn't go that far as it's not the standard build pattern used by most users.

PR #2355, opened 2026-07-24
Update samtools checksum version number

### LINKED ISSUE #2354: Samtools checksum changes output between 1.23.1 and 1.24
The whole point of a checksum is that it is stable. 

I checked to see what causes this and it is this PR: https://github.com/samtools/samtools/pull/2329

I see that indeed the last two columns are supposed to change because of this PR as highlighted here: https://github.com/samtools/samtools/pull/2329/changes.

That is all well and good. But the header still states:

```
# Checksum 1.0 for file:
```

The issue can be resolved by making this `# Checksum 2.0`. (Or reverting the change, but that seems undesirable)

rhpvorderman:
For changes made in #2329 . 

There should be some way to discern that the checksum from 1.23.1 and 1.24 other than that this is a checksum failure. These checksums are not compatible, so the version should be updated.

Fixes #2354 (but only if this ends up in a 1.24.1 release  ASAP)

### DIFF
--- a/bam_checksum.c
+++ b/bam_checksum.c
@@ -565,7 +565,7 @@ int checksum_report(char *fn, opts *o,
         return checksum_bamseqchksum(o, all, noRG, h);
 
     // headers
-    fprintf(o->fp, "# Checksum 1.0 for file:%s%s\n",
+    fprintf(o->fp, "# Checksum 2.0 for file:%s%s\n",
             o->tabs ? "\t" : " ", fn);
     fprintf(o->fp, "# Aux tags:%s%s\n",
             o->tabs ? "\t" : "          ", o->tag_str);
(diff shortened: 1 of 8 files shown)


rhpvorderman:
So now I see that there is some header format checking code. So the capability to use different headers was added, but no updates were made to the header following a checksum algorithm change? What is the rationale behind this?

whitwham:
@rhpvorderman We have had a discussion amongst ourselves about how to handle the different checksums.  I think that we will up the checksum version (maybe to 2.0, maybe to 1.1) but we will also handle v1.0 checksums (by deliberately using the "broken" mate crc32 calculation).

rhpvorderman:
That sounds excellent. Thanks for the quick resolution!

rhpvorderman:
Superseded by: https://github.com/samtools/samtools/pull/2365

PR #2056, opened 2024-05-17
fixmate no longer removes PAIRED flag.

whitwham:
No longer removes paired flag from reads that have no mate (either removed or missing).  

Fixes #2052 using the idea that the PAIRED flag is a sequencing technology indicator not a feature of alignment.  This is a change in the behaviour of fixmate.

PR #2357, opened 2026-07-30
Strip UR tags from header SQ lines

### LINKED ISSUE #2327: Feature Request: Option to strip UR tags from CRAM @SQ header lines
Hi,

Following up on #2283, I would like to suggest adding an option to samtools view to strip the UR tag from header lines when generating CRAM files.

Currently, removing the UR tag requires manual header editing prior to conversion, which is cumbersome in automated pipelines.

Thanks for considering this,
Florian

whitwham:
The new `--remove-ur` option will delete all the UR tags from SQ lines in the header. This commit depends https://github.com/samtools/htslib/pull/2056.  This should fix #2327.

### DIFF
--- a/test/test.pl
+++ b/test/test.pl
@@ -1973,16 +1973,6 @@ sub test_view
 
     my $out = "$$opts{tmp}/view";
 
-    # test UR removal, do this first so a file with UR is created
-    # for later tests.
-    my $bam_with_ur_out = sprintf("%s.test%03d.bam", $out, $test);
-    run_view_test($opts,
-                  msg => "$test: SAM -> BAM -> SAM",
-                  args => ['-b', '--remove-ur', $sam_with_ur, '--no-PG'],
-                  out => $bam_with_ur_out,
-                  compare_sam => $sam_no_ur);
-    $test++;
-
     # SAM -> BAM -> SAM
     my $bam_with_ur_out = sprintf("%s.test%03d.bam", $out, $test);
     run_view_test($opts,
@@ -2129,6 +2119,14 @@ sub test_view
                   out => sprintf("%s.test%03d.sam", $out, $test),
                   stdin => $cram_with_ur_out,
                   compare => $sam_with_ur);
+    $test++;
+
+    my $bam_with_no_ur_out = sprintf("%s.test%03d.bam", $out, $test);
+    run_view_test($opts,
+                  msg => "$test: SAM -> BAM -> SAM",
+                  args => ['-b', '--remove-ur', $sam_with_ur, '--no-PG'],
+                  out => $bam_with_no_ur_out,
+                  compare_sam => $sam_no_ur);
     $test++;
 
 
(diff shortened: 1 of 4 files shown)


test/test.pl
```diff
@@ -1973,6 +1973,16 @@ sub test_view
...
     my $out = "$$opts{tmp}/view";
 
+    # test UR removal, do this first so a file with UR is created
+    # for later tests.
+    my $bam_with_ur_out = sprintf("%s.test%03d.bam", $out, $test);
+    run_view_test($opts,
+                  msg => "$test: SAM -> BAM -> SAM",
+                  args => ['-b', '--remove-ur', $sam_with_ur, '--no-PG'],
+                  out => $bam_with_ur_out,
+                  compare_sam => $sam_no_ur);
```
jkbonfield: The comment states this is creating a file with UR codes for later tests, but the command line is creating one without UR tags.   I'm not sure if this was intending to validate anything, or is simply a place to make sure the new code paths are executed.  Either way, it ought to be named better with `my $bam_without_ur_out` as the variable name.

The comment however makes sense for the next test, which does the same thing without `--remove-ur`, however it also causes the test harness to report:

```
"my" variable $bam_with_ur_out masks earlier declaration in same scope at test/test.pl line 1987.
```

So I suspect the test isn't as intended.

jkbonfield (commented):
I'm not sure this command line option is going to help.  It's likely to confuse the users as to what it's doing.

It's removing the UR tag on the input data, but that means if we combine it with a CRAM creation it adds them back again unless we explicitly do a `-O cram,remove_ur` option.

Eg:

```
$ samtools view -T ../htslib/test/ce.fa ../htslib/test/ce#5b.sam -o  _.cram
$ samtools view --remove-ur -h ../htslib/test/ce#5b.sam -O cram -T ../htslib/test/ce.fa |samtools view -h -T ../htslib/test/ce.fa -|grep SQ
SQ	SN:CHROMOSOME_I	LN:1009800	M5:8ede36131e0dbf3417807e48f77f3ebd UR:/nfs/...
```

The documentation just says `Remove all the UR tags from @SQ lines in the header`.  *the* header is a bit misleading, but my interpretation would be what we get out has UR tags removed (so either removing on input and maintaining that state for output, or removing on output only).

I think when this option is enabled it should also call the relevant `hts_set_opt` call to remove it on all output file descriptors.

jkbonfield:
Claude spotted the typo you've already fixed, but it also found an incorrect PR number listed in the commit message:

"_3. Commit message inconsistencies. The first commit message has a typo ("delate" → "delete") and says "This should fix #2056" (i.e., the samtools PR referencing itself), while the actual PR description correctly says "fix #2327". Worth amending the commit message before merge so history isn't misleading_."

It's not right with 2056 though.  It's not samtools PR self-referencing, but mixing up htslib PR with the samtools issue.  2056 is something different in this repo.

whitwham:
Yes I know it is wrong.  I changed it in the PR.

test/test.pl
whitwham: The test works as intended.  The comment may be misleading.

PR #2365, opened 2026-08-18
Move checksum version to 2.0

### LINKED ISSUE #2354: Samtools checksum changes output between 1.23.1 and 1.24

whitwham:
Checksum calculations differed between samtools v1.24 and previous
releases.  In earlier releases, the "--check-mate" option produced incorrect
checksums.  This was fixed in v1.24 but made already existing checksums
invalid.

Therefore, the fixed checksum is now designated Checksum 2.0 while the
previous one is left at Checksum 1.0 in order differentiate the versions.

A new "--version-1" option will set samtools checksum to version 1.0
behaviour for backwards compatibility.

Should fix https://github.com/samtools/samtools/issues/2354.

### DIFF
--- a/bam_checksum.c
+++ b/bam_checksum.c
@@ -737,13 +737,13 @@ int checksum(sam_global_args *ga, opts *o, version_number *ver, char *fn) {
                 u32_to_le(b->core.mtid,  mate);
                 u64_to_le(b->core.mpos,  mate+4);
                 u64_to_le(b->core.isize, mate+12);
-                c.mate = hts_crc32(c.seq, mate, 20);
+                c.mate = hts_crc32(c.seq, mate, sizeof(mate));
             } else {
                 // this replicates the mistake in checksum v1.0
                 uint8_t mate[4+8+8];
                 u32_to_le(b->core.mtid,  mate);
                 u64_to_le(b->core.mpos,  mate+4);
-                c.mate = hts_crc32(c.seq, mate, 12);
+                c.mate = hts_crc32(c.seq, mate, 12); // mistake was here
             }
         }
 
--- a/bam_checksum.c
+++ b/bam_checksum.c
@@ -732,11 +732,19 @@ int checksum(sam_global_args *ga, opts *o, version_number *ver, char *fn) {
 
         // flag + seq + rnext + pnext + tlen
         if (o->check_mate) {
-            uint8_t mate[4+8+8];
-            u32_to_le(b->core.mtid,  mate);
-            u64_to_le(b->core.mpos,  mate+4);
-            u64_to_le(b->core.isize, mate+12);
-            c.mate = hts_crc32(c.seq, mate, 20);
+            if (ver->major >= 2) {
+                uint8_t mate[4+8+8];
+                u32_to_le(b->core.mtid,  mate);
+                u64_to_le(b->core.mpos,  mate+4);
+                u64_to_le(b->core.isize, mate+12);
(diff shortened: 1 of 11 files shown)


whitwham:
@rhpvorderman Does this help you?  If not, let me know and I will try something else.

bam_checksum.c
```diff
@@ -727,11 +732,19 @@ int checksum(sam_global_args *ga, opts *o, char *fn) {
...
-            u32_to_le(b->core.mtid,  mate);
-            u64_to_le(b->core.mpos,  mate+4);
-            u64_to_le(b->core.isize, mate+12);
-            c.mate = hts_crc32(c.seq, mate, 20);
+            if (ver->major >= 2) {
+                uint8_t mate[4+8+8];
+                u32_to_le(b->core.mtid,  mate);
+                u64_to_le(b->core.mpos,  mate+4);
+                u64_to_le(b->core.isize, mate+12);
+                c.mate = hts_crc32(c.seq, mate, 20);
```
rhpvorderman: It is also possible to use `sizeof(mate)` instead of a literal 20 here. That should have the same result.

rhpvorderman:
Yes this would help me massively. Thanks!

EDIT: And sorry for my very late reply. I was on holiday for 3 weeks.

whitwham:
> Yes this would help me massively. Thanks!
> 
> EDIT: And sorry for my very late reply. I was on holiday for 3 weeks.

Well, I have been on holiday for the best part of 2 weeks so no harm done.

whitwham:
New problem: if someone has been using the fixed current v1.0, then moving it to v2.0 may cause problems.

rhpvorderman:
It is possible for v1.0 to simply check both checksums? All that does is increase the probability of a false positive correct by 2x. Which effectively turns a 32-bit CRC in a 31-bit CRC, which is still pretty reliable.

jkbonfield:
> It is possible for v1.0 to simply check both checksums? All that does is increase the probability of a false positive correct by 2x. Which effectively turns a 32-bit CRC in a 31-bit CRC, which is still pretty reliable.

I'm not sure if that is sufficient for merging checksum files together as that's done without parsing the BAM files so it cannot compute alternatives and pick which works.

However it's a small window of time that it's different (one release) and maybe we just have to document it as a caveat in the man page.

jkbonfield:
I squashed the last commit, rebased it, and added a trivial new one to add to the man page.  I'll merge later today if you don't object to the new sentence in the manual.

whitwham:
> I squashed the last commit, rebased it, and added a trivial new one to add to the man page. I'll merge later today if you don't object to the new sentence in the manual.

Ah, it looks like I didn't push my last change to the man page.  Never mind, your text works just as well (if not better).

jkbonfield:
That explains why we've been confused over the status of whether it's ready. :)

Anyway we got there in the end.  Thanks.

rhpvorderman:
Thanks all for fixing this issue! I think adding an explanation to the man page is an elegant way to solve it.

PR #2366, opened 2026-08-18
Check string aux tag types before accessing data

daviesrob:
Following 681b50ea, add more checks on aux tag types to ensure that the stored data is a NUL-terminated string before accessing it.

In most cases tags with unexpected types are simply ignored as if they don't exist.  An exception is samtools stats, which already reports an error if the tag it expects to split on is not present, so now it also does if the tag has the wrong type. The stats code is also made more efficient so that it doesn't try to `strdup()` the tag value unless it needs to add a new hash entry.

The stats man page is updated to note that the tag used for the `-S` option must have a string type.

### DIFF
--- a/bam.c
+++ b/bam.c
@@ -38,7 +38,7 @@ const char *bam_get_library(sam_hdr_t *h, const bam1_t *b)
     kstring_t lib = { 0, 0, NULL };
     rg = (char *)bam_aux_get(b, "RG");
 
-    if (!rg)
+    if (!rg || *rg != 'Z')
         return NULL;
     else
         rg++;
--- a/bam_ampliconclip.c
+++ b/bam_ampliconclip.c
@@ -657,7 +657,9 @@ static inline int tag_original_data(bam1_t *orig, kstring_t *oa_tag) {
 
     // if there is an existing OA tag the new one gets appended to it
     if ((old_oa_tag = bam_aux_get(orig, "OA"))) {
-        res |= ksprintf(oa_tag, "%s", bam_aux2Z(old_oa_tag)) < 0;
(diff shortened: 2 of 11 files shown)


bam.c
```diff
@@ -38,7 +38,7 @@ const char *bam_get_library(sam_hdr_t *h, const bam1_t *b)
     kstring_t lib = { 0, 0, NULL };
     rg = (char *)bam_aux_get(b, "RG");
 
-    if (!rg)
```
jkbonfield, line 41: The fundamental reason these bugs exist is because the htslib API is poor.  People don't want to ask "find me an RG tag of any type", they want "find me the RG:Z" tag.

We know with the current API you *should* be calling `bam_aug_get` followed by a type check or a call to `bam_aux2Z` (which oddly does H too), but the evidence shows it's an easy trap to fall into and people have historically fallen into it repeatedly.  Hence this PR.

However it feels like the wrong fix to simply correct all those historic mistakes without fixing the cause of them occurring.

Surely it would be easier to have:

```
if (!(rg = bam_aux_getZ(b, "RG")))
    return NULL;
```

The code is simpler, less error prone, and we're changing all this code already so it seems like a lost opportunity to not do it using a newer more sensible API.  Yes we could merge in this PR and then do a subsequent htslib one to extend the API and a second samtools PR to change everything you've done here, but it feels like it's doubling up the effort.

daviesrob: Feel free to add such an interface if you think it would be useful.  However, as this fixes the bugs in a way that can be easily back-ported to earlier release branches, I'd prefer to go with this for now.  Upgrading to the new API for the next full release would not be hard.

PR #2367, opened 2026-08-18
Spelling corrections [minor]

jmarshall:
Mostly in comments, but a few in strings that may be displayed to the user.


daviesrob:
Looks good, thanks.

PR #2368, opened 2026-08-19
Improve checks around khash insertions

daviesrob:
Add checks to ensure entries have been correctly inserted into hash tables.  Ensure that an error is reported in some way if not, and where necessary that any data structures are tidied up so they are not left in a half-constructed state and memory leaks are prevented.

### DIFF
--- a/bam2depth.c
+++ b/bam2depth.c
@@ -602,16 +602,19 @@ static int fastdepth_core(depth_opt *opt, uint32_t nfiles, char **fn,
             if (k == kh_end(overlaps[i])) {
                 // not seen before
                 hts_pos_t endpos = bam_endpos(b[i]);
-
                 // Don't add if mate location is known and can't overlap.
                 if (b[i]->core.mpos == -1 ||
                     (b[i]->core.tid == b[i]->core.mtid &&
                      b[i]->core.mpos <= endpos)) {
-                    k = kh_put(olap_hash, overlaps[i], bam_get_qname(b[i]),
-                               &ret);
-                    if (ret < 0)
+                    char *name_copy = strdup(bam_get_qname(b[i]));
+                    if (!name_copy)
                         return -1;
-                    kh_key(overlaps[i], k) = strdup(bam_get_qname(b[i]));
+
+                    k = kh_put(olap_hash, overlaps[i], name_copy, &ret);
+                    if (ret < 0) {
+                        free(name_copy);
+                        return -1;
+                    }
                     kh_value(overlaps[i], k) = endpos;
                 }
             } else {
--- a/bam_rmdupse.c
+++ b/bam_rmdupse.c
@@ -61,7 +61,13 @@ static lib_aux_t *get_aux(khash_t(lib) *aux, const char *lib)
         int ret;
         char *p = strdup(lib);
         lib_aux_t *q;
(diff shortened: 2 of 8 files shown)


bam2depth.c
```diff
@@ -602,16 +602,19 @@ static int fastdepth_core(depth_opt *opt, uint32_t nfiles, char **fn,
...
-                    if (ret < 0)
+                    char *name_copy = strdup(bam_get_qname(b[i]));
+                    if (!name_copy)
+                        return -1;
+
+                    k = kh_put(olap_hash, overlaps[i], name_copy, &ret);
+                    if (ret < 0) {
+                        free(name_copy);
                         return -1;
-                    kh_key(overlaps[i], k) = strdup(bam_get_qname(b[i]));
```
jkbonfield, line 614: Why all the song and dance (doubling the number of lines involved) instead of a simple:

```
if (!(kh_key(overlaps[i], k) = strdup(bam_get_qname(b[i]))) {
    ret = -1;
    goto err;
}
```

(Note we probably ought to be tidying up with the err: label instead of `return -1` here.)

Yes the hash table has a key with a NULL value at this point, but the very next thing it's doing is failing and bombing out, which will tidy everything up.  It's good maybe to do the complicated do-si-do move in htslib where we don't know how the caller is handling the error, but samtools basically just cascades up to main and exiting on any memory failure.

FWIW the code looks valid, should work and you've done it now so fair enough, but as a general rule I don't like unnecessary complication as it increases the risks.

bam2depth.c
```diff
@@ -602,16 +602,19 @@ static int fastdepth_core(depth_opt *opt, uint32_t nfiles, char **fn,
             if (k == kh_end(overlaps[i])) {
                 // not seen before
                 hts_pos_t endpos = bam_endpos(b[i]);
-
```
jkbonfield, line 605: Minor nitpick, in case you're editing for other reasons.  Was this an accidental change?  I don't see the point, and IMO it was more readable with the white space there to separate the commented blocks of code.

bam_rmdupse.c
```diff
@@ -81,6 +87,7 @@ static inline int sum_qual(const bam1_t *b)
 static inline elem_t *push_queue(queue_t *queue, const bam1_t *b, int endpos, int score)
 {
     elem_t *p = kl_pushp(q, queue);
+    if (!p) { perror(NULL); exit(EXIT_FAILURE); }
```
jkbonfield, line 90: Given we're being so cautious to free up any memory before exiting elsewhere, this feels like a "I give up" statement!  It's a curious Jekyll and Hyde approach to error handling. :-)

Although frankly that's exactly what this deserves.  Rmdup and rmdupse were officially obsoleted in 2017 by markdup, and it no longer appears in the samtools usage statement.  I don't really have an issue with it leaking memory on the way out if it fails, as it's not likely we'll be fuzz testing it and no one should be using it.  It just seems at odds with the painstakingly checks elsewhere.

jkbonfield (commented):
Approving despite comments as they're minor and the biggest issues are with defunct code anyway so I'm happy to simply let them go.

However I think we probably ought not to be spending too much time naval gazing obsolete and/or deprecated code given the list of other things to do.

PR #2369, opened 2026-08-19
Improve error reporting in samtools dict

daviesrob:
Be less optimistic about memory allocations working.

Change `args` in `dict_main()` to be on the stack so it is automatically cleaned up on exit, and make pointers to command-line arguments const.

Keep the return value from `hts_getline()` so that it's possible to tell if it reported a file read error.

Check `ferror()` on the output so at least some write errors can be detected.

Propagate errors up to `main()` via function return values instead of relying on `exit()` to bail out when things go wrong.

### DIFF
--- a/dict.c
+++ b/dict.c
@@ -39,40 +39,44 @@ KSEQ_INIT(gzFile, gzread)
 
 typedef struct _args_t
 {
-    char *output_fname, *alt_fname;
-    char *assembly, *species, *uri;
+    const char *output_fname, *alt_fname;
+    const char *assembly, *species, *uri;
     int  alias, header;
     khash_t(str) *is_alt;
 }
 args_t;
 
-static void write_dict(const char *fn, args_t *args)
+static int write_dict(const char *fn, args_t *args)
 {
-    hts_md5_context *md5;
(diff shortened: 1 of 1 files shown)


dict.c
```diff
@@ -169,8 +197,17 @@ static int dict_usage(void)
...
+    args_t args = {
+        NULL, // output_fname
+        NULL, // alt_fname
+        NULL, // assembly
+        NULL, // species
+        NULL, // uri
+        0, // alias
+        1, // header
+        NULL // is_alt
+    };
```
jkbonfield, line 210: The modern self-describing way to do this avoiding the need for comments is with C99 named allocators.  It's less error prone too as editing the struct wouldn't horribly break the code.

Eg:

```
args_t args = {
    .output_fname = NULL,
    .alt_fname = NULL,
    .assembly = NULL,
    .species = NULL,
    .uri = NULL,
    .alias = 0,
    .header = 1,
    .is_alt = NULL
};
```

We already use this better language feature elsewhere.

daviesrob: Updated to use designators, as suggested.

PR #2370, opened 2026-08-19
Improve error checking in samtools phase

daviesrob:
Be less optimistic about memory allocations working.

Check for `ks_getuntil()` failure in `loadpos()`.

Ensure errors are propagated by callers.

Try to leak less memory when dealing with errors.

While `phase` is mostly unsupported, it's probably a good idea to fix the more obvious problems with it given that bug fixing is now fashionable.

### DIFF
--- a/phase.c
+++ b/phase.c
@@ -1,7 +1,7 @@
 /*  phase.c -- phase subcommand.
 
     Copyright (C) 2011 Broad Institute.
-    Copyright (C) 2013-2016, 2019, 2024 Genome Research Ltd.
+    Copyright (C) 2013-2016, 2019, 2024, 2026 Genome Research Ltd.
 
     Author: Heng Li <lh3@sanger.ac.uk>
 
@@ -62,7 +62,7 @@ typedef struct {
     samFile* out[3];
     sam_hdr_t* out_hdr[3];
     // alignment queue
-    int n, m;
+    size_t n, m;
     bam1_t **b;
 } phaseg_t;
 
@@ -162,18 +162,22 @@ static int **count_all(int l, int vpos, nseq_t *hash)
 // phasing
 static int8_t *dynaprog(int l, int vpos, int **w)
 {
-    int *f[2], *curr, *prev, max, i;
-    int8_t **b, *h = 0;
+    int *f[2]= { NULL, NULL }, *curr, *prev, max, i;
+    int8_t **b = NULL, *h = NULL;
     uint32_t x, z = 1u<<(l-1), mask = (1u<<l) - 1;
     f[0] = calloc(z, sizeof(int));
     f[1] = calloc(z, sizeof(int));
     b = calloc(vpos, sizeof(int8_t*));
+    if (!f[0] || !f[1] || !b)
+        goto fail;
     prev = f[0]; curr = f[1];
     // fill the backtrack matrix
     for (i = 0; i < vpos; ++i) {
         int *wi = w[i], *tmp;
         int8_t *bi;
         bi = b[i] = calloc(z, 1);
+        if (!bi)
+            goto fail;
         /* In the following, x is the current state, which is the
          * lexicographically smaller local haplotype. xc is the complement of
          * x, or the larger local haplotype; y0 and y1 are the two predecessors
@@ -193,6 +197,8 @@ static int8_t *dynaprog(int l, int vpos, int **w)
         uint32_t max_x = 0;
         int which = 0;
         h = calloc(vpos, 1);
+        if (!h)
+            goto fail;
         for (x = 0, max = 0, max_x = 0; x < z; ++x)
             if (prev[x] > max) max = prev[x], max_x = x;
         for (i = vpos - 1, x = max_x; i >= 0; --i) {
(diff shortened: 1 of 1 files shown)


phase.c
```diff
@@ -539,23 +612,44 @@ static khash_t(set64) *loadpos(const char *fn, sam_hdr_t *h)
 
     hash = kh_init(set64);
     str = calloc(1, sizeof(kstring_t));
-
     ks = ks_init(fp);
-    while (ks_getuntil(ks, 0, str, &dret) >= 0) {
+
+    if (!hash || !str || !fp)
+        goto fail;
```
jkbonfield: `ks = ks_init(fp)` can also fail.  Did you mean `!ks` instead of `!fp` here?

phase.c
```diff
@@ -755,6 +864,10 @@ int main_phase(int argc, char *argv[])
         if (vpos == max_vpos) {
             max_vpos = max_vpos? max_vpos<<1 : 128;
             cns = realloc(cns, max_vpos * 8);
```
jkbonfield: Elsewhere you change such things to:

```
uint64_t *new_cns = realloc(cns, max_vpos * 8);
if (new_cns) {
    cns = new_cns;
} else {
    // handle error
}
```

Although personally I don't care too much about such errors, as we're basically exiting at this point and whether we free it or the OS frees it makes little difference.  It does matter when it's data driven though (eg fuzzing), but when genuinely oom on small fixed-sized allocations then there's really little to do.

phase.c
```diff
@@ -460,13 +514,19 @@ static int phase(phaseg_t *g, const char *chr, int vpos, uint64_t *cns, nseq_t *
...
+    free(regmask); regmask = NULL;
+    free(sitemask); sitemask = NULL;
     seqs = calloc(n_seqs, sizeof(frag_t*));
+    if (!seqs)
+        goto fail;
     for (k = 0, i = 0; k < kh_end(hash); ++k)
         if (kh_exist(hash, k) && kh_val(hash, k).vpos < vpos && !kh_val(hash, k).single)
             seqs[i++] = &kh_val(hash, k);
     n_seqs = i;
-    ks_introsort_rseq(n_seqs, seqs);
```
jkbonfield, line 469: Not that this is your code, but I spent some time grepping for `ks_introsort_rseq` given it's an explicit function here.

Then I found that it's from here:

```
#define ks_introsort(name, n, a) ks_introsort_##name(n, a)
```

Meh... exploiting internal implementation knowledge of the k* templating system rather than the official `ks_introsort(rseq, n_seqs, seqs)`.  :-(

jkbonfield:
If we're being anal about correctness of defunct code, I'll point out that it's triggering undefined behaviour too:

```
$ samtools phase ~/scratch/data/9827_2#49.1m.bam > /dev/null
phase.c:396:13: runtime error: null pointer passed as argument 1, which is declared to never be null
/usr/include/string.h:48:14: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior phase.c:396:13 
phase.c:396:19: runtime error: null pointer passed as argument 2, which is declared to never be null
/usr/include/string.h:48:14: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior phase.c:396:19 
```

The cause is `memmove(g->b, g->b + i, (g->n - i) * sizeof(void*));` at the end of the `dump_aln` function.

In this call, `g->b` is NULL, along with `i` and `g->n` being 0.  So it's basically `memmove(NULL, NULL, 0)`.
We could add an `if (g->b)` guard maybe.

phase.c
```diff
@@ -326,6 +362,11 @@ static uint64_t *genmask(int vpos, const uint64_t *pcnt, int *_n)
     }
     *_n = n;
     return list;
+
+ fail:
+    print_error_errno("phase", "Couldn't generate mask");
+    free(list);
+    return NULL;
```
jkbonfield, line 371: Unfortunately this is a change of behaviour.

My input data (local):

```
../htslib/test/test_view -N 10000 -p /tmp/_x.bam ~/scratch/data/9827_2#49.1m.bam
samtools phase /tmp/_x.bam
```

This calls genmask with vpos=2, but at no point does it grow `list`.  So it ends up as NULL at the end because it didn't get expanded, and is then treated as an error.  This was glossed over before because we didn't check errors, but it also didn't use it because `*n == 0`.

We could change the fail case to have `*n = -1` and the fail condition should check `if (!mask && n_masked < 0)`.

I think this highlights the problem of attempting to polish legacy code that has no meaningful test data!

phase.c
daviesrob: Yes.

phase.c
daviesrob: Yes, I'd missed this one, along with another case in `readaln()` (and a couple of other issues in that function).

phase.c
daviesrob: I've solved this by making `genmask()` always allocate a list so it has something to return, and skipping the `regmask` bit completely if `n_masked == 0` as it isn't accessed in that case.

daviesrob:
I was unable to reproduce the NULL pointer error when running on your file, however there's obviously no point in calling `memmove()` if there are no items left to move, so it's easily avoided.  I've also fixed up all the memory leaks in `main_phase()` now, so it's looking a lot more tidy when running with sanitizers.

PR #2371, opened 2026-08-20
Improve memory handling when building translation tables for merge

daviesrob:
Check that hash table insertions and kstring functions worked.

Ensure pointer ownership is transferred correctly when setting hash table keys.

Clean up kstrings on failure to avoid leaking them.

### DIFF
--- a/bam_sort.c
+++ b/bam_sort.c
@@ -969,7 +969,32 @@ static void free_merged_header(merged_header_t *merged_hdr) {
     free(merged_hdr);
 }
 
-static void bam_translate(bam1_t* b, trans_tbl_t* tbl)
+static int warn_missing_tag(const char *tag_type, const char *tag_val,
+                            const char *qname, kh_c2c_t *trans_table) {
+    char *tmp = strdup(tag_val);
+    if (!tmp)
+        return -1;
+    fprintf(stderr,
+            "[bam_translate] %s tag \"%s\" on read \"%s\" encountered "
+            "with no corresponding entry in header, tag lost. "
+            "Unknown tags are only reported once per input file for "
+            "each tag ID.\n",
+            tag_type, tag_val, qname);
+    // Prevent future whinges
+    int in_there = 0;
+    khiter_t k = kh_put(c2c, trans_table, tmp, &in_there);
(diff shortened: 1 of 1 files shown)


bam_sort.c
jkbonfield: Given we have above:

```
        } else if (decoded_rg) {
            char *tmp = strdup(decoded_rg);
```

The `if (tmp)` here should always be true unless we've failed to allocate memory, in which case it's silent rather than reporting the issue.  Similarly for the `if (decoded_pg)` section below.

daviesrob: Yes.  This exists simply to stop the error message above from being printed more than once.  So the consequence of not being able to insert an entry is that you get the message multiple times.  This is both unlikely as the allocations will all be small, and not a major problem if it happens.

jkbonfield: I don't understand.

The conditional isn't on whether it succeeded in inserting an entry, but on whether `strdup` returned NULL or not.

When it does, we do the fprintf and add to the trans table.  When it doesn't, we silently ignore the allocation failure.  This doesn't make sense and doesn't seem to be related to the comment.   Why are we not treating a memory allocation failure as a serious issue?

daviesrob: Because in this case an allocation failure isn't that serious (or really that likely unless string in the `RG:Z` tag was very long).  All that's happening it that it's adding an entry in the hash table so if the same RG name turns up it returns a hash entry and is removed silently at line 985 above.  So the only consequence of not adding the entry is to cause the message on line 992 to be printed more than once if the same read group name turns up.  The BAM aux tag will still be deleted, and the program may be able to finish as long as it doesn't fail an allocation somewhere else.

daviesrob: After a bit more refactoring, errors here will now be reported as a failure.

daviesrob:
Having noticed that `bam_translate()` wasn't checking the result from `bam_aux_append()`, which could result in silent aux tags removal, I've added an extra commit to fix that.  It also refactors the missing tag reports to remove duplication, take them out of the main `bam_translate()` flow, and allow errors with adding the stub hash entry to be caught properly.

PR #2376, opened 2026-09-02
Markdup Refactor

whitwham:
A lot of the work being done in markdup was in one function that was far too big.  This PR breaks that function down into smaller functions.  There should be no change in output or speed.


daviesrob:
Looks OK as a rearrangement.  Would you like the commits squashed on merge?

whitwham:
> Looks OK as a rearrangement. Would you like the commits squashed on merge?

I don't mind.  The individual commits are not that important.

daviesrob:
OK, will squash then.

PR #2379, opened 2026-09-08
stats: grow the coverage ring buffer instead of wrapping around

### LINKED ISSUE #2378: stats: coverage distribution (COV, -t/-g target percentage) wraps around a fixed ring buffer for spliced alignments and drops counts on buffer reallocation
I am not an expert so maintainers, please feel free to ditch this if I am missing something; but it made sense to me when I came across it so thought I would share this information. I attached a potential fix in case it helps.

#### Are you using the latest version of samtools and HTSlib? If not, please specify.

Yes: `develop` @ `ce612d2` built against HTSlib `develop` @ `e503e04` (`samtools ce612d2 / Using htslib 1.24-1-ge503e04`). The same output comes from 1.19.2 (Ubuntu package), 1.10 and 1.9 built from their tags.

- OS: Linux 6.18 (Ubuntu 24.04 container), `uname -sr`: Linux 6.18.44
- compiler: gcc 13.3.0

`samtools stats` accumulates the coverage distribution (the `COV` rows, and the `SN percentage of target genome with coverage > N` line under `-t`/`-g`) in a ring buffer of `5 × nbases` cells (1,500 for reads up to 300 bp) that is indexed modulo its size relative to the current read start (`stats.c`, `round_buffer_insert_read` / `round_buffer_lidx2ridx`). A block that ends more than 1,500 positions after the read start — the far exon of a spliced RNA-seq alignment — wraps around and is added to the positions just after the read start. Two reads are enough to show it:

```sh
A=$(printf 'A%.0s' $(seq 50)); Q=$(printf 'I%.0s' $(seq 50))
printf '@HD\tVN:1.4\tSO:coordinate\n@SQ\tSN:ref\tLN:5000\n' > spliced.sam
printf 'r1\t0\tref\t1\t60\t50M\t*\t0\t0\t%s\t%s\n' "$A" "$Q" >> spliced.sam
printf 'r2\t0\tref\t1\t60\t50M1450N50M\t*\t0\t0\t%s%s\t%s%s\n' "$A" "$A" "$Q" "$Q" >> spliced.sam
samtools stats spliced.sam | grep '^COV'
```

Expected (positions 1–50 have depth 2, positions 1501–1550 have depth 1):

```
COV	[1-1]	1	50
COV	[2-2]	2	50
```

Got:

```
COV	[3-3]	3	50
```

Changing the intron to `1000N` (block within the buffer) gives the expected two rows, and shrinking the example showed the condition is exactly `block end − read start > 5 × nbases`: any spliced read with an intron longer than ~1.4 kb at 100 bp read length. On a simulated six-exon gene (3,000 100-bp reads, introns 0.3–12 kb) the covered-position count comes out 2,591 against a true 2,197, and with `-t` on the exon list `percentage of target genome with coverage > 0 (%)` is reported as `117.77` (true value 99.86).

A second, smaller defect sits in `realloc_buffers()`: when a read longer than any seen so far (≥ 300 bp at first) forces the buffers to grow while coverage is pending, the pending ring buffer is copied with

```c
memcpy(rbuffer, stats->cov_rbuf.buffer+stats->cov_rbuf.start, n);
```

where `n` is a number of `int` cells, not bytes, so three quarters of the pending counts are dropped whenever the pending window sits near the end of the ring. Ten 100-bp reads at positions 1401–1446 followed by a 350-bp read at 1451 give `COV [11-11] 11 10` instead of `11 50` (`COV [1-1] 1 300` instead of `260`). On long-read data with growing lengths this repeats at every new maximum length.

A patch that grows the ring buffer on demand (doubling, linearising the pending counts) and reuses that for the reallocation, with two regression tests under `test/stat/` (both fail on unmodified `develop`, the full `make test` passes with the patch), is ready and will follow as a PR.

Found in Mytochondria, a volunteer project that checks the numerical core of research software and verifies every finding by execution (methods and harnesses: https://github.com/cindykrafft/mytochondria/tree/main/audits/samtools)

---
_Generated by [Claude Code](https://claude.ai/code)_

cindykrafft:
Fixes #2378.

`samtools stats` accumulates the coverage distribution (`COV` rows and the `-t`/`-g`
target percentage) in a ring buffer of `5 × nbases` cells indexed modulo its size
relative to the current read start. An aligned block ending more than that beyond the
read start — the far exon of a spliced alignment — wrapped around onto positions near
the read start. With 100-bp reads any intron longer than about 1.4 kb triggers it; on a
simulated six-exon gene the covered-position count was 18 % too high and the `-t`/`-g`
percentage was reported as 117.77 %.

This makes `round_buffer_insert_read()` grow the buffer (doubling, linearising the
pending counts) whenever a block would reach beyond it, through a new
`round_buffer_resize()`. The same helper replaces the copy in `realloc_buffers()`, which
passed element counts to `memcpy()` where byte counts were needed and dropped three
quarters of the pending counts whenever a longer read arrived while coverage was pending
near the end of the ring (mixed-length or long-read input).

Memory: the buffer now grows to the longest span between a read start and the end of
one of its blocks, so a 1-Mb intron costs 4 MB of `int`s; previously the size was fixed
at `5 × nbases`. A cap with a warning could be added if that is a concern.

Tests: `test/stat/20_spliced.sam` (the two-read wrap case) and
`test/stat/21_mixed_lengths.sam` (the reallocation case), compared on their `COV` rows;
both fail on unmodified `develop` and pass with this change. `make test` otherwise
unchanged (numbers in the audit's `upstream/README.md`).

NEWS entry included under "Bug fixes".

Assisted-by: Claude:claude-fable-5-1


daviesrob:
Looks OK, thanks.  I've confirmed that the results match those from the original code when modified to start with a buffer that doesn't need to be resized.

PR #2384, opened 2026-09-18
Avoid repeated strlen scans in aux tag parsing

mikebamaworldx23-crypto:
## Summary

This change removes repeated strlen() calls while parsing --aux/-M style tag lists. The loop now checks the current pointer state directly instead of rescanning the remaining string on each iteration, which avoids unnecessary O(n) work on long tag lists.

## Why this helps

parse_aux_list() previously called strlen(optarg) on each loop iteration. For long comma-separated aux tag lists, this repeatedly scans the suffix of the same string. Checking optarg[0] and optarg[1] avoids those repeated scans with no change in behavior.

## Validation

- git diff --check
- Build tooling is not installed in this environment, so a full compiler run was not possible here.


daviesrob:
Please take note of the [Contribution guide](https://github.com/samtools/samtools/blob/develop/CONTRIBUTING.md), particularly the parts on [signing your work](https://github.com/samtools/samtools/blob/develop/CONTRIBUTING.md#signing-your-work) and [AI policy](https://github.com/samtools/samtools/blob/develop/CONTRIBUTING.md#ai-policy).  We need a "Signed-of-by:" with a real email address, along with "Assisted-by:" (NOT "Co-authored-by") for the AI model you used.

As for the change, it looks reasonable at first glance, although using `strlen()` here probably isn't as bad as it looks.  `strlen()` has for a long time now had an `__attribute_pure__` annotation, so the compiler should be able to optimize the repeated calls away as long as the string passed to the function remains unchanged.

